PR #238, opened 2014-12-31
Add Group contributions section

maurizi:
Also fixes a TODO on the existing user profile contributions section

### DIFF
--- a/src/nyc_trees/apps/users/views/group.py
+++ b/src/nyc_trees/apps/users/views/group.py
@@ -65,7 +65,7 @@ def group_detail(request):
     else:
         block_percent = "0.0%"
 
-    events_held = Event.objects.filter(group=group, ends_at__gt=now())
+    events_held = Event.objects.filter(group=group, ends_at__lt=now())
     num_events_held = events_held.count()
 
     num_event_attendees = EventRegistration.objects \
--- a/src/nyc_trees/apps/users/views/group.py
+++ b/src/nyc_trees/apps/users/views/group.py
@@ -5,13 +5,19 @@ from __future__ import division
 
 from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
+from django.utils.timezone import now
+
+from libs.sql import get_group_tree_count
 
 from apps.core.helpers import user_is_group_admin
 from apps.core.models import Group
-from apps.users.models import Follow
 
+from apps.users.models import Follow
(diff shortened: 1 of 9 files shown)


jwalgran:
Reviewing this now

jwalgran:
The top level ``libs`` module is intended to contain functions that are generally useful across the entire application. The SQL statements are survey-specific, so organizing them under ``apps/survey/sql.py`` may be more appropriate.

maurizi:
@jwalgran I debated putting `sql.py` into `survey`.  Ultimately though, it doesn't *import* anything from `survey`, because it uses raw SQL, and it won't be used by anything *in* `survey`, so putting it there seems wrong (It will be used in the User Profile, Group Detail, and the dashboard)

`libs/` seemed like a good fallback

jwalgran:
``test_contributions_section_contents`` isn't covering the "Events" section.

maurizi:
@jwalgran Can you take another look at this?

jwalgran:
I have completed a feature and I now have time to review the open pull requests assigned to me.

src/nyc_trees/apps/users/views/group.py
```diff
@@ -45,13 +51,42 @@ def group_detail(request):
...
+    else:
+        block_percent = "0.0%"
+
+    events_held = Event.objects.filter(group=group, ends_at__lt=now())
+    num_events_held = events_held.count()
+
+    num_event_attendees = EventRegistration.objects \
+        .filter(event__in=events_held) \
+        .filter(did_attend=True) \
+        .distinct('user') \
```
jwalgran: We should ask Parks to confirm this method of attendance counting (unless they already have). It may be desirable to count "2" if a person attends two, distinct events. If we don't yet have a ruling, it need not block the PR.

jwalgran:
+1

src/nyc_trees/apps/users/views/group.py
maurizi: I added https://github.com/azavea/nyc-trees/issues/251 to get feedback from Parks on this.

dboyer:
@maurizi Parks has confirmed that the items in the contributions section on the Groups page should be - 

- trees mapped
- events held
- % of the group's area mapped
- event attendees

Event attendees should be total attendees - a straight up count of the number of people at all events. This is not a unique attendee count; people should be counted twice if they attend more than one event.

PR #244, opened 2015-01-02
Add support for running Vagrant environments in parallel

hectcastro:
This changeset builds on the environmental variable virtual machine IP assignment by making IP address assignment to the application virtual machine occur via DHCP. In addition, the `NYC_TREES_TEST_DB_NAME` environmental variable provides a way to change the test database.

### DIFF
--- a/src/nyc_trees/nyc_trees/settings/base.py
+++ b/src/nyc_trees/nyc_trees/settings/base.py
@@ -47,7 +47,8 @@ DATABASES = {
         'USER': environ.get('NYC_TREES_DB_USER', 'nyc_trees'),
         'PASSWORD': environ.get('NYC_TREES_DB_PASSWORD', 'nyc_trees'),
         'HOST': environ.get('NYC_TREES_DB_HOST', 'localhost'),
-        'PORT': environ.get('NYC_TREES_DB_PORT', 5432)
+        'PORT': environ.get('NYC_TREES_DB_PORT', 5432),
+        'TEST_NAME': environ.get('NYC_TREES_TEST_DB_NAME', 'test_nyc_trees')
     }
 }
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -95,33 +95,32 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     services.vm.hostname = "services"
     services.vm.network "private_network", ip: SERVICES_IP
 
-
     services.vm.synced_folder ".", "/vagrant", disabled: true
 
     # Graphite Web
     services.vm.network "forwarded_port", {
       guest: 8080,
-      host: ENV.fetch("NYC_TREES_PORT_8080", 8080)
+      host: 8080
     }.merge(VAGRANT_NETWORK_OPTIONS)
     # Kibana
     services.vm.network "forwarded_port", {
       guest: 5601,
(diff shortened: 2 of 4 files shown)


hectcastro:
@azavea-bot rebuild

src/nyc_trees/nyc_trees/settings/base.py
```diff
@@ -47,7 +47,8 @@
         'USER': environ.get('NYC_TREES_DB_USER', 'nyc_trees'),
         'PASSWORD': environ.get('NYC_TREES_DB_PASSWORD', 'nyc_trees'),
         'HOST': environ.get('NYC_TREES_DB_HOST', 'localhost'),
-        'PORT': environ.get('NYC_TREES_DB_PORT', 5432)
+        'PORT': environ.get('NYC_TREES_DB_PORT', 5432),
+        'TEST_NAME': environ.get('NYC_TREES_TEST_DB_NAME', 'test_nyc_trees')
```
maurizi: Since we already specify a default in `defaults/main.yml`, consider removing the default value of `test_nyc_trees` here.

hectcastro: I think that the process of setting both defaults has their own purpose. The one in Ansible guards against a scenario where the environmental variable isn't set. The one inside the Django settings file guards against a scenario where you deploy the application without our Ansible setup (unlikely, but the other settings are setup this way).

kdeloach:
Rebuilding my VMs fixed the above issue.

There are 2 issues remaining:

  1. Incorrect nginx config in TEST mode
  1. Unable to browse website in TEST mode due to empty `ALLOWED_HOSTS` Django setting

hectcastro:
> There are 2 issues remaining:

> Incorrect nginx config in TEST mode

I believe this is the Gunicorn configuration file, right?

This should be fixed in https://github.com/azavea/nyc-trees/commit/91ade66037d8c110015c974aed3b0d85c1e44644.

> Unable to browse website in TEST mode due to empty ALLOWED_HOSTS Django setting

This should be fixed in https://github.com/azavea/nyc-trees/commit/4319ca1574b5bd37629a2ecef1d28b96b8fa81da. 

Both fixes I referenced are part of `feature/hmc/blue-green`, which will be up as a PR tomorrow or the day after. Since the changes in this PR didn't cause the issues, should I just leave the fixes in that branch?

kdeloach:
Yup, just recording known issues issues I ran across during testing in case other people run into the same problems.

kdeloach:
I tested this by provisioning 2 sets of nyc VMs that point to the same services VM and running the unit tests simultaneously.

VM 1:

    VAGRANT_ENV=TEST NYC_TREES_SERVICES_IP=33.33.33.30 v provision app

VM 2:

    VAGRANT_ENV=TEST NYC_TREES_SERVICES_IP=33.33.33.30 NYC_TREES_TEST_DB_NAME='foo_test_nyc_trees' v provision app

I confirmed that omitting `NYC_TREES_TEST_DB_NAME` on VM 2 results in unit test collisions and failures. I also confirmed that both the unit tests and UI tests work correctly when run in parallel.

kdeloach:
+1

hectcastro:
Great job reviewing this, @kdeloach.

PR #247, opened 2015-01-03
Add blockface fixture for testing purposes.

maurizi:
Some of the blockfaces from the 'blockface_live' table were MultiLineStrings,
so I changed the type of the field to accomodate that.

This fixture can be loaded via running:
    'manage.py loaddata apps/survey/fixtures/blockface.json'


maurizi:
@hectcastro I was initially planning to run `loaddata` during our Ansible provisioning, but I don't see anywhere where we are running `manage.py migrate`, and loading a fixture would have to come *after* creating/migrating all the tables.

Do we plan to run `manage.py migrate` during Ansible provisioning, and haven't gotten to it, or are we purposefully not including it in our provisioning steps?

hectcastro:
Unfortunately, migrations can't be applied via the Ansible provisioning process because that process occurs at AMI creation time. When the AMI is created, it doesn't have a connection to a database.

All of the steps around a migration should be encapsulated in a script that we can execute from an application server following a deploy. I started that script [here](https://github.com/azavea/nyc-trees/blob/feature/hmc/blue-green/scripts/syncdb.sh), but it hasn't been merged yet.

jwalgran:
I would like to add a ``load_dev_data`` management command to this pull. The inaugural function could truncate the tree, survey, and blackface tables and load in the included fixture. It would give us a nice foundation for interactive, reset-able user testing.

maurizi:
@jwalgran Seems simple and helpful, I'll add it.

maurizi:
@jwalgran I added the management command you suggested

jwalgran:
+1
These line up with OSM nicely.
screen shot 2015-01-05 at 12 09 52 pm

PR #248, opened 2015-01-05
Bump Django version to 1.7.2

hectcastro:
Bugfix release: https://docs.djangoproject.com/en/dev/releases/1.7.2/

### DIFF
--- a/src/nyc_trees/requirements/base.txt
+++ b/src/nyc_trees/requirements/base.txt
@@ -1,4 +1,4 @@
-Django==1.7
+Django>=1.7.2
 ipython==2.3.0
 psycopg2==2.5.4
 Pillow==2.6.1


steventlamb:
I think pinning to a patch number is discouraged in semver generally and with pypi specifically, and the preference order is as follows, descending:
1) Pin to major/minor (`==1.7`) and upgrade often
2) Constrain to major/minor/match minimum (`>=1.7.2`)
3) Pin to patch (`==1.7.2`)

I know we've been burned by loose dependencies with other package managers, but I've found pypi to be very safe for this type of thing. Consider (1) or (2).

steventlamb:
(expanding previous)
It's hard to see a case in which (1) would be better than (2), even if it is more technically correct. The advantage of (2) is that it will force pip to upgrade when it sees that django is less than `1.7.2`.

hectcastro:
Do you know if `pip` will upgrade a 1.7.1 install to 1.7.2 if something like `pip install -r requirements.txt` is run? I'm mostly worried about cutting a new AMI that installs 1.7.2, but everyone else if developing locally against 1.7.1.

steventlamb:
I tried leaving `==1.7` as is and it turns out neither `pip install -r base.txt` nor `pip install -r base.txt --upgrade` will bump it up. I switched it to `>=1.7.2` and ran `pip install -r base.txt` and I didn't need to include `--upgrade` to force the upgrade.

So it looks like (1) is not viable without forcibly destroying environments, but (2) works without friction.

hectcastro:
@steventlamb I changed the version constraint to `>=` and confirmed the behavior you described with an Ansible provision of `1.7` upgrading to `1.7.2`.

steventlamb:
+1

PR #249, opened 2015-01-05
Add progress map page

RickMohr:
* Using OpenStreetMap data for now
* Initial extent includes all of New York City
* The map is not currently 100% height, pending styling fixes from the design team
* `map.js` is currently very simple, and will presumably sprout more options

### DIFF
--- a/src/nyc_trees/js/src/map.js
+++ b/src/nyc_trees/js/src/map.js
@@ -1,0 +1,29 @@
+"use strict";
+
+var $ = require('jquery'),
+    L = require('leaflet');
+
+module.exports = {
+    create: create
+};
+
+function create(options) {
+    options = $.extend({
+        domId: 'map'
+    }, options);
+
+    var map = L.map(options.domId),
+        osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+        osmAttrib = 'Map data © <a href="http://openstreetmap.org">OpenStreetMap</a> contributors',
+        osm = new L.TileLayer(osmUrl, {attribution: osmAttrib});
+
+    map.addLayer(osm);
+
+    // New York City bounds from http://www.mapdevelopers.com/geocode_bounding_box.php
+    map.fitBounds([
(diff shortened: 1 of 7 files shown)


src/nyc_trees/js/src/map.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+var L = require('leaflet');
+
+module.exports = {
+    create: create
+};
+
+function create() {
```
maurizi: Consider having the `create` function take the dom id of the map as an argument, to avoid hardcoding the dom element in a non-entry-point module (and in features coming very soon, to allow for more than one map on a page).

src/nyc_trees/js/src/map.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+var L = require('leaflet');
+
+module.exports = {
+    create: create
+};
```
maurizi: Consider making exports be the `create` function itself:
```javascript
module.exports = create;
```

maurizi:
Did you double-check that the changes in `npm-shrinkwrap.json` only affected the new Leaflet dependency and the `from` line of everything else?

We had to change at least on transitive dependencies version to avoid hard to reproduce Gulp issues, so we should make sure that we don't reintroduce that.

maurizi:
+1

RickMohr:
@maurizi, the changes in `npm-shrinkwrap.json` affected no `resolved` lines, so I think we are ok.

    $ git diff -U0 9d46fc01a8 | grep resolved
    +      "resolved": "https://registry.npmjs.org/leaflet/-/leaflet-0.7.3.tgz"

src/nyc_trees/js/src/map.js
RickMohr: @maurizi, I exported `create` this way because the module will probably have additional exports.

src/nyc_trees/js/src/map.js
RickMohr: Changed.

PR #250, opened 2015-01-05
Add a /geocode JSON API endpoint

jwalgran:
This JSON API will be used by the client-side map component to locate addresses. We have opted to use the python-omgeo library to have flexibility in our selection of a geocoder service (or services, omgeo allows setting up multiple services in a fall back arrangement).

I added a single unit test, which is actually an integration test, since it makes a real geocode request to the configured service. The majority of the functionality is wrapped up in python-omgeo, so I opted to not mock it out and leave only a low-value test of the response marshaling.


RickMohr:
Otherwise looks great.

jwalgran:
@RickMohr I made a sizable refactor and implemented your suggestions. Please take another look.

RickMohr:
Looks excellent +1

PR #252, opened 2015-01-05
Add "geolocate" button to Leaflet control container

RickMohr:


src/nyc_trees/js/src/map.js
```diff
@@ -3,6 +3,10 @@
 var $ = require('jquery'),
     L = require('leaflet');
 
+var zoom = {
+    NEIGHBORHOOD: 16
```
jwalgran: I like having zoom levels with proper names.

jwalgran:
The map successfully found my location and centered the map on Mesa, AZ. I think we should take a page from OTM and not zoom to the user's location if they are outside NYC.

jwalgran:
Definitely ok to open an issue for restricting the geolocation to a bounding box rather than address it now.

jwalgran:
On Chrome, Safari, and Firefox the "minus" zoom button is offscreen, and the entire toolbar is only visible if I maximize my browser window.

screen shot 2015-01-05 at 2 25 43 pm

RickMohr:
@jwalgran, I don't see this problem in Chrome or Firefox on Windows.

Did you scroll all the way down?

jwalgran:
Scrolling isn't really possible because the slippy map fills the entire window

RickMohr:
The slippy map is currently `height:800px` pending css work. For me in Firefox that means its bottom including the minus button is scrolled off the bottom and there's a vertical scrollbar.

jwalgran:
+1

OS X like to hide scrollbars to keep the window clutter free. Sometimes they are too hidden.

PR #257, opened 2015-01-06
Add legend to progress map

RickMohr:
* Change `home` app to routes pattern
* Add `legend.html` template with bootstrap modal
* Legend entries are context-driven, with CSS class to define color
* Add "legend" button to map


kdeloach:
Unable to view the leaflet controls in lower left corner on some mobile devices. Otherwise, +1.

RickMohr:
@kdeloach, that's because the map height is currently fixed to 800px, waiting on CSS work. Since you are the second person to be confused by that I've changed it to 400px.

Issue #256, opened 2015-01-06
Debugserver doesn't reload when there are changes


kdeloach:
The gunicorn debuserver used to reload when there were Python script changes. Now, when provisioned as dev, I have to reload it manually.

This may be fixed by 91ade66037d8c110015c974aed3b0d85c1e44644.

maurizi:
@kdeloach Is this fixed?

steventlamb:
I had this issue for a spell and now it's fixed.

kdeloach:
Verified fixed.

Issue #242, opened 2014-12-31
Create application server IAM roles


hectcastro:
Create an IAM role for the staging and production AWS accounts named `AppServerInstanceProfile` with the following policy (we don't have IAM write access):

```javascript
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "ses:SendEmail",
        "ses:SendRawEmail"
      ],
      "Resource": "*"
    }
  ]
}
```

(This allows us to send e-mail via SES via EC2 instances.)

---

1. Click "Roles"

![screen_shot_2014-12-31_at_15_22_17](https://cloud.githubusercontent.com/assets/43639/5590401/3211bb1a-9101-11e4-94f2-1a50e550d411.png)

2. Create New Role

![screen_shot_2014-12-31_at_15_22_24](https://cloud.githubusercontent.com/assets/43639/5590403/50c6a3ea-9101-11e4-9066-7c2f51933aab.png)

3. Name Role

![screen shot 2014-12-31 at 15 22 59](https://cloud.githubusercontent.com/assets/43639/5590407/5e7641a8-9101-11e4-86da-72e4faa7350f.png)

4. Select Amazon EC2

![screen_shot_2014-12-31_at_15_23_05](https://cloud.githubusercontent.com/assets/43639/5590413/7a15bb3c-9101-11e4-9310-2e942e16bf75.png)

5. Set Policy

![screen shot 2014-12-31 at 15 23 12](https://cloud.githubusercontent.com/assets/43639/5590415/8c1f48e8-9101-11e4-8019-fc17dd08260e.png)

![screen shot 2014-12-31 at 15 23 24](https://cloud.githubusercontent.com/assets/43639/5590418/956364a2-9101-11e4-8311-f28af2ef3366.png)

6. Finalize

![screen shot 2014-12-31 at 15 23 28](https://cloud.githubusercontent.com/assets/43639/5590420/9e1249a6-9101-11e4-84ff-54c5d4d4c7ef.png)

hectcastro:
@dboyer Just tested this and it works on both the staging and production accounts.

dboyer:
Great. Thanks for handling that.

PR #258, opened 2015-01-06
Add support for blue/green deployment integration via Jenkins

hectcastro:
I split this pull request into a few commits that seemed to group logically, but were still all necessary to support blue/green deployments in some way. Each commit includes a more detailed description of what it includes and why.

The goal with this PR is to provide a starting point for deployment integration via Jenkins.

Resolves: #256


deployment/troposphere/app_template.py
```diff
@@ -26,13 +37,13 @@
...
-    Description='Physical resource ID of an AWS::IAM::Role for the '
-                'application servers'
-))
+# app_server_instance_profile_param = t.add_parameter(Parameter(
+#     'AppServerInstanceProfile', Type='String',
+#     Default=
+#     'arn:aws:iam::900325299081:instance-profile/AppServerInstanceProfile',
+#     Description='Physical resource ID of an AWS::IAM::Role for the '
+#                 'application servers'
+# ))
```
maurizi: Is there a reason to check in this commented code?

deployment/troposphere/app_template.py
```diff
@@ -41,29 +52,118 @@
...
+))
+
+#
+# Auto Scaling Group Resources
 #
 app_server_launch_config = t.add_resource(asg.LaunchConfiguration(
     'lcAppServer',
     ImageId=Ref(app_server_ami_param),
-    IamInstanceProfile=Ref(app_server_instance_profile_param),
+    # IamInstanceProfile=Ref(app_server_instance_profile_param),
```
maurizi: Is there a reason to check in this commented code?

deployment/troposphere/app_template.py
hectcastro: This is pending until NYC can create IAM roles for us to assign to the launch configurations. We don't have write permission on IAM in the AWS accounts they gave us.

See also: https://github.com/azavea/nyc-trees/issues/242

maurizi: Noted

README.md
```diff
@@ -183,11 +184,18 @@ After the private hosted zones exist, create the data store stack:
 $ make data-store-stack
 ```
 
-#### Launch the Tile and Application Servers
+#### Launch the Application Servers
 
-Now that the VPC and data store stacks are setup, we can launch the tile and applications server stacks, which will make use of the most recent AMIs available:
+Now that the VPC and data store stacks are setup, we can launch the application server stack, which will make use of the most recent AMIs available:
```
maurizi: Are the tile servers removed here because work on the tiler is in-progress?  Or does application encompass the entire redeployable part of the stack? (e.g application == tiler + django app + monitoring)?

hectcastro: Removed because in-progress.

maurizi:
Overall this looks good to me, though I haven't tried to run it.

It's nice to see that this is all encapsulated in a few top-level `make` commands.

hectcastro:
For the record, console output from a Jenkins job that replaced a blue stack with a green one: http://civicci01.internal.azavea.com/job/nyc-trees-staging-deployment/7/console

PR #259, opened 2015-01-06
Force square group images

jfrankl:
This will force the group icons to have the same height and width.

Downside: forcing this using css means we need to specify a height and width. We'll need to use media queries to adjust this based on screen size, which is a bit of a pain, but not bad.

In terms of style, I suspect that these will change a bit once we are looking at some real logos, but this will prevent the "oval effect" for now.

### DIFF
--- a/src/nyc_trees/apps/users/templates/groups/detail.html
+++ b/src/nyc_trees/apps/users/templates/groups/detail.html
@@ -15,7 +15,9 @@
                         <div class="col-xs-4">
                             {# TODO: Show a default image if group.image is null #}
                             {% if group.image %}
-                            <img class="group-branding" src="{{group.image.url}}"/>
+                            <div class="group-branding">
+                                <img src="{{group.image.url}}"/>
+                            </div>
                             {% endif %}
                         </div>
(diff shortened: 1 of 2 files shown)


RickMohr:
The current design requires people to provide a square logo that looks good clipped in a circle. Even though it's a cool design it seems to me like asking for trouble. Creating logos is not easy and people will want to use existing ones, and will then be frustrated or complain when they can't make them look right. I think we should consider a different design that can accommodate rectangular logos with a reasonable range of aspect ratios.

jfrankl:
Rick, I hear what you're saying, but I think the square shape will help to create a consistent look across the site and is also a familiar concept to internet users.

Both twitter and facebook impose a square shape on logos, so I believe that organizations are somewhat familiar with this concept—they'll be able to use that same logo here. 

If we don't impose some dimensions, it gets really challenging to think about the design, in a way where you can make it look acceptable with a horizontally oriented logo, but then a square or a vertical logo completely throws off the page.

I'm open to reopening this later in the process and after the final design comps have been reviewed by Parks.

PR #262, opened 2015-01-06
Topic/group detail edit event lists

steventlamb:


jwalgran:
+1
I pulled this down and was able to see the current and past events on the group edit page and the "load more..." button works.

PR #266, opened 2015-01-07
Add support for log shipping to Logstash via Beaver

hectcastro:
This changeset adds support for Nginx and Django log shipping to the shared Logstash instance via Beaver. Nginx is already configured to emit its logs in JSON format. Beaver simply slurps them up and passes them along to Logstash. Django logs are shipped line-by-line unless they are a traceback, in that case the entire traceback is tucked into a single message.

screen shot 2015-01-07 at 13 36 32

screen shot 2015-01-07 at 15 15 40


kdeloach:
Works great +1

deployment/ansible/roles/nyc-trees.beaver/templates/django.conf.j2
```diff
@@ -0,0 +1,5 @@
+[/var/log/{upstart/nyc-trees-app.log,nyc-trees-app.log}]
```
jwalgran: For my education, is the curly brace section indicating that we want to pull from both files?

hectcastro: Yep.

PR #268, opened 2015-01-07
Add tiler w/ basic blockface layer

maurizi:
There is some opportunity for refactoring here, particularly with the deployment tasks, but I wanted to get this up for feedback sooner rather than later.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -133,6 +133,18 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     tiler.vm.synced_folder ".", "/vagrant", disabled: true
 
+    if Vagrant::Util::Platform.windows? || Vagrant::Util::Platform.cygwin?
+      tiler.vm.synced_folder "src/tiler", "/opt/tiler/", type: "rsync", rsync__exclude: ["node_modules/"]
+    else
+      tiler.vm.synced_folder "src/tiler", "/opt/tiler/"
+    end
+
+    # Windshaft via Nginx
+    tiler.vm.network "forwarded_port", {
+      guest: 80,
+      host: 7000
+    }.merge(VAGRANT_NETWORK_OPTIONS)
+
     tiler.vm.provision "ansible" do |ansible|
       ansible.playbook = "deployment/ansible/tile-servers.yml"
       ansible.groups = ANSIBLE_GROUPS.merge(ANSIBLE_ENV_GROUPS)
--- a/deployment/ansible/roles/nyc-trees.app/defaults/main.yml
+++ b/deployment/ansible/roles/nyc-trees.app/defaults/main.yml
@@ -2,6 +2,7 @@
 app_home: /opt/app
 app_config_home: /etc/nyc-trees.d/env
 app_deploy_branch: "master"
+
 app_config:
   - { file: "NYC_TREES_DB_NAME", content: "{{ postgresql_database }}" }
   - { file: "NYC_TREES_DB_USER", content: "{{ postgresql_username }}" }
@@ -15,6 +16,7 @@ app_config:
   - { file: "DJANGO_MEDIA_ROOT", content: "{{ app_media_root }}" }
   - { file: "DJANGO_POSTGIS_VERSION", content: "{{ app_postgis_version }}" }
   - { file: "DJANGO_SECRET_KEY", content: "{{ app_secret_key }}" }
+  - { file: "TILER_HOST", content: "//{{ tiler_host }}" }
 
 app_sauce_labs_config:
   - { file: "SELENIUM_HOST", content: "{{ lookup('env', 'SELENIUM_HOST') }}" }
--- a/deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
+++ b/deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
@@ -3,4 +3,6 @@ dependencies:
   - { role: "azavea.nginx", nginx_delete_default_site: True }
   - { role: "azavea.mapnik" }
   - { role: "azavea.daemontools" }
+  - { role: "azavea.build-essential" }
+  - { role: "azavea.swapfile", when: developing_or_testing }
   - { role: "azavea.nodejs" }
--- a/deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
+++ b/deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
@@ -1,0 +1,6 @@
+---
+dependencies:
+  - { role: "azavea.nginx", nginx_delete_default_site: True }
+  - { role: "azavea.mapnik" }
+  - { role: "azavea.daemontools" }
+  - { role: "azavea.nodejs" }
--- a/deployment/ansible/roles/nyc-trees.tiler/tasks/app.yml
+++ b/deployment/ansible/roles/nyc-trees.tiler/tasks/app.yml
@@ -1,0 +1,21 @@
+---
+- name: Clone project source
+  git: repo=https://github.com/azavea/nyc-trees.git
+       dest=/opt/nyc-trees
+       version="{{ tiler_deploy_branch }}"
+  when: packing
+
+- name: Ensure that tiler_home exists
+  file: path="{{ tiler_home }}"
+        state=directory
+  when: packing
+
+- name: Synchronize tiler application into tiler_home
+  synchronize: archive=no
+               checksum=yes
+               compress=no
+               recursive=yes
+               set_remote_user=no
+               src=/opt/nyc-trees/src/tiler/
+               dest="{{ tiler_home }}/"
+  when: packing
--- a/deployment/ansible/roles/nyc-trees.tiler/tasks/dependencies.yml
+++ b/deployment/ansible/roles/nyc-trees.tiler/tasks/dependencies.yml
@@ -1,0 +1,5 @@
+---
+- name: Install tiler javascript dependencies
+  command: npm install --unsafe-perm
+  args:
+    chdir: "{{ tiler_home }}"
--- a/deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
+++ b/deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
@@ -12,4 +12,4 @@ setuid nyc-trees
 chdir {{ tiler_home }}
 env HOME="/var/lib/nyc-trees"
 
-exec envdir /etc/nyc-trees.d/env node server.js > {{ tiler_log }}
+exec envdir /etc/nyc-trees.d/env node server.js >> {{ tiler_log }} 2>&1
--- a/deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
+++ b/deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
@@ -1,0 +1,15 @@
+description	"nyc-trees-tiler"
+
+{% if 'development' in group_names or 'test' in group_names -%}
+start on (vagrant-mounted)
+{% else %}
+start on (local-filesystems and net-device-up IFACE!=lo)
+{% endif %}
+stop on shutdown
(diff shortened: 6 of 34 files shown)


maurizi:
@azavea-bot rebuild

maurizi:
I believe this is failing to build on Jenkins because the tiler VM is running out of memory when compiling Windshaft dependencies.  I'm working on adding a swapfile to the VM to prevent this.

src/tiler/server.js
```diff
@@ -0,0 +1,54 @@
...
+        base_url_notable: '/:cache_buster/:dbname',
+
+        grainstore: {
+            datasource: {
+                user: dbUser,
+                password: dbPassword,
+                host: dbHost,
+                port: dbPort,
+                geometry_field: 'geom',
+                srid: 4326
```
hectcastro: Is it OK to hardcode this to 4326?

src/nyc_trees/nyc_trees/settings/base.py
```diff
@@ -286,3 +286,5 @@
 # If geocoding a string produces no results, this string will be
 # appended for a second attempt.
 GEOCODE_FALLBACK_SUFFIX = ', New York, NY'
+
+TILER_URL = environ['TILER_HOST']
```
hectcastro: All of the other environmental variables pulled in this settings file use `environ.get` with some default. `localhost` seems like the best option for a default value.

src/nyc_trees/apps/core/views.py
```diff
@@ -0,0 +1,35 @@
...
+from apps.survey.models import Blockface, Survey
+
+
+def js_settings(request):
+    blockface_updated_at = _get_last_updated_datetime(Blockface)
+    survey_updated_at = _get_last_updated_datetime(Survey)
+
+    max_timestamp = lambda *datetimes: timegm(max(*datetimes).utctimetuple())
+
+    return render_to_response('core/settings.js', {
```
hectcastro: Is this something to provide settings dynamically to the tiler? Trying to make sense of what this code is responsible for.

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
```diff
@@ -0,0 +1,11 @@
...
+
+start on ({{ "vagrant-mounted" if developing_or_testing else "local-filesystems and net-device-up IFACE!=lo" }})
+stop on shutdown
+
+respawn
+setuid nyc-trees
+chdir {{ tiler_home }}
+env HOME="/var/lib/nyc-trees"
+
+exec envdir /etc/nyc-trees.d/env node server.js > {{ tiler_log }}
```
hectcastro: This should be changed to the following so that `STDERR` and `STDOUT` are directed:

```
exec envdir /etc/nyc-trees.d/env node server.js >> {{ tiler_log }} 2>&1
```

Also need the `>>` so that we append vs. truncate the logs.

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
```diff
@@ -0,0 +1,11 @@
+description	"nyc-trees-tiler"
+
+start on ({{ "vagrant-mounted" if developing_or_testing else "local-filesystems and net-device-up IFACE!=lo" }})
+stop on shutdown
+
+respawn
+setuid nyc-trees
+chdir {{ tiler_home }}
+env HOME="/var/lib/nyc-trees"
```
hectcastro: Is this necessary with the `setuid` stanza?

deployment/ansible/roles/nyc-trees.tiler/tasks/dependencies.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Install tiler javascript dependencies
+  command: npm install
```
hectcastro: It looks like NPM drops privileges to use `gyp` during this process, which generated a bunch of warnings and changes the directories `gyp` uses. Adding `--unsafe-perm` here alleviates that.

deployment/ansible/roles/nyc-trees.tiler/tasks/app.yml
```diff
@@ -0,0 +1,21 @@
...
+       dest=/opt/nyc-trees
+       version="{{ tiler_deploy_branch }}"
+  when: packing
+
+- name: Ensure that tiler_home exists
+  file: path="{{ tiler_home }}"
+        state=directory
+  when: packing
+
+- name: Synchronize Tiler application into tiler_home
```
hectcastro: Other references to the tiler inside a `name` attribute are lowercase. Consider making this lowercase as well for consistency.

deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+dependencies:
+  - { role: "azavea.nginx", nginx_delete_default_site: True }
+  - { role: "azavea.mapnik" }
+  - { role: "azavea.daemontools" }
+  - { role: "azavea.nodejs" }
```
hectcastro: `g++` is missing during the Node.js dependency installation process. Need to add the `azavea.build-essential` role.

deployment/ansible/roles/nyc-trees.app/defaults/main.yml
```diff
@@ -2,6 +2,9 @@
 app_home: /opt/app
 app_config_home: /etc/nyc-trees.d/env
 app_deploy_branch: "master"
+
+tiler_host: localhost
```
hectcastro: I think that this variable should be pulled out into `groups_vars` like in https://github.com/azavea/nyc-trees/pull/270. For the Packer `group_vars` file, we can set it to `tiler.service.nyc-trees.internal`.

Vagrantfile
```diff
@@ -61,6 +61,8 @@ else
 end
 
 SERVICES_IP = ENV.fetch("NYC_TREES_SERVICES_IP", "33.33.33.30")
+TILER_PORT = ENV.fetch("NYC_TREES_PORT_7000", 7000)
```
hectcastro: After some local testing, we should be able to undo most of the changes in here and use the host-only IP address of the tiler VM (without a port) as `tiler_host` instead of `localhost:{{ tiler_port }}`.

Keeping the port forwarding rule seems file to convenience.

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
maurizi: Yes, it is defintely necessary.

src/nyc_trees/apps/core/views.py
maurizi: It's for providing dynamic settings (the tiler URL) to the browser JS code.

src/tiler/server.js
maurizi: I think so.  At the very least, it's as valid as hardcoding any other aspect of our DB schema, like table names, columns names, SQL statements, etc.

src/nyc_trees/nyc_trees/settings/base.py
maurizi: I struggled with this.  I don't like providing default values in 2 places, and there is already a default in the Ansible playbooks.  I can add it here too though.  (FYI, the default would be `localhost:7000`, not `localhost`)

deployment/ansible/roles/nyc-trees.tiler/meta/main.yml
maurizi: Yeah, it builds without `g++` in place, but it definitely seems to prefer `g++` over whatever the fallback compiler is.  Easy enough to add that role in, will do.

deployment/ansible/roles/nyc-trees.app/defaults/main.yml
maurizi: Will do

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
hectcastro: Confirmed. Thanks, Node.js.

src/nyc_trees/nyc_trees/settings/base.py
hectcastro: I get the dislike for setting defaults in two different places, but in this case I think it is necessary because I view the Ansible world and the Django settings world as separate. Although unlikely, someone could deploy this app without our provisioning steps. In that case, they wouldn't get a default for this variable if they missed setting the environmental variable.

maurizi:
@hectcastro @RickMohr This has been updated and all comments have been addressed.

src/nyc_trees/apps/core/urls.py
```diff
@@ -4,16 +4,11 @@
...
 
 
-# These URLs have the prefix 'quiz/'
 urlpatterns = patterns(
     '',
-    url(r'^js-settings/$',
-        route(GET=TemplateView.as_view(template_name='core/settings.js',
-                                       content_type='application/javascript')),
-        name='js_settings'),
+    url(r'^js-settings/$', js_settings, name='js_settings'),
```
RickMohr: We are not actually hitting this URL, right? Do we know we will?

maurizi: We've been using it for a while: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/core/templates/base.html#L59

RickMohr: I see. I thought we had decided to save the round trip by inlining it, but that's probably not a big issue since it will load in parallel with the JS bundle.

When we make a partial to load the page-specific JavaScript we should move all JavaScript loading there so we don't load JavaScript unnecessarily on JavaScript-less pages.

RickMohr:
+1

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
```diff
@@ -0,0 +1,11 @@
+description	"nyc-trees-tiler"
+
+start on ({{ "vagrant-mounted" if developing_or_testing else "local-filesystems and net-device-up IFACE!=lo" }})
```
hectcastro: This template conditional doesn't evaluate properly. If I change it to something that should trigger the `else`, it keeps `vagrant-mounted`. See the fix I used for the application Upstart script in https://github.com/azavea/nyc-trees/commit/cfb558b1e72eb0cd1d7341127a2bd94fc5afad1b.

src/nyc_trees/apps/core/views.py
```diff
@@ -0,0 +1,35 @@
...
+        "db_name": settings.DATABASES['default']['NAME'],
+        "cache_buster": {
+            "progress": max_timestamp(blockface_updated_at, survey_updated_at)
+        }
+    })
+
+
+def _get_last_updated_datetime(Model):
+    try:
+        return Model.objects.latest('updated_at').updated_at
```
jwalgran: We should add DB indexes on ``updated_at`` for model we will be used in a tiler query so we can generate the cache buster as quickly as possible.

deployment/ansible/roles/nyc-trees.tiler/templates/upstart-nyc-trees-tiler.conf.j2
maurizi: Done

src/nyc_trees/apps/core/views.py
maurizi: I'm not going to do this right now, but I added #275 so we don't forget

hectcastro:
:+1: 

Not too worried about the duplication between the `nyc-trees.app` and `nyc-trees.tiler` roles for now. Overall, this looks really good.

After merging, I plan to follow-up with support for log shipping and connecting Windshaft to StatsD. Then we can move forward toward getting the tiler into the staging environment.

PR #269, opened 2015-01-07
Event check-in page

kdeloach:
Changes:

1. New button on event detail page to check-in users (for group admins only)
1. Check-in page which lists RSVPed users and a button to toggle check-in status
1. RSVP increase limit button on check-in page


src/nyc_trees/apps/event/models.py
```diff
@@ -57,6 +57,16 @@ def training_summary(self):
     def has_space_available(self):
         return self.eventregistration_set.count() < self.max_attendees
 
+    @property
+    def starting_soon(self):
+        # TODO: Implement
+        return True
```
maurizi: I don't think it hurts to implement this now.  The time period of "soon" may not be set, but the logic of checking the `ends_at` date won't be changing

src/nyc_trees/apps/event/views.py
```diff
@@ -209,18 +209,62 @@ def start_event_map_print_job(request, event_slug):
...
+        rsvp.did_attend = did_attend
+        rsvp.save()
+    except EventRegistration.DoesNotExist:
+        return HttpResponseForbidden()
+    return {
+        'group': request.group,
+        'event': event,
+        'user': rsvp.user,
+        'did_attend': did_attend
+    }
```
maurizi: I looks like the only difference between `check_in_user_to_event` and `un_check_in_user_to_event` is the value of `did_attend`.  Consider refactoring this like so:
```python
def check_in_user_to_event(request, event_slug, username):
   return _toggle_event_checkin(request, event_slug, username, True)

def check_in_user_to_event(request, event_slug, username):
   return _toggle_event_checkin(request, event_slug, username, False)

@transaction.atomic
def _toggle_event_checkin(request, event_slug, username, did_attend):
    event = get_object_or_404(Event, group=request.group, slug=event_slug)
    try:
        rsvp = EventRegistration.objects.get(event=event,
                                             user__username=username)
        rsvp.did_attend = did_attend
        rsvp.save()
    except EventRegistration.DoesNotExist:
        return HttpResponseForbidden()
    return {
        'group': request.group,
        'event': event,
        'user': rsvp.user,
        'did_attend': did_attend
    }
```

src/nyc_trees/apps/event/models.py
kdeloach: Will do.

src/nyc_trees/apps/event/views.py
kdeloach: Good suggestion. Now that I think about it, this could just be 1 view with a case around the HTTP method.

src/nyc_trees/apps/event/templates/event/admin_checkin.html
```diff
@@ -1,5 +1,67 @@
...
+    <div class="row">
+        <div class="col-sm-12">
+            <p>Nobody has RSVPd, yet!</p>
+        </div>
+    </div>
+    {% endif %}
+
+    <div class="row">
+        <a href="javascript:void(0)"
+           id="increase_rsvp_limit"
```
jwalgran: Our convention is to have all ids and classes use ``-`` instead of ``_``

src/nyc_trees/apps/event/templates/event/admin_checkin.html
```diff
@@ -1,5 +1,67 @@
...
+           data-verb="POST"
+           data-url="{% url 'increase_rsvp_limit' group_slug=group.slug event_slug=event.slug %}"
+           data-text="Increase RSVP limit"
+           class="col-sm-12 btn btn-primary">Increase RSVP limit ({{ event.max_attendees }})</a>
```
jwalgran: It may be confusing to show the current limit on the button. "Am I increasing the limit by 30 people?" I do understand the need to provide some feedback to the user indicating that the button actually did something. Double check with @jfrankl to see if this is the best way to show that feedback.

src/nyc_trees/apps/event/templates/event/partials/checkin_button.html
```diff
@@ -0,0 +1,21 @@
+{% comment %}
+
+The following variables must be in scope:
```
jwalgran: I like this comment listing all the context requirements of the template

src/nyc_trees/apps/event/views.py
jwalgran: Naked boolean function arguments are always mysterious. It would be a little clearer if we use a kwarg. The name "toggle" is a little misleading, because it does not flip the existing value.

```python
_update_event_checkin(request, event_slug, username, did_attend=False)

jwalgran:
I will pull this down and test it out tomorrow morning (1/9)

src/nyc_trees/apps/event/templates/event/admin_checkin.html
kdeloach: Thanks, looks like a copy & paste oversight.

src/nyc_trees/apps/event/templates/event/admin_checkin.html
kdeloach: Agreed, here's what it looks like for reference:
image

src/nyc_trees/apps/event/views.py
kdeloach: This diff is outdated.

jwalgran:
+1
Works as advertised

src/nyc_trees/apps/event/models.py
kdeloach: Doing this in a separate PR.

PR #272, opened 2015-01-08
Enable Django flatpages for training material

RickMohr:
To test:
* Run migrations
* In the Django admin, click "Add" next to "Flat Pages"
* Enter `/test/` for the `URL`, and some HTML content.
* Choose the only site, probably labeled `example.com`. (The name doesn't matter, but in production we should set it to the actual site name so users aren't confused.)
* Browse to the page you just created, at `/training/test`

### DIFF
--- a/src/nyc_trees/apps/home/templates/flatpages/default.html
+++ b/src/nyc_trees/apps/home/templates/flatpages/default.html
@@ -1,0 +1,5 @@
+{% extends "base.html" %}
+
+{% block content %}
+    {{ flatpage.content }}
+{% endblock content %}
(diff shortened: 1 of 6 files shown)


src/nyc_trees/apps/home/templates/flatpages/default.html
```diff
@@ -0,0 +1,5 @@
+{% extends "base.html" %}
+
+{% block content %}
```
steventlamb: I put some pretty basic text and it was smushed up against the navbar. It might be useful to add some mandatory padding so that not every flatpage has to do their own padding. Probably makes sense to play with it more when we have some data though.

steventlamb:
+1 simple and works. I pulled it down and tested it as described.

src/nyc_trees/apps/home/templates/flatpages/default.html
RickMohr: Seems like something to address, with design input, when we are focusing on training materials.

PR #273, opened 2015-01-08
Add groups_to_follow page

steventlamb:


steventlamb:
@jfrankl Styling is minimal, is that acceptable?

screenshot from 2015-01-08 17 07 52

jfrankl:
Instead of having this template extend base.html could you have it extend
src/nyc_trees/apps/core/templates/base_two_frames.html?
Here's an example of how base_two_frames.html works:

https://github.com/azavea/nyc-trees/blob/9239c9af7638768a5d5b53f6c3b938c818a516c0/src/nyc_trees/apps/users/templates/groups/settings.html

The aside block would contain the page title and the description beneath
it. The main block would contain the rest.

On Thu, Jan 8, 2015 at 5:08 PM, Steve Lamb <notifications@github.com> wrote:

> @jfrankl <https://github.com/jfrankl> Styling is minimal, is that
> acceptable?
>
> [image: screenshot from 2015-01-08 17 07 52]
> <https://cloud.githubusercontent.com/assets/1699455/5671808/ebe0f896-9758-11e4-877d-c4bd881fb0ae.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/azavea/nyc-trees/pull/273#issuecomment-69258799>.
>



-- 
Jeff Frankl, UI/UX Designer

Azavea  |  340 N 12th St, Suite 402, Philadelphia, PA
Web azavea.com  |  Blog azavea.com/blogs  |  Twitter @azavea
<http://twitter.com>

src/nyc_trees/apps/home/routes.py
```diff
@@ -19,3 +21,7 @@
 
 training_list_page = route(GET=do(render_template('home/training.html'),
                                   v.training_list_page))
+
+groups_to_follow = route(GET=do(render_template('home/groups_to_follow.html'),
+                                update({'chunk_size': 2}),
```
kdeloach: Consider moving `chunk_size` to the view. They contain data needed to render the templates so I think that's where most people would expect to find this.

src/nyc_trees/js/src/group_list.js
```diff
@@ -1,4 +1,41 @@
 "use strict";
 
+var $ = require('jquery');
+
+var attrs = {name: 'data-group-name',
+             description: 'data-group-description'},
+    dom = {searchInput: '[data-class="group-search"]',
```
kdeloach: `data-class` is an interesting idea.

src/nyc_trees/js/src/group_list.js
```diff
@@ -1,4 +1,41 @@
...
+
+$(dom.searchInput).on('input', function (e) {
+    var query = $(e.target).val();
+
+    $(dom.viewAll).hide();
+
+    if (query === '') {
+        $groups.show();
+    } else {
+        $.each($groups, function(__, group) {
```
kdeloach: Consider displaying a message if there were no search results.

src/nyc_trees/apps/home/urls.py
```diff
@@ -17,7 +17,9 @@
     url(r'^jobs/(?P<job_id>\d+)/$', r.retrieve_job_status,
         name='retrieve_job_status'),
 
-    url(r'^training/$', r.training_list_page),
+    url(r'^training/$', r.training_list_page, name="training_list_page"),
+
+    url(r'^training/groups_to_follow/$', r.groups_to_follow),
```
kdeloach: This may be a flatpages thing but you get a 404 if the trailing slash is omitted. Don't know what we can do about that.

kdeloach:
+1

src/nyc_trees/apps/home/urls.py
steventlamb: This is not a flatpages thing, we're hijacking certain `training/` urls to make custom, regular django pages and exposing the rest as flat pages.

IIRC, the django convention is to require all urls to end in a trailing slash, and use the `APPEND_SLASH` middleware to force requests to have one. You verified that a `404` happened?

https://docs.djangoproject.com/en/1.7/ref/settings/#append-slash

src/nyc_trees/js/src/group_list.js
steventlamb: Great idea.

src/nyc_trees/js/src/group_list.js
steventlamb: I think it's a best practice to only use `id` and `data-*` attributes with behaviors and never `class` to keep interop with designers smooth. I like to take this further and avoid using `id` when performance is not a consideration. I'm not sure if the latter is advisable.

src/nyc_trees/apps/home/urls.py
kdeloach: It appends `/` for groups, events, etc. pages but not the `training/groups_to_follow` page. I can test again later to verify.

src/nyc_trees/apps/home/routes.py
steventlamb: `chunk_size` is an optional param, and the vanilla `group_list` view (from a different app) does not provide it. In this case we want to reuse everything about that view by importing it but stick something else in the context, which I think is done most easily in the more powerful route layer. Does that make sense?

src/nyc_trees/apps/home/urls.py
steventlamb: No need in the meantime. I'll attempt to reproduce and fix and let you know if I have any trouble.

src/nyc_trees/apps/home/routes.py
kdeloach: I think creating a new view would still be simpler than the `update` decorator but I would get another opinion.

maurizi: I agree with @kdeloach here.  Adding a new view that calls `group_list` and then adds `chunk_size` to the context should be pretty small and much easier to grok

steventlamb: Noted. It will be done.

kdeloach:
For the search, I recall that you should be able to search for text associated with a group. Was that supposed to be part of this task or was that carded separately?

src/nyc_trees/js/src/group_list.js
```diff
@@ -1,4 +1,41 @@
...
+            var $group = $(group),
+                name = $group.attr(attrs.name),
+                description = $group.attr(attrs.description);
+
+            if (name.indexOf(query) !== -1 ||
```
jwalgran: We should lower or upper case all our strings so our comparison is case insensitive

steventlamb:
Not sure what you mean by text associated. In this changeset, you can search by name or by description.

jwalgran:
There is an "affiliation" field that does not yet exist on the group model (a late addition to the project, wireframe pg. 39) that will also need to be part of the search index. I'll add an issue so we don't forget.

src/nyc_trees/js/src/group_list.js
steventlamb: This is already done in the template:
https://github.com/azavea/nyc-trees/pull/273/files#diff-55927da6a21d8a4a5666708880315de6R21

jwalgran: Sorry I missed that.

src/nyc_trees/apps/home/urls.py
steventlamb: I could reproduce, and I get it now. I spoke to soon. this is kind of a flat pages issue, because flat pages is not requiring trailing slashes in their pattern, you never bottom out to the `APPEND_SLASH` redirect.
https://docs.djangoproject.com/en/1.7/ref/contrib/flatpages/#using-the-urlconf

steventlamb:
@jwalgran @jfrankl @kdeloach @maurizi All comments addressed.

src/nyc_trees/apps/home/routes.py
```diff
@@ -6,6 +6,7 @@
 from django_tinsel.decorators import route, render_template
 from django_tinsel.utils import decorate as do
 
+from apps.core.decorators import update
```
kdeloach: Does `update` still exist?

kdeloach: > ImportError: cannot import name update

kdeloach:
+1

src/nyc_trees/apps/home/routes.py
steventlamb: sigh. fixed!

maurizi:
LGTM

PR #274, opened 2015-01-09
Add event starting soon logic

kdeloach:


kdeloach:
@azavea-bot rebuild

src/nyc_trees/apps/event/models.py
```diff
@@ -58,9 +61,16 @@ def has_space_available(self):
...
+    def starting_soon_from_now(self):
+        return self.starting_soon(timezone.now())
+
+    def starting_soon(self, dt):
+        """
+        Return True if datetime (dt) is within the event starting period.
+        """
+        starting_soon_period = timedelta(hours=1)
+        return dt >= self.begins_at - starting_soon_period \
+               and dt < self.ends_at
```
maurizi: Consider rewriting this as:
```python
return self.begins_at - starting_soon_period <= dt < self.ends_at
```

maurizi:
+1 after you fix your lint

kdeloach:
@azavea-bot rebuild

kdeloach:
Thanks, fixed lint.

src/nyc_trees/apps/event/models.py
```diff
@@ -58,9 +61,16 @@ def has_space_available(self):
         return self.eventregistration_set.count() < self.max_attendees
 
     @property
-    def starting_soon(self):
-        # TODO: Implement
-        return True
+    def starting_soon_from_now(self):
+        return self.starting_soon(timezone.now())
+
+    def starting_soon(self, dt):
```
steventlamb: I think this api would be a little cleaner as:
```python
# it's nice to put config-related, arbitrary values up top.
_STARTING_SOON_WINDOW = timedelta(hours=1)

# ...

# This way you don't need a separate method that exists just for testing.
def starting_soon(self, seed_dt=None):
    seed_dt = seed_dt or timezone.now()
    return (seed_dt >= self.begins_at - _STARTING_SOON_WINDOW and
               seed_dt < self.ends_at)
```

This shouldn't require changes to any of the other files, other than names.

src/nyc_trees/apps/event/tests.py
```diff
@@ -135,3 +138,20 @@ def test_rsvp_limit_increase(self):
         self.event = Event.objects.get(id=self.event.id)
         self.assertEqual(15, context['max_attendees'])
         self.assertEqual(15, self.event.max_attendees)
+
+    def test_starting_soon(self):
```
steventlamb: This is an excellent set of test cases. The only drawback is that they don't verify that the `starting_soon == True` for the entire length of the event, only the exact moment of start. Consider adding one more assertion for `hour=12, minute=n` where `0 < n < 60`

src/nyc_trees/apps/event/models.py
kdeloach: I have more of a preference for the existing syntax.

steventlamb:
+1

src/nyc_trees/apps/event/models.py
maurizi: Ok

src/nyc_trees/apps/event/tests.py
kdeloach: True, I'll add this to the test case.

src/nyc_trees/apps/event/models.py
kdeloach: Cool, I'll update this. Is the underscore prefix necessary on constants though?

steventlamb: Not required, it's somewhat conventional, a workaround to the fact that python modules don't support private declarations. It just states your intention not to use it elsewhere. A minor point.

src/nyc_trees/apps/event/tests.py
kdeloach: The test looks much cleaner after this update.

PR #277, opened 2015-01-09
Revise check-in page layout

kdeloach:
The layout has changed from the wireframe to provide a visual cue that
the RSVP limit has increased after pressing the "increase RSVP limit"
button.


jwalgran:
Taking a look at this now

jwalgran:
+1 on the new indicator.

The content is pressed up against the sides of the main section

screen shot 2015-01-09 at 12 59 56 pm

I took a look at the group page and there is a ``<main class="col-sm-8">`` tag that looks like it is providing the border

screen shot 2015-01-09 at 1 01 58 pm

kdeloach:
The formatting issues should be fixed now.

PR #281, opened 2015-01-09
Simplify fetchAndReplace buttons

kdeloach:
This adds better support for having many fetchAndReplace buttons on the
same page and also fixes an issue with the followGroup button where
clicking it would result in nested container markup.


kdeloach:
Fixed failing JS tests.

jwalgran:
+1
Supporting multiple targets is a good addition
I manually tested all three button types (follow, rsvp, and check in)

PR #282, opened 2015-01-09
Switch event pages to two frame template

jfrankl:
Converting a few more event pages to the new template


kdeloach:
This needs to be rebased.

kdeloach:
+1

PR #284, opened 2015-01-09
Fix settings.js view

jwalgran:
The JS settings view was recently moved out of the urls file into a proper view function. The new view pipeline was no longer returning the correct content-type header and was not running the context processors. This commit restores those items.


maurizi:
+1

PR #285, opened 2015-01-10
Normalize settings.js naming

jwalgran:
NOTE: This PR implements my preference and does not have to be merged.

The template is named settings.js, and I often look for the wrong name when searching for the view or url that uses it. This commit removes the ambiguity, changing the url to /settings.js and all reference in Python to settings_js.


src/nyc_trees/apps/core/templates/base.html
```diff
@@ -1,5 +1,5 @@
-{% load utils %}
```
steventlamb: why this change?

steventlamb:
+1

RickMohr:
Fine with me to change

kdeloach:
+1

maurizi:
+1

jwalgran:
Closed in favor of #295

PR #287, opened 2015-01-10
Remove bounds context processor

jwalgran:
Context processors can be a little mysterious. It is easier understand the response pipeline when the view function is responsible for generating all the context used by the template.

I originally added this context processor when the JS settings template was not being passed any custom context. Now that the JS settings template is backed by a full view, the context_processor can be eliminated.


jwalgran:
I will work up an alternative to this pull + #285 that restores the context processor and renders ``window.config = ...`` in ``base.html``

jwalgran:
Closed in favor of #295

PR #289, opened 2015-01-10
Add maximum image size validation to Group form

jwalgran:
We want to prevent large profile images from slowing down requests to the group profile page.


src/nyc_trees/apps/users/forms.py
```diff
@@ -47,6 +50,21 @@ class Meta:
...
         ]
 
+    def clean_image(self):
+        image = self.cleaned_data.get('image', False)
+        try:
+            if image._size > settings.MAX_GROUP_IMAGE_SIZE_IN_BYTES:
+                size_string = humanize_bytes(image._size)
+                max_size_string = humanize_bytes(
+                    settings.MAX_GROUP_IMAGE_SIZE_IN_BYTES)
+                raise ValidationError('Files must be less than %s in size. '
```
maurizi: Consider changing `Files` to `Images` or `Logos`

maurizi:
For a good UX experience we should probably bump up the `nginx` max file upload size to something like 20MB (possibly only for this URL).

That should be large enough for any image, while still preventing malicious users from abusing image uploads.

This way, we expand our opportunity to give good error messages back to the user that stay in the Django application, instead of forcing them to a generic "this upload was too large" nginx error page.

jwalgran:
We explicitly want small images, so that clients are not downloading multi-meg files that are only showing in a 100x100 box anyway.

src/nyc_trees/apps/users/forms.py
jwalgran: That is a good idea

jwalgran:
I pushed an update that includes help text on the form

screen shot 2015-01-12 at 8 54 50 am

maurizi:
+1.  I was able to upload some filesa and see appropriate error validation.

jwalgran:
Sorry for the fixup commit

PR #294, opened 2015-01-12
Add support for capturing Windshaft logs and metrics

hectcastro:
This changeset adds support for Windshaft log shipping via Beaver to the shared Logstash instance and populates the StatsD configuration in Windshaft.


kdeloach:
+1

PR #295, opened 2015-01-12
Embed JS settings in base template

jwalgran:
Embedding the global config settings into the base template allows us to skip making an additional HTTP request on every page load.

This is an alternative to #285 and #287


src/nyc_trees/nyc_trees/context_processors.py
```diff
@@ -3,11 +3,44 @@
...
+            'xmin': float(s.NYC_BOUNDS[0]),
+            'ymin': float(s.NYC_BOUNDS[1]),
+            'xmax': float(s.NYC_BOUNDS[2]),
+            'ymax': float(s.NYC_BOUNDS[3])
+        }
+    }
+
+
+def tiler_cache_busters(request):
+    blockface_updated_at = _get_last_updated_datetime(Blockface)
```
maurizi: Consider checking if this is an AJAX request (https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.HttpRequest.is_ajax) before making these DB calls.

We don't need the tile cache busters for HTML partials, and it's probably a safe bet that an AJAX request will mean we're returning an HTML partial, not a full page that extends `base.html`.

Something like this:
```python
def tiler_cache_busters(request):
    if request.is_ajax():
        return {}
    ...
```

maurizi:
+1
I think this is a better approach than what we're currently doing from a performance standpoint.

RickMohr:
+1 otherwise

kdeloach:
+1

jwalgran:
@azavea-bot rebuild

jwalgran:
@azavea-bot rebuild

jwalgran:
@azavea-bot rebuild

maurizi:
@azavea-bot rebuild

PR #296, opened 2015-01-12
Use updated Kibana role with support for version 4 beta 3

hectcastro:
This changeset bumps the version of Kibana used to query the Logstash index to 4 beta 3.

See also: https://github.com/azavea/ansible-kibana/pull/3


RickMohr:
Sorry I didn't notice this assignment for a few days.
I successfully reprovisioned and saw that the Kibana version went from 4.0.0-beta2 to 4.0.0-beta3 and still shows data.
+1

PR #298, opened 2015-01-13
Crash gulp build if browserify fails

jwalgran:
We want the gulp process to have an non-zero exit code if there is a problem with bundling.

There may be a better, or more generic way to do this, but this catches our most recent problem.

### DIFF
--- a/src/nyc_trees/gulpfile.js
+++ b/src/nyc_trees/gulpfile.js
@@ -153,6 +153,7 @@ function browserifyTask(bundler) {
         })
         .bundle()
         .on('error', gutil.log.bind(gutil, 'Browserify Error'))
+        .on('error', process.exit.bind(process, 1))
         .pipe(source('common.js'));
 
     var bundles = entryBundles.concat(commonBundle);


src/nyc_trees/gulpfile.js
```diff
@@ -153,6 +153,7 @@ function browserifyTask(bundler) {
         })
         .bundle()
         .on('error', gutil.log.bind(gutil, 'Browserify Error'))
+        .on('error', process.exit.bind(process, 1))
```
maurizi: Can we put a condition around this so that `npm run watch` will keep running if Browserify fails to bundle the JS?

It might be a little tricky with the way the `watchify` / `browserify` tasks are setup, but it's nice to be able to leave the `watch` script running and not worry about accidentally saving a syntax error.

jwalgran: That would be nice. Tried to do it and failed in my first attempt.

maurizi: Well it's a nice to have really.

We could be a little messy and just pass in a boolean `isWatchify` parameter to `browserifyTask`, and use that plus `gulpif` to gate whether or not error should exit the process.

kdeloach: How about `function() { if (!args.debug) process.exit(1); }`?

maurizi: @kdeloach The `debug` flags does not necessarily imply that `watchify` is running.  Though gating on `debug` would at least prevent Jenkins builds from breaking.

jwalgran: @kdeloach I tried to do something like that, looking for "watch" in ``process.argv``. Crashing is easy, but I was having problems getting watch to resume properly.

jwalgran:
I am inclined to close this. It is too specific. If _any_ of the tasks fails, gulp should exit 1, not just the browserify task.

maurizi:
@jwalgran I say merge it.  It is at least a *little* better than what we have now, and should catch the majority of issues.

maurizi:
+1

src/nyc_trees/gulpfile.js
kdeloach: Got it.

kdeloach:
If this will display a better error message on Jenkins I say merge. +1

PR #300, opened 2015-01-13
Add styling and UTF8grid data to progress map tile layer

maurizi:
- Expands the progress layer to show survey/reservation information
 - Refactors the context processor / tiler URL generation
 - Adds a watch command for the tiler

### DIFF
--- a/src/nyc_trees/apps/core/templates/core/partials/config.js
+++ b/src/nyc_trees/apps/core/templates/core/partials/config.js
@@ -8,8 +8,8 @@
             "geocode": "{% url 'geocode' %}",
             "layers": {
                 "progress": {
-                    "tiles": "{{ tiler_url }}/{{ cache_buster.progress }}/{{ db_name }}/progress/{z}/{x}/{y}.png",
-                    "grids": "{{ tiler_url }}/{{ cache_buster.progress }}/{{ db_name }}/progress/{z}/{x}/{y}.grid.json"
+                    "tiles": "{{ progress_tiles_url }}",
+                    "grids": "{{ progress_grids_url }}"
                 }
             }
         },
--- a/src/nyc_trees/apps/survey/models.py
+++ b/src/nyc_trees/apps/survey/models.py
@@ -14,10 +14,16 @@ class Blockface(NycModel, models.Model):
     geom = models.MultiLineStringField()
     is_available = models.BooleanField(default=True)
 
+    def __unicode__(self):
+        return '%s (available: %s)' % (self.pk, self.is_available)
+
 
 class Territory(NycModel, models.Model):
     group = models.ForeignKey(Group)
     blockface = models.ForeignKey(Blockface, unique=True)
+
+    def __unicode__(self):
+        return '%s -> %s' % (self.group, self.blockface_id)
 
 
 class Survey(NycModel, models.Model):
@@ -28,7 +34,8 @@ class Survey(NycModel, models.Model):
     # a User object to automatically cascade and delete the Survey
     user = models.ForeignKey(User, on_delete=models.PROTECT)
     teammate = models.ForeignKey(User, null=True, on_delete=models.PROTECT,
-                                 related_name='surveys_as_teammate')
+                                 related_name='surveys_as_teammate',
+                                 blank=True)
     is_flagged = models.BooleanField(default=False)
 
 
@@ -51,4 +58,7 @@ class BlockfaceReservation(NycModel, models.Model):
     blockface = models.ForeignKey(Blockface, on_delete=models.PROTECT)
     is_mapping_with_paper = models.BooleanField(default=False)
     expires_at = models.DateTimeField()
-    canceled_at = models.DateTimeField(null=True)
+    canceled_at = models.DateTimeField(null=True, blank=True)
+
+    def __unicode__(self):
+        return '%s -> %s' % (self.user, self.blockface_id)
--- a/src/nyc_trees/nyc_trees/context_processors.py
+++ b/src/nyc_trees/nyc_trees/context_processors.py
@@ -3,30 +3,19 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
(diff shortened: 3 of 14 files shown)


maurizi:
@jwalgran This has been updated.

I accidentally rebased on `develop` and deleted all of your comments :cry:.

I didn't change anything in the older commits though, the only changes are in the new commits, 600b5ce, 3d6ba86, and d11671c

src/nyc_trees/nyc_trees/context_processors.py
```diff
@@ -10,23 +10,12 @@
 
 from django.utils.timezone import make_aware, utc
 
-from apps.survey.models import Blockface, Survey
+from apps.survey.models import (Blockface, Survey, Territory,
+                                BlockfaceReservation)
+from apps.users.models import TrustedMapper
 
 
 def settings(request):
```
jwalgran: The original version of the ``settings`` context processor was named ``settings`` because all it did was poke setting through to the context. If we are going to expand it, I would rather give it a more accurate name, and undo the ``import settings as s`` aliasing.

Alternatively, ``settings`` could remain a stateless context processor and we move the request-based tiler logic into a separate function.

maurizi: I'm down for renaming it.  How does `config_js` sound, to match the HTML partial?

src/nyc_trees/apps/core/templates/core/partials/config.js
```diff
@@ -8,8 +8,8 @@
             "geocode": "{% url 'geocode' %}",
             "layers": {
                 "progress": {
-                    "tiles": "{{ tiler_url }}/{{ cache_buster.progress }}/{{ db_name }}/progress/{z}/{x}/{y}.png",
-                    "grids": "{{ tiler_url }}/{{ cache_buster.progress }}/{{ db_name }}/progress/{z}/{x}/{y}.grid.json"
+                    "tiles": "{{ progress_tiles }}",
+                    "grids": "{{ progress_grids }}"
```
RickMohr: How about `progress_tiles_url_template` and `progress_grids_url_template`?

src/nyc_trees/apps/survey/models.py
```diff
@@ -28,7 +34,8 @@ class Survey(NycModel, models.Model):
     # a User object to automatically cascade and delete the Survey
     user = models.ForeignKey(User, on_delete=models.PROTECT)
     teammate = models.ForeignKey(User, null=True, on_delete=models.PROTECT,
-                                 related_name='surveys_as_teammate')
+                                 related_name='surveys_as_teammate',
+                                 blank=True)
```
RickMohr: Wouldn't this also have to be `null=True`?

maurizi: ... It is?

src/nyc_trees/nyc_trees/context_processors.py
jwalgran: It is a little odd to name it after a specific template, since context processor output is available to all templates. ``config`` may be clear enough.

src/tiler/sql/progress.sql
```diff
@@ -0,0 +1,19 @@
...
+  LEFT OUTER JOIN survey_territory AS turf
+    ON block.id = turf.blockface_id
+  LEFT OUTER JOIN survey_survey AS survey
+    ON block.id = survey.blockface_id
+  LEFT OUTER JOIN survey_blockfacereservation AS reservation
+    ON (block.id = reservation.blockface_id
+        AND reservation.canceled_at IS NULL
+        AND reservation.expires_at > now() at time zone 'utc')
+  ORDER BY block.id, survey.created_at DESC NULLS LAST
+) AS query
```
RickMohr: Nice. Clear. Consider a comment explaining that the `DISTINCT ON` and `ORDER BY` handle the case of multiple surveys of a blockface.

RickMohr:
Otherwise +1. Hooray for colored blocks!

src/tiler/sql/user_progress.sql
```diff
@@ -0,0 +1,26 @@
+(SELECT
+  DISTINCT ON (block.id)
+  block.geom, block.id, turf.group_id,
+  CASE
+    WHEN (survey.user_id IS NOT DISTINCT FROM <%= user_id %>
```
kdeloach: Is `IS NOT DISTINCT FROM` equivalent to `=`?

maurizi: `IS NOT DISTINCT FROM` is essentially "null-safe" equality.  See the bottom of [this page](http://www.postgresql.org/docs/9.1/static/functions-comparison.html) for more info

src/tiler/sql/progress.sql
```diff
@@ -0,0 +1,19 @@
+(SELECT
+  DISTINCT ON (block.id)
```
kdeloach: Nice feature `DISTINCT ON`.

src/tiler/sql/user_progress.sql
kdeloach: Cool. Does that mean surveys do not have to be associated with a user?

maurizi: Surveys have to be associated with a user, but not every block will have a Survey.

The reason this column could be `NULL` is because our `LEFT OUTER JOIN` might not find any rows in the Survey table matching our Blockface.

src/nyc_trees/nyc_trees/context_processors.py
maurizi: Done

src/nyc_trees/apps/core/templates/core/partials/config.js
maurizi: I went with `progress_tiles_url` and `progress_grids_url`

maurizi:
@kdeloach @jwalgran @RickMohr All comments addressed

kdeloach:
Nice job +1

PR #301, opened 2015-01-13
Fix broken unit test

maurizi:
It is unclear to me why this only started failing recently.


maurizi:
@azavea-bot rebuild

jwalgran:
+1

PR #304, opened 2015-01-14
Legend is user-aware and colors match tiler colors

RickMohr:
* Use different legend entries depending on whether user is logged in
* Use the same names and colors the tiler uses


maurizi:
+1

PR #306, opened 2015-01-15
Add tracking of training steps

steventlamb:
This needs unit tests, which I will add before merging. I wanted to open this PR ASAP to confirm that the technique is the right one.


src/nyc_trees/apps/home/views.py
```diff
@@ -32,6 +35,28 @@ def retrieve_job_status(request):
     pass
 
 
+def redirect_to_flat_page(url):
+    def fn(request, *args, **kwargs):
+        return flatpage(request, url)
+    return fn
+
+
+def mark_user_on_success(attr):
```
jwalgran: I like this as a wrapper. It looks like we are going to trigger "marks" not on the initial navigation to the page, but when the person clicks "Next" to move on to another section. This design has us covered. We can just adjust the "Next" button url and the redirect connected to the route.

steventlamb:
A problem with this approach that I couldn't figure out how to solve is that if the `/basic_training/` and `/intro_video/` flat pages do not exist in the database, these endpoints will just 404. Seems like something we need to solve with a placeholder fixture, maybe.

jwalgran:
Parks/TreeKit is working on the structure of the training materials. Right now, they are proposing 4 major sections, which would be shown as checklist items on the dashboard:

1. Getting Started
2. The Mapping Method
3. Tree Data
4. Tree Surrounding

(reference: http://sgowbu.axshare.com/#p=treekit_2015_census_sitemap)

We should use those 4 as our boolean columns: ``visited_training_section_*``

jwalgran:
@steventlamb Is this ready for another review?

jwalgran:
@azavea-bot rebuild

steventlamb:
Closing this because I've been reworking it and pushing work in progress commits in the interim. I will reopen when it is ready. Sorry for the noise!

PR #311, opened 2015-01-15
Remove the port override section from the README

maurizi:
These are no longer used

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,17 +46,17 @@ running in the VMs. You can change the host port numbers by setting
 the environment variables listed in the ``Env Variable Override``
 column.
 
-Service                | Port  | URL                                              | Env Variable Override
----------------------- | ----- | ------------------------------------------------ | ---------------------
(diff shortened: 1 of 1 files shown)


hectcastro:
@azavea-bot rebuild

hectcastro:
(Previous build failure was due to conflicts with the Vagrant user testing instance for NYC.)

Back to the PR. These port overrides are no longer needed because instead of overriding the ports, we're overriding the IPs assigned to machines and letting Vagrant automatically correct port collisions. I think that it may be worth adding the two IP address environmental variables as part of this PR:

- `NYC_TREES_SERVICES_IP` – Overrides the IP address of the `services` virtual machine (default: `33.33.33.20`)
- `NYC_TREES_TILER_IP` – Overrides the IP address of the `tiler` virtual machine (default: `33.33.33.20`)

The application virtual machine gets a `33.33.33.*` IP address via DHCP.

maurizi:
I think we can add the IP address info in a later PR

PR #317, opened 2015-01-16
Add endpoint for calculating GeoJSON polygon of a group's territory

maurizi:
Does not attempt to cluster disjoint groups of blockfaces into separate
polygons.

After some research, it seems like the best course of action *may* be to
manually create polygons for those groups which have disjoint territories.


RickMohr:
+1

maurizi:
@azavea-bot rebuild

PR #318, opened 2015-01-19
Event page shows location on map at top of page

RickMohr:


RickMohr:
Closing this for now as it makes more sense combined with the event edit page.

PR #321, opened 2015-01-19
Show small map with event location on event detail and edit pages.

RickMohr:


hectcastro:
@azavea-bot rebuild

jwalgran:
The mini maps should not be scrollable or zoomable, to prevent accidentally moving them when you are trying to scroll the page.

jwalgran:
I was able to move an event point over the city via the Django admin, then see it on both the event and event edit pages.

+1 after the map interactions are disabled.

Issue #320, opened 2015-01-19
Can't provision app server with Firefox 34.0+build2-0ubuntu0.14.04.1


jwalgran:
```
TASK: [nyc-trees.app | Install Firefox for UI tests] ************************** 
failed: [app] => {"failed": true}
stderr: E: Version '34.0+build2-0ubuntu0.14.04.1' for 'firefox' was not found

stdout: Reading package lists...
Building dependency tree...
Reading state information...

msg: 'apt-get install 'firefox=34.0+build2-0ubuntu0.14.04.1'' failed: E: Version '34.0+build2-0ubuntu0.14.04.1' for 'firefox' was not found

**Comment by steventlamb** [comment]
I ran into this as well. I worked around it by commenting this block out and making sure firefox was installed via `sudo apt-get install firefox` when provisioning was done.

**Comment by maurizi** [comment]
Looks like a new version of Firefox was released, and the old version is no longer available from the repos.

Last time this happened (firefox releases about every 6 weeks) we just bumped the Firefox version we're using.

If we want a more permanent fix, we can install froma `.deb` file: http://security.ubuntu.com/ubuntu/pool/main/f/firefox/

**Comment by jwalgran** [comment]
@hectcastro do you have a preference for how we deal with vanishing firefox versions? Installing from a .deb is probably closest to what we are trying to do with our existing version pinning.

**Comment by hectcastro** [comment]
With newer versions of Ansible, we should be able to glob it:

```diff
-  apt: pkg=firefox=34.0+build2-0ubuntu0.14.04.1 state=present
+  apt: pkg=firefox="3*" state=present
```

jwalgran:
Even better

kdeloach:
Fixed by #322?

jwalgran:
I think so, have not run into any provisioning problems since

PR #322, opened 2015-01-19
Glob Firefox and Xvfb test dependencies

hectcastro:
This changeset uses APT globbing support to relax the version specification associated with Firefox and Xvfb.

Attempts to resolve: #320


maurizi:
I removed firefox from my `app` VM and reprovisioned with this branch.  It succesfully installed firefox.
+1

Issue #297, opened 2015-01-12
Wrap ``vagrant provision`` so that it exits with -1 if the VM is not running


jwalgran:
``vagrant provision`` is exiting with 0 if the VM is not running, which can lead to builds marked as "good" when the provisioner failed.

```
$ vagrant halt app
$ vagrant provision app
==> app: VM is not currently running. Please, first bring it up with `vagrant up` then run this command.
$ echo $?
0

**Comment by hectcastro** [comment]
Jenkins runs the execute shell step with `-xe` by default, which prevents the retry mechanism in `vagrant-up.sh` from doing its thing. When a failure occurs during provisioning, the cleanup execute shell step is running `vagrant halt` before the first retry can happen.

I'm trying to remedy this by reversing the default of `-e` with with `set +e`, which doesn't halt the script on first failure. Instead, the last line of the script determines the outcome. If all of the Vagrant provisioning retries fail, the conditional below will short circuit into a failure. If any of them succeed, then the status of the build will be determined by `test.sh`.

```bash
set +e

./scripts/vagrant-up.sh && ./scripts/test.sh 
```

jwalgran:
That makes more sense. I was not sure why the VM stopped after the failure. Do you think we should still "correct" the way "vagrant provision" exits?

jwalgran:
using ``vagrant reload`` takes care of this

PR #324, opened 2015-01-19
Fix Vagrant provision false positive

hectcastro:
This changeset tries to address the scenario where a `vagrant provision` on a halted virtual machine returns an exit status of 0. Now a step to bring up the specific virtual machine is wrapped in an exponential backoff function along with the provisioning step. If bringing the virtual machine up fails, provisioning will never be attempted.

Attempts to resolve: #297

### DIFF
--- a/scripts/vagrant-up.sh
+++ b/scripts/vagrant-up.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
+
 DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)
 
 source "${DIR}/vagrant-env.sh"
@@ -8,5 +10,5 @@ vagrant up --no-provision
 
 for vm in services tiler app;
 do
-  with_backoff vagrant provision ${vm}
+  with_backoff vagrant reload --provision ${vm}
 done


hectcastro:
This build failure should be resolved by https://github.com/azavea/nyc-trees/pull/322.

hectcastro:
@azavea-bot rebuild

jwalgran:
Sorry for the delay. Looking at this now

jwalgran:
I was able to get a 0 exit code by waiting for the ``vagrant up`` phase to pass for each machine, then shutting off the machine via the Virtual Box GUI.

https://gist.github.com/jwalgran/7a77c2eb709a5ac019ac

jwalgran:
@hectcastro is this a work in progress or should I give it another shot?

hectcastro:
@azavea-bot rebuild

hectcastro:
Perhaps. I'm a bit out of ideas on how to tweak this after the `reload` change.

jwalgran:
Ok. I'll try it out again today.

jwalgran:
+1
I repeatedly crashed the VMs and it kept retrying to provision. When it ran out of attempts, it exited "1"

PR #328, opened 2015-01-20
Add reservable layer

jwalgran:
To guide a person in selecting blocks to add to their reservation, we need a simple available/not available layer.

I am adding this to the initial view of the reservation page for bootstrapping purposes only. When the "add" mode is implemented, it will show this layer only in that mode.

To implement this, I refactored the tile server and config context processor to be easily extensible.

### DIFF
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -21,7 +21,8 @@ var Windshaft = require('windshaft'),
     styles = files('./style'),
 
     interactivity = {
-        progress: 'id,group_id,survey_type'
+        progress: 'id,group_id,survey_type',
+        reservable: 'id,group_id,survey_type,restriction'
     },
 
     config = {
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -3,7 +3,7 @@
 var Windshaft = require('windshaft'),
     fs = require('fs'),
     _ = require('lodash'),
-
+    files = require('./files'),
     port = process.env.PORT || 4000,
 
     dbUser = process.env.NYC_TREES_DB_USER || 'nyc_trees',
@@ -17,10 +17,12 @@ var Windshaft = require('windshaft'),
     statsdHost = process.env.NYC_TREES_STATSD_HOST || 'localhost',
     statsdPort = process.env.NYC_TREES_STATSD_PORT || 8125,
 
-    progressSql = fs.readFileSync('sql/progress.sql', {encoding: 'utf8'}),
-    userProgressSql = _.template(fs.readFileSync('sql/user_progress.sql', {encoding: 'utf8'})),
+    queries = files('./sql'),
+    styles = files('./style'),
 
-    progressStyle = fs.readFileSync('style/progress.mss', {encoding: 'utf8'}),
+    interactivity = {
+        progress: 'id,group_id,survey_type'
+    },
 
     config = {
         base_url: '/:cache_buster/:dbname/:type',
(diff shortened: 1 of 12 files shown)


src/tiler/server.js
```diff
@@ -48,37 +46,77 @@ var Windshaft = require('windshaft'),
...
+
+        req2params: function(req, callback) {
+            try {
+                // Windshaft needs a table name, even if you are providing
+                // a custom sql statement.
+                req.params.table = 'survey_blockface';
+                req.params.sql = this.req2sql(req);
+                req.params.style = this.req2style(req);
+                req.params.interactivity = this.req2interactivity(req);
+                callback(null, req);
```
maurizi: I like following the Windshaft naming conventions, but I don't think we should add our own functions as "options" to the Windshaft config file.

Consider moving `req2context`, `req2interactivity`, `req2Style`, and `req2sql` into standalone functions below `config`

jwalgran: They were initially all standalone functions, but some of them are not pure and depend on elements of the config object. I tried leaving the pure functions standalone, but then similar functions were in separate places, which made it harder to follow what was going on.

jwalgran: I can try moving them back out an see how it looks.

jwalgran:
@maurizi lots of new stuff here including the restructuring we discussed, an additional interactivity item on the reservable layer, and some styling and layer visibility changes. Ready for a second pass.

jwalgran:
New fat line styling at zoom 19

screen shot 2015-01-20 at 11 27 43 am

jwalgran:
@azavea-bot rebuild

kdeloach:
I get an error when trying to view the grid JSON for progress map tiles.

URL: http://33.33.33.20/1420649640/nyc_trees/progress/16/19298/24642.grid.json?user=1

Response:

> {"error":"ERROR:  column \"restriction\" does not exist\nLINE 1: ...ELECT ST_AsBinary(\"geom\") AS geom,\"group_id\",\"id\",\"restricti...\n                                                             ^\n\nFull sql was: 'SELECT ST_AsBinary(\"geom\") AS geom,\"group_id\",\"id\",\"restriction\",\"survey_type\" FROM -- The 'DISTINCT ON' and 'ORDER BY' clause are needed so that we only use\n-- the latest survey for a blockface\n(SELECT\n  DISTINCT ON (block.id)\n  block.geom, block.id, turf.group_id,\n  CASE\n    WHEN (survey.user_id IS NOT DISTINCT FROM 1\n          OR survey.teammate_id IS NOT DISTINCT FROM 1) THEN 'surveyed-by-me'\n    WHEN survey.id IS NOT NULL THEN 'surveyed-by-others'\n    WHEN reservation.user_id IS NOT DISTINCT FROM 1 THEN 'reserved'\n    WHEN (block.is_available AND reservation.id IS NULL\n          AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'\n    ELSE 'unavailable'\n  END AS survey_type\n  FROM survey_blockface AS block\n  LEFT OUTER JOIN survey_territory AS turf\n    ON block.id = turf.blockface_id\n  LEFT OUTER JOIN survey_survey AS survey\n    ON block.id = survey.blockface_id\n  LEFT OUTER JOIN survey_blockfacereservation AS reservation\n    ON (block.id = reservation.blockface_id\n        AND reservation.canceled_at IS NULL\n        AND reservation.expires_at > now() at time zone 'utc')\n  LEFT OUTER JOIN users_trustedmapper AS trustedmapper\n    ON (turf.group_id = trustedmapper.group_id\n        AND trustedmapper.user_id = 1)\n  ORDER BY block.id, survey.created_at DESC NULLS LAST\n) AS query WHERE \"geom\" && ST_SetSRID('BOX3D(-74.00527954101562 40.69625781943066,-73.98056030273438 40.71499673928158)'::box3d, 4326)'\n"}

jwalgran:
Yes. I made an incorrect assumption about how interactivity works in Windshaft. I will fix it.

jwalgran:
@kdeloach the progress grid should now be fixed

maurizi:
+1, after you rebase your fixup commits.

I think the refactor to the tiler should make it easier to add more layers.

kdeloach:
+1 Looking good.

jwalgran:
Just pushed up the clean version. Will merge it in once the PR builder is green.

PR #330, opened 2015-01-20
Allow changing event location on a large map

RickMohr:
* `EventMap.js` creates a map with a highlighted and changeable location
* On AddEvent/EditEvent pages add a second pane with a large map
* Show static crosshairs on the large map
* Modify location via "Edit" "Done", and "Cancel" operations
* Require saved location to be within New York City bounds

### DIFF
--- a/src/nyc_trees/js/src/event.js
+++ b/src/nyc_trees/js/src/event.js
@@ -2,11 +2,11 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    mapModule = require('./map'),
+    EventMap = require('./EventMap'),
     fetchAndReplace = require('./fetchAndReplace');
 
 var $map = $('#map');
-mapModule.create({
+new EventMap({
     location: L.latLng($map.data('lat'), $map.data('lon')),
     static: true
 });
--- a/src/nyc_trees/apps/core/test_utils.py
+++ b/src/nyc_trees/apps/core/test_utils.py
@@ -46,7 +46,7 @@ def make_event(group, **kwargs):
         'title': "The event",
         'contact_email': "a@b.com",
         'address': "123 Main St",
-        'location': Point(0, 0),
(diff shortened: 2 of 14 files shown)


jwalgran:
Looking at this now

jwalgran:
Event unit tests are broken because the dummy points they use are not within NYC

src/nyc_trees/js/src/event.js
```diff
@@ -2,11 +2,11 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    mapModule = require('./map'),
+    EventMap = require('./EventMap'),
     fetchAndReplace = require('./fetchAndReplace');
 
 var $map = $('#map');
-mapModule.create({
+new EventMap({
```
jwalgran: Using ``new`` without assigning the value to anything looks a little odd. Is there a reason to move away from the ``mapModule.create(...)`` style?

jwalgran:
Some strange crosshair behavior if you resize the window. Lines appear and disappear.

Screencap
http://cl.ly/2D1k0y1K0t32/crosshair-strangeness-on-resize.mov

src/nyc_trees/js/src/event.js
RickMohr: I created a class so there would be a place to store the Leaflet `circleMarker` so it could be repositioned when you change the location.

jwalgran:
+1 when the tests are green.

Works well. I think it would be OK to push the crosshair bug out to another issue.

RickMohr:
The crosshairs are steady during resize in Chrome on Windows. What browser were you using?

jwalgran:
The screencap was from Chome 39.0.2171.95 on OS X 10.10.1

RickMohr:
@azavea-bot rebuild

RickMohr:
@azavea-bot rebuild

PR #339, opened 2015-01-21
Add "My Reservations" page

kdeloach:
This splits the existing reservations page into 2 pages. Clicking
"Reservations" in the sidebar will display your currently reserved blockfaces.
Clicking the "New" button will take you to the next page to reserve
blockfaces.


maurizi:
Looks good to me

jwalgran:
+1 Layer works perfectly.

If the user has blockfaces reserved, it makes sense to set the map view to the extent of those blockfaces.     It should be relatively easy to get a queryset of the blockfaces and drop the extent into a view context. https://docs.djangoproject.com/en/1.7/ref/contrib/gis/geoquerysets/#extent

We can split this feature off into a separate issue if you like, or roll it into this PR.

RickMohr:
Note that @maurizi just did this for the event list page, purely client side -- https://github.com/maurizi/nyc-trees/commit/08126836a6bfca3fbc03b1b359d3bc15be5b48e0

src/tiler/sql/user_reservations.sql
```diff
@@ -0,0 +1,9 @@
+(SELECT
```
steventlamb: for my education, how does this file get used? I can't find a reference to `user_reservations` in this diff.

steventlamb:
+1

maurizi:
@RickMohr Leaflet can't compute boundaries of tile layers client side.
We'll have to go with the approach Justin outlined of computing a bounding
box with Django.

src/tiler/sql/user_reservations.sql
kdeloach: Map tile requests made by an authenticated user are prefixed with 'user_' (See https://github.com/kdeloach/nyc-trees/blob/topic/my-reservations/src/tiler/server.js#L99)

kdeloach:
Cool, it looks like I can reuse a good amount of Mike's code for fitting the bounding box to the reservations extent, so I would like to do this after both PRs are merged.

PR #341, opened 2015-01-22
Add SES/boto based email backend for prod

jwalgran:
Using boto on a properly provisioned EC2 keeps the application out of the business of managing email credentials.

All boto interaction is isolated in the ``BotoMailer`` class, which has no dependencies on Django and can be tested in isolation.

This backend is configured for production only because it requires EC2 metadata services that aren't available on vagrant VMs.

### DIFF
--- a/src/nyc_trees/apps/core/mail/backends/boto.py
+++ b/src/nyc_trees/apps/core/mail/backends/boto.py
@@ -5,8 +5,10 @@ from __future__ import division
 
 import json
 import logging
+import sys
 import traceback
 
+from django.conf import settings
 from django.core.mail.backends.base import BaseEmailBackend
 
 from django_statsd.clients import statsd
@@ -21,13 +23,18 @@ class EmailBackend(BaseEmailBackend):
     def __init__(self, fail_silently=False, aws_region='us-east-1'):
         super(EmailBackend, self).__init__(fail_silently=fail_silently)
         self.mailer = BotoMailer(aws_region)
+        self.check_quota = getattr(settings, 'EMAIL_BOTO_CHECK_QUOTA', False)
 
     def send_messages(self, email_messages):
         if not email_messages:
             return
         sent_message_count = 0
-        remaining_quota = self.mailer.get_remaining_message_quota()
-        self._log_quota(remaining_quota)
+
+        if self.check_quota:
+            remaining_quota = self.mailer.get_remaining_message_quota()
+            self._log_quota(remaining_quota)
+        else:
+            remaining_quota = sys.maxint
 
         if len(email_messages) <= remaining_quota:
             for email_message in email_messages:
--- a/src/nyc_trees/apps/core/mail/backends/boto.py
+++ b/src/nyc_trees/apps/core/mail/backends/boto.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+import json
+import logging
+import traceback
+
+from django.core.mail.backends.base import BaseEmailBackend
+
+from django_statsd.clients import statsd
+
+from nyc_trees.libs.boto_mailer import BotoMailer
+
+logger = logging.getLogger(__name__)
+
+
+class EmailBackend(BaseEmailBackend):
+
+    def __init__(self, fail_silently=False, aws_region='us-east-1'):
+        super(EmailBackend, self).__init__(fail_silently=fail_silently)
+        self.mailer = BotoMailer(aws_region)
+
+    def send_messages(self, email_messages):
+        if not email_messages:
+            return
+        sent_message_count = 0
+        remaining_quota = self.mailer.get_remaining_message_quota()
+        self._log_quota(remaining_quota)
+
+        if len(email_messages) <= remaining_quota:
+            for email_message in email_messages:
+                if self._send_message(email_message):
+                    sent_message_count += 1
+            return sent_message_count
+        else:
+            raise SESQuotaException("Attempted to send %d messages with only "
+                                    "%d remaining. Wait before retrying" %
+                                    (len(email_messages), remaining_quota),
+                                    email_messages)
+
+    def _send_message(self, email_message):
+        message = email_message.message()
+        try:
+            response = self.mailer.send_message(
+                message.as_bytes(linesep='\r\n'),
+                email_message.from_email,
+                email_message.to)
+            self._log_message(email_message, response)
+            return response
+        except Exception:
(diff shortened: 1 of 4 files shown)


src/nyc_trees/requirements/base.txt
```diff
@@ -10,3 +10,4 @@ django-floppyforms==1.3.0
 python-dateutil==2.2
 django-statsd-mozilla==0.3.14
 python-omgeo==1.7.1
+boto==2.34.0
```
maurizi: Shouldn't this go in `requirements/production.txt`?

src/nyc_trees/apps/core/mail/backends/boto.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __init__(self, fail_silently=False, aws_access_key_id=None,
+                 aws_secret_access_key=None, aws_region=None):
+        super(EmailBackend, self).__init__(fail_silently=fail_silently)
+        self.aws_access_key_id =\
+            aws_access_key_id or settings.EMAIL_AWS_ACCESS_KEY_ID
+        self.aws_secret_access_key =\
+            aws_secret_access_key or settings.EMAIL_AWS_SECRET_ACCESS_KEY
+        self.aws_region =\
+            aws_region or settings.EMAIL_AWS_REGION
```
maurizi: Why allow passing these values in via the `__init__` method, instead of only using them from `settings`?

We won't be be explicitly instantiating this backend, right?

src/nyc_trees/apps/core/mail/backends/boto.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def close(self):
+        # SES connections are not stateful, so there is no cleanup needed
+        self.connection = None
+
+    def send_messages(self, email_messages):
+        if not email_messages:
+            return
+        sent_message_count = 0
+        if not self.open():
```
maurizi: I don't see a `return` statement in `self.open()`.  Where does this truthy return value come from when we connect succefully?

maurizi:
The overall design here looks good, it's very simple and easy to understand.

Should we add checks to `connection.get_send_quota()` prior to sending emails?
It wouldn't be great if some but not all of the emails of a batch email are sent.

src/nyc_trees/apps/core/mail/backends/boto.py
```diff
@@ -0,0 +1,95 @@
...
+            aws_secret_access_key or settings.EMAIL_AWS_SECRET_ACCESS_KEY
+        self.aws_region =\
+            aws_region or settings.EMAIL_AWS_REGION
+
+    def open(self):
+        try:
+            if not self.connection:
+                self.connection = boto.ses.connect_to_region(
+                    self.aws_region,
+                    aws_access_key_id=self.aws_access_key_id,
```
hectcastro: If the node this code runs on has IAM role support, the connection can just be:

```python
self.connection = boto.ses.connect_to_region(self.aws_region)
```

Boto will automatically hit the metadata API behind the scenes and pull in the `aws_access_key_id` and `aws_secret_access_key`: https://github.com/boto/boto/blob/bbbf9d248856db997080f51805575f0634da4090/boto/provider.py#L392-L402

We can add support for providing the credentials explicitly, but the `connect_to_region` would need to branch two ways:

- If credentials are provided, use them
- If no credentials are provided, don't supply the arguments and let Boto try to get them from the metadata

src/nyc_trees/nyc_trees/settings/production.py
```diff
@@ -41,30 +41,10 @@ def get_env_setting(setting):
...
-
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#email-use-tls
-EMAIL_USE_TLS = True
-
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#server-email
-SERVER_EMAIL = EMAIL_HOST_USER
+EMAIL_BACKEND = 'apps.core.mail.backends.boto.EmailBackend'
+EMAIL_AWS_ACCESS_KEY_ID = environ.get('EMAIL_AWS_ACCESS_KEY_ID', None)
+EMAIL_AWS_SECRET_ACCESS_KEY = environ.get('EMAIL_AWS_SECRET_ACCESS_KEY', None)
+EMAIL_AWS_REGION = environ.get('EMAIL_AWS_REGION', None)
```
hectcastro: Does it make sense to make the environment variable `AWS_REGION` vs. having it e-mail specific? Regardless, it would probably be safe to default it to `us-east-1`.

src/nyc_trees/apps/core/mail/backends/boto.py
```diff
@@ -0,0 +1,95 @@
...
+            self._log_message(email_message, response)
+            return response
+        except Exception:
+            self._log_message_failure(email_message,
+                                      traceback.format_exc())
+            if not self.fail_silently:
+                raise
+
+    def _log_connect(self, connection):
+        statsd.incr('email.connection.success')
```
hectcastro: Nice.

jwalgran:
On a properly provisioned EC2, this code successfully sent an email

```python
from boto_mailer import BotoMailer

b = BotoMailer('us-east-1')
b.send_plain_text_message("Body text from boto", "Boto test", "info@azavea.com", "jwalgran@azavea.com")

**Comment by maurizi** [comment]
+1, this looks really clean.

File: src/nyc_trees/libs/boto_mailer.py [hunk 7]
```diff
@@ -0,0 +1,32 @@
...
+        message['To'] = to
+        self.send_message_bytes(message.as_string(), from_email, to)
+
+    def send_message_bytes(self, message_bytes, from_email, to):
+        connection = boto.ses.connect_to_region(self.aws_region)
+        return connection.send_raw_email(message_bytes, from_email, to)
+
+    def remaining_message_quota(self, message_count):
+        response = self.connection.get_send_quota()
+        quota = response['GetSendQuotaResponse']['GetSendQuotaResult']
```
hectcastro: Access to the quota API calls depends on https://github.com/azavea/nyc-trees/issues/344.

jwalgran:
Access to the quota API calls depends on #344.

6208133 includes an update to the quota error handling. The original messages are now returned with the quota exception for flexibility in handling how to resend the messages.

I also added a [statsd gauge](http://statsd.readthedocs.org/en/latest/types.html#gauges) to record the quota every time we check it, so we can visualize how close we really get to using it up.

jwalgran:
@hectcastro f5c7a39 adds a setting that removes the dependency on #344

hectcastro:
:+1:

PR #342, opened 2015-01-23
Refactor app and tiler dependencies and add filter plugins

hectcastro:
This pull request contains two commits because I couldn't get the refactoring to work without the addition of the customer filters.

One commit adjusts the shared dependencies between the application and tiler so that they exist in one place (`nyc-trees.common`). The other adds custom filters to help determining a node's group membership.

### DIFF
--- a/deployment/ansible/filter_plugins/custom_filters.py
+++ b/deployment/ansible/filter_plugins/custom_filters.py
@@ -1,0 +1,45 @@
+class FilterModule(object):
+    ''' Additional filters for use within Ansible. '''
+
+    def filters(self):
+        return {
+            'is_not_in': self.is_not_in,
+            'is_in': self.is_in,
+            'some_are_in': self.some_are_in
+        }
+
+    def is_not_in(self, *t):
+        """Determnies if there are no elements in common between x and y
+
+            x | is_not_in(y)
+
+        Arguments
+        :param t: A tuple with two elements (x and y)
+        """
+        groups_to_test, all_group_names = t
+
+        return set(groups_to_test).isdisjoint(set(all_group_names))
+
+    def is_in(self, *t):
+        """Determnies if all of the elements in x are a subset of y
+
+            x | is_in(y)
+
+        Arguments
+        :param t: A tuple with two elements (x and y)
+        """
+        groups_to_test, all_group_names = t
+
+        return set(groups_to_test).issubset(set(all_group_names))
+
+    def some_are_in(self, *t):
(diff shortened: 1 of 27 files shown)


maurizi:
At a high level this change looks good.  I haven't had the opportunity to test it out yet.

hectcastro:
There is a subtle difference between `issubset` and `isintersecting`. The former works like `AND` and the latter `OR`. Mainly important for the (if `development` or `test`) condition.

I think for the single element comparisons usage could go either way. I like the idea of using both so that examples exist of how they differ.

hectcastro:
Filters have been renamed. I tested both the development and testing setups locally today and did not detect any issues.

maurizi:
+1

jwalgran:
@azavea-bot rebuild

jwalgran:
Ok to merge. Test failure was unrelated, and may be due to upstream geocoder issues unrelated to this change.

deployment/ansible/filter_plugins/custom_filters.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def is_not_in(self, *t):
+        """Determnies if there are no elements in common between x and y
+
+            x | is_not_in(y)
+
+        Arguments
+        :param t: A tuple with two elements (x and y)
+        """
+        groups_to_test, all_group_names = t
```
kdeloach: Any special reason these are unpacked like this instead of defined as arguments?

deployment/ansible/filter_plugins/custom_filters.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def filters(self):
+        return {
+            'is_not_in': self.is_not_in,
+            'is_in': self.is_in,
+            'some_are_in': self.some_are_in
+        }
+
+    def is_not_in(self, *t):
+        """Determnies if there are no elements in common between x and y
```
kdeloach: Typo "Determnies"

kdeloach:
+1

deployment/ansible/filter_plugins/custom_filters.py
hectcastro: [Core plugins](https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/filter_plugins/core.py) had a similar method signature. Looking at them again, these might be better fixed to two elements.

hectcastro: Fixed in https://github.com/azavea/nyc-trees/pull/361.

PR #345, opened 2015-01-23
Topic/integrate training step tracking ui

steventlamb:


steventlamb:
This is a first pass at linking up training pages with tracking and sequencing and I think this system is robust enough to absorb spec changes and nonhomogenous training page types without too much difficulty.

Remaining items:
1 - Tighten order enforcement - the UI guides you toward stepping through in order, but doesn't stop you from hitting urls out of sequence and completing them.
2 - statsd logging
3 - integrating other pages (quiz) and verifying that it behaves as expected

steventlamb:
(closed on accident. reopened)

scripts/check.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+set -x
+
+# Run flake8 against the Django codebase and output a known string so that
+# the Jenkins text finder plugin can detect a failed check and mark the build
+# unstable. This command should only fail the build if the `vagrant ssh`
+# command itself fails.
+vagrant ssh app -c "flake8 /opt/app/apps --exclude migrations || echo flake8 check failed"
```
kdeloach: I will use this often.

src/nyc_trees/apps/home/training/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+                                'Groups To Follow', '5 minutes',
+                                route(GET=do(
+                                    render_template(
+                                        'home/groups_to_follow.html'),
+                                    v.groups_to_follow)))
+
+training_list_page = route(GET=do(render_template('home/training.html'),
+                                  v.training_list_page))
+
+training_summary = TrainingGateway(training_list_page,
```
kdeloach: Consider adding `routes.py`?

src/nyc_trees/apps/home/training/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+                                   getting_started, groups_to_follow)
+
+getting_started.previous_step = training_summary
```
kdeloach: Would it make sense to have `next_step` and `previous_step` constructor arguments on `TrainingStep` instead of this?

src/nyc_trees/apps/home/training/decorators.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import division
+
+from functools import wraps
+
+from django.db import transaction
+
+from django.contrib.flatpages.views import flatpage
+
+
+def render_flatpage(url):
```
kdeloach: Nice parity with `render_template`.

src/nyc_trees/apps/home/training/__init__.py
kdeloach: Consider letting `TrainingGateway` setup these links... ex `TrainingGateway(step1, step2, step3, ...)` where the constructor accepts variable number of steps in the constructor.

src/nyc_trees/apps/home/training/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+getting_started.next_step = the_mapping_method
+the_mapping_method.previous_step = getting_started
+the_mapping_method.next_step = tree_data
+tree_data.previous_step = the_mapping_method
+tree_data.next_step = tree_surroundings
+tree_surroundings.previous_step = tree_data
+tree_surroundings.next_step = groups_to_follow
+groups_to_follow.previous_step = tree_surroundings
+groups_to_follow.next_step = training_summary
```
kdeloach: Consider using `None` as the "null" step instead. It's easy to overlook that `TrainingGateway` is the sentinel to end this "linked list".

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+                next_step = step
+                is_next = True
+            else:
+                is_next = False
+            training_steps.append({'step': step,
+                                   'is_complete': is_complete,
+                                   'is_next': is_next})
+            last_was_complete = is_complete
+
+        return training_steps
```
kdeloach: Consider simplifying with a reduce, ex:

    def append_step(acc, step):
        if not result:
            return [{'step': step, 'is_next': True}]
        prev_step = acc[-1]
        acc.append({'step': step, 'is_complete': ..., 'is_next': ...})
        return acc
    return reduce(append_step, steps, None)

src/nyc_trees/apps/home/urls.py
```diff
@@ -17,11 +22,20 @@
...
+    url(r'^training/pure/getting_started/', **getting_started.pure_kwargs()),                          # NOQA
+    url(r'^training/pure/the_mapping_method/', **the_mapping_method.pure_kwargs()),                    # NOQA
+    url(r'^training/pure/tree_data/', **tree_data.pure_kwargs()),                                      # NOQA
+    url(r'^training/pure/tree_surroundings/', **tree_surroundings.pure_kwargs()),                      # NOQA
+    url(r'^training/pure/groups_to_follow/$', **groups_to_follow.pure_kwargs()),                       # NOQA
+    url(r'^training/progress/the_mapping_method/', **the_mapping_method.previous_step.mark_kwargs()),  # NOQA
+    url(r'^training/progress/tree_data/', **tree_data.previous_step.mark_kwargs()),                    # NOQA
+    url(r'^training/progress/tree_surroundings/', **tree_surroundings.previous_step.mark_kwargs()),    # NOQA
+    url(r'^training/progress/groups_to_follow/$', **groups_to_follow.previous_step.mark_kwargs()),     # NOQA
+    url(r'^training/progress/$', **training_summary.previous_step.mark_kwargs()),                      # NOQA
```
kdeloach: Could these training steps return a route instead of kwargs?

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+from django.core.urlresolvers import reverse
+from django.contrib.auth.decorators import login_required
+from django_tinsel.decorators import route
+from django_tinsel.utils import decorate as do
+from apps.home.training.decorators import render_flatpage, mark_user
+from django.contrib.flatpages.views import flatpage
+
+_PURE_TEMPLATE = '%s_pure'
+_MARK_PROGRESS_TEMPLATE = '%s_mark_progress'
+_TRAINING_FINISHED_TEMPLATE = 'training_finished_%s'
```
kdeloach: Consider renaming these or adding comments; It wasn't clear, at first, what kind of templates these are. It appears the first 2 are string formats for Django urls, and the last one is the format for properties on the `User` model.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+            steps.append(step)
+            step = step.next_step
+        return steps
+
+    def pure_kwargs(self):
+        return {'name': '%s_pure' % self.name,
+                'view': login_required(self.view)}
+
+
+class TrainingStep(object):
```
kdeloach: Consider using inheritance/mixins to distinguish flatpage training steps from normal page training steps.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+
+class TrainingGateway(object):
+    def __init__(self, view, first_step, last_step):
+        self.previous_step = last_step
+        self.next_step = first_step
+        self.view = view
+        self.name = None
+
+    def is_flatpage(self):
+        return False
```
kdeloach: Was this class supposed to be inherited by `TrainingStep`? It's unclear.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+    def mark_progress_url(self):
+        return reverse(_MARK_PROGRESS_TEMPLATE % self.name)
+
+    def pure_kwargs(self):
+        kwargs = {'url': '/%s/' % self.name} if self.is_flatpage() else {}
+        return {'name': _PURE_TEMPLATE % self.name,
+                'view': login_required(self.view),
+                'kwargs': kwargs}
+
+    def mark_kwargs(self):
```
kdeloach: It's a little mysterious why this class contains both `pure_kwargs`/ `mark_kwargs` and `pure_url`/`mark_progress_url`. Consider inheritance/mixins to reduce these code paths.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+from django_tinsel.utils import decorate as do
+from apps.home.training.decorators import render_flatpage, mark_user
+from django.contrib.flatpages.views import flatpage
+
+_PURE_TEMPLATE = '%s_pure'
+_MARK_PROGRESS_TEMPLATE = '%s_mark_progress'
+_TRAINING_FINISHED_TEMPLATE = 'training_finished_%s'
+
+
+class TrainingGateway(object):
```
kdeloach: This class has some useful methods, but I don't see any value in trying to make it "look like" a training step. It just confused me.

Consider making `get_step`, `is_complete`, `get_context`, and `get_steps` module-level functions and ditching `TrainingGateway`.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                is_next = False
+            training_steps.append({'step': step,
+                                   'is_complete': is_complete,
+                                   'is_next': is_next})
+            last_was_complete = is_complete
+
+        return training_steps
+
+    def get_steps(self):
```
kdeloach: Consider just returning a static array of steps unless there is a requirement that steps have to be dynamic/mutable.

src/nyc_trees/apps/home/tests.py
```diff
@@ -2,3 +2,96 @@
...
+from django.contrib.flatpages.models import FlatPage
+
+from django_tinsel.utils import decorate as do
+
+from apps.core.models import User
+from apps.core.test_utils import make_request
+from apps.home.training.decorators import mark_user, render_flatpage
+
+
+class TrainingTrackingTest(TestCase):
```
kdeloach: I wonder if it would be worthwhile to test that a user can actually make it all the way through the training process. Maybe in a future PR?

src/nyc_trees/apps/home/templatetags/training.py
```diff
@@ -0,0 +1,16 @@
...
+from django import template
+
+from apps.home.training import training_summary
+
+register = template.Library()
+
+
+@register.filter
+def flatpage_next_url(flatpage):
+    flatpage_name = flatpage.url[1:-1]
```
kdeloach: Consider adding a comment for this line.

src/nyc_trees/apps/home/urls.py
```diff
@@ -17,11 +22,20 @@
...
 
-    url(r'^training/$', r.training_list_page, name="training_list_page"),
-
-    url(r'^training/groups_to_follow/$', r.groups_to_follow),
-
-    # "training" instead of "training/" because the flatpages admin interface
-    # insists that the "URL" (really a URL segment) start with a leading slash
-    url(r'^training(?P<url>.*/)$', include('django.contrib.flatpages.urls')),
+    url(r'^training/$', **training_summary.pure_kwargs()),                                             # NOQA
+    url(r'^training/pure/getting_started/', **getting_started.pure_kwargs()),                          # NOQA
```
kdeloach: What's the idea behind `/pure/`?

kdeloach:
Verified that everything works as expected.

src/nyc_trees/apps/home/training/__init__.py
steventlamb: good idea, that would be more consistent with other apps.

src/nyc_trees/apps/home/training/types.py
steventlamb: No, not really. They both implement the same interface through ducktyping, but don't share code or data so inheritance is not useful.  It's hard to convey that in how they are defined, because interfaces and abstract classes are not a thing in python.

src/nyc_trees/apps/home/training/types.py
steventlamb: Sounds like a good idea, because then `is_flatpage` can then be a typecheck, ie, `isinstance(step, FlatpageTrainingStep)`

src/nyc_trees/apps/home/training/types.py
steventlamb: Yeah, better names coming.

src/nyc_trees/apps/home/urls.py
steventlamb: No, we need dynamically generated `name='<step_name>_pure'` and `name='<step_name>_mark_progress'` kwargs as well, because the steps will look them up.

src/nyc_trees/apps/home/training/__init__.py
steventlamb: Can't do that in this case, `next_step` must be a step-like object in order for the current step (groups_to_follow) to have a destination to bind its `mark_progress` endpoint to.

steventlamb: we should probably talk about how this could be clearer, or done better.

src/nyc_trees/apps/home/training/types.py
steventlamb: I'm trying to avoid code duplication and scope configuration to as few files as possible. I expect the list of steps in `__init__.py` to continue changing. A static list would require to write out the steps again. The fact that these lists are small means walking them to build the list will be cheap.

src/nyc_trees/apps/home/tests.py
steventlamb: Agreed.

jwalgran:
Looking at this now

src/nyc_trees/apps/core/models.py
```diff
@@ -88,6 +95,10 @@ def clean(self):
     def get_absolute_url(self):
         return reverse('user_detail', args=[self.username])
 
+    def completed_training(self):
```
jwalgran: There are two types of training, online and field, so to keep things unambiguous, we should probably name this ``completed_online_training``

jwalgran: I just realized that this kind of shadows the existing ``online_training_complete`` boolean flag. I think we may reduce possible confusion by moving the calculation of whether or not training is complete out to the view.

kdeloach:
+1

jwalgran:
Adding a comment explaining what "pure" means in the context of the feature is a good idea.

jwalgran:
+1
I checked this out, ran the management command, and I was able to run through all the pages and have them correctly recorded on the dash/training page.

src/nyc_trees/apps/home/training/types.py
steventlamb: (deferring this to a later refactor)

src/nyc_trees/apps/core/models.py
steventlamb: Are you suggesting we delete that boolean field as well? It's not currently used, except in this helper
```
def user_has_online_training(user):
    return user.is_authenticated() and user.online_training_complete
```

We could delete the boolean field and always use the model method. Or we can pull it off into a helper. I don't know if the view level is the best place to determine this, it seems like we're going to have multiple places that check if the user has completed training.

jwalgran: As long as a Parks admin can mark someone as trained without the person actually clicking through the process, then we are covered.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -0,0 +1,165 @@
...
+from django_tinsel.utils import decorate as do
+from apps.home.training.decorators import render_flatpage, mark_user
+from django.contrib.flatpages.views import flatpage
+
+_PURE_URL_NAME_TEMPLATE = '%s_pure'
+_MARK_PROGRESS_URL_NAME_TEMPLATE = '%s_mark_progress'
+_TRAINING_FINISHED_BOOLEAN_FIELD_TEMPLATE = 'training_finished_%s'
+
+
+class AbstractTrainingStep(object):
```
kdeloach: Nice.

kdeloach:
Looks good, merge away.

steventlamb:
cool. This is finally ready to merge, once the PR builder goes green.

PR #346, opened 2015-01-23
Add training quiz

kdeloach:
How to test this:

1. You should be able to access the intro quiz here: [http://localhost:8000/quiz/](http://localhost:8000/quiz/)
2. Completing a quiz should record your result here: [http://localhost:8000/admin/users/trainingresult/](http://localhost:8000/admin/users/trainingresult/)
3. Re-taking the quiz should preserve your highest score (If you score a 2, then re-take the quiz and score a 1, your score in the DB should remain a 2)

### DIFF
--- a/src/nyc_trees/apps/quiz/models.py
+++ b/src/nyc_trees/apps/quiz/models.py
@@ -11,6 +11,7 @@ class Quiz(object):
     passing_score - int; How many correct answers are needed to pass the quiz
     """
     def __init__(self, title, questions, passing_score):
+        assert len(questions) > 0
         assert passing_score <= len(questions)
         self.title = unicode(title)
         self.questions = list(questions)
@@ -26,7 +27,7 @@ class Quiz(object):
         for field_name, values in post_fields.iteritems():
             # Parse field name format like "question.1", "question.2", etc.
             parts = field_name.split('.')
-            if len(parts) == 2:
+            if len(parts) == 2 and parts[0] == 'question':
                 try:
                     value = int(values[0])
                     question_index = int(parts[1])
@@ -54,6 +55,7 @@ class Question(object):
     choices - iterable<string>
     """
     def __init__(self, text, answer, choices):
+        assert len(choices) > 0
         assert answer >= 0
         assert answer < len(choices)
         self.text = unicode(text)
--- a/src/nyc_trees/apps/quiz/models.py
+++ b/src/nyc_trees/apps/quiz/models.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+
+class Quiz(object):
+    """
+    title - string
+    questions - iterable<Question>
+    passing_score - int; How many correct answers are needed to pass the quiz
+    """
+    def __init__(self, title, questions, passing_score):
+        assert passing_score <= len(questions)
+        self.title = unicode(title)
+        self.questions = list(questions)
+        self.passing_score = int(passing_score)
+
+    @classmethod
+    def extract_answers(cls, post_fields):
+        """
+        Return dict of {question_index => answer_index, ...} from post fields.
+        post_fields - QueryDict; Probably from request.POST
+        """
+        result = []
+        for field_name, values in post_fields.iteritems():
+            # Parse field name format like "question.1", "question.2", etc.
+            parts = field_name.split('.')
+            if len(parts) == 2:
+                try:
+                    value = int(values[0])
+                    question_index = int(parts[1])
+                    result.append((question_index, value))
+                except ValueError:
+                    pass
+        return dict(result)
+
+    def score(self, answers):
+        """
+        How many correct answers are selected?
+        answers - dict of {question_index => answer_index, ...}
+        """
+        result = 0
+        for i, question in enumerate(self.questions):
+            if answers.get(i, -1) == question.answer:
+                result += 1
+        return result
+
+
+class Question(object):
+    """
+    text - string
+    answer - int; Index of the correct answer in choices iterable
+    choices - iterable<string>
+    """
+    def __init__(self, text, answer, choices):
+        assert answer >= 0
+        assert answer < len(choices)
+        self.text = unicode(text)
+        self.answer = int(answer)
+        self.choices = list(choices)
+
+
+# The keys represent the quiz_slug part of urls.
+quizzes = {
+    'intro_quiz': Quiz(
+        title='Introduction Quiz',
+        passing_score=2,
+        questions=(
+            Question(
+                text='Lorem ipsum dolor sit amet, consectetur '
+                     'adipiscing elit?',
+                answer=0,
+                choices=('Answer A', 'Answer B', 'Answer C')),
+            Question(
+                text='Lorem ipsum dolor sit amet, consectetur '
+                     'adipiscing elit?',
+                answer=1,
+                choices=('Answer A', 'Answer B', 'Answer C')),
+            Question(
+                text='Lorem ipsum dolor sit amet, consectetur '
+                     'adipiscing elit?',
+                answer=2,
+                choices=('Answer A', 'Answer B', 'Answer C')),
+        )
+    )
+}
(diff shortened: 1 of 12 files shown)


kdeloach:
Early reviews welcome, I'm writing unit tests now.

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+
+class Quiz(object):
```
steventlamb: making a separate app for quizzes is probably a good thing, and using a class to model a quiz is also probably a good thing. However, I think having a `models.py` that lives in a django `app` but does not contain django models is a recipe for confusion, given that `models.py` is such a ubiqutious convention in django.

In a PR for a similar feature, I made a `training` python package inside the `home` app, but it probably deserves its own app. I also ran into this `models.py` problem, and I chose `types.py`, but I can't say I feel great about that choice of name.

kdeloach: These models used to live in a module called `quizzes.py` but I moved it here to mimic how achievements are defined (Ref: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/users/models.py#L15)

kdeloach: But I do agree it could be confusing.

kdeloach:
I updated the description with instructions on how to test.

src/nyc_trees/apps/quiz/models.py
maurizi: I put `AchievementDefinition` in `users/models.py` because they very directly relate to the `Achievement` Django Model.

Since that does not apply here, that might not be the right idea.  Should this live in the same file as `QuizResult`?

kdeloach: The Django model closest related to this is `TrainingResult` in `apps.users.models`. I don't see much value in moving these but I'm open to suggestions.

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
...
+        post_fields - QueryDict; Probably from request.POST
+        """
+        result = []
+        for field_name, values in post_fields.iteritems():
+            # Parse field name format like "question.1", "question.2", etc.
+            parts = field_name.split('.')
+            if len(parts) == 2:
+                try:
+                    question_index = int(parts[1], 10)
+                    value = int(values[0])
```
kdeloach: TODO swap lines

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    def __init__(self, title, questions, passing_score):
+        assert passing_score <= len(questions)
+        self.title = unicode(title)
+        self.questions = list(questions)
+        self.passing_score = int(passing_score)
+
+    @classmethod
+    def extract_answers(cls, post_fields):
+        """Return dict of {answer_index => selected_choice_index, ...}
```
kdeloach: TODO comment

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
...
+                try:
+                    question_index = int(parts[1], 10)
+                    value = int(values[0])
+                    result.append((question_index, value))
+                except ValueError:
+                    pass
+        return dict(result)
+
+    def score(self, answers):
+        return sum(1 for i, question in enumerate(self.questions)
```
kdeloach: TODO comment

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
...
+            Question(
+                text='Lorem ipsum dolor sit amet, consectetur '
+                     'adipiscing elit?',
+                answer=1,
+                choices=('Answer A', 'Answer B', 'Answer C')),
+            Question(
+                text='Lorem ipsum dolor sit amet, consectetur '
+                     'adipiscing elit?',
+                answer=2,
+                choices=('Answer A', 'Answer B', 'Answer C')),
```
kdeloach: TODO remove trailing comma

src/nyc_trees/js/src/quiz.js
```diff
@@ -0,0 +1,58 @@
...
+    if (validate(id)) {
+        hide(id);
+        show(id + 1);
+    }
+}
+
+function prev(e) {
+    /*jshint validthis:true */
+    var id = +$(this).data('question');
+    $(dom.warning).addClass('hidden');
```
kdeloach: TODO not needed?

src/nyc_trees/apps/quiz/models.py
```diff
@@ -0,0 +1,79 @@
...
+        """Return dict of {answer_index => selected_choice_index, ...}
+        post_fields - QueryDict; Probably from request.POST
+        """
+        result = []
+        for field_name, values in post_fields.iteritems():
+            # Parse field name format like "question.1", "question.2", etc.
+            parts = field_name.split('.')
+            if len(parts) == 2:
+                try:
+                    question_index = int(parts[1], 10)
```
steventlamb: Isn't `10` the default?

src/nyc_trees/apps/quiz/models.py
steventlamb: fwiw, this "bare generator" syntax has a tendency to confuse developers. Consider putting a set of `[]` to make it an intermediate list via comprehension, since the size of `self.questions` will always be small.

more context:
http://python-history.blogspot.com/2010/06/from-list-comprehensions-to-generator.html (search for 'syntax')

src/nyc_trees/apps/quiz/models.py
steventlamb: I like the trailing comma on the last member of a dict. No need to worry about special cases, and makes diffs a little cleaner when they are modified. It's not a syntax problem in python.

steventlamb:
+1 works as expected. I tried making another quiz and it was easy to do. I tried retaking a quiz a few times and the best result persists in the `TrainingResult` object, but duplicate objects are not created.

Right now there's no indication of past success or failure when you return to the quiz list page and restart a quiz. There's no persistence of which questions were already answered correctly, which may (or may not) be in the spec at this point. This is stuff that will need to be addressed but probably need not block this PR from merging.

src/nyc_trees/apps/quiz/models.py
kdeloach: Thanks, I have JavaScript on the brain.

src/nyc_trees/apps/quiz/models.py
kdeloach: I got carried away with code golf here. I'll expand this.

kdeloach:
Updated with minor revisions and added some basic unit tests. We should wait until the real training quiz data exists before writing UI tests.

src/nyc_trees/apps/quiz/templates/quiz/quiz_complete_page.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% if passed_quiz %}
+<h3>Great job!</h3>
+<p>You answered {{ score }}/{{ quiz.questions|length }}
+    question{{ quiz.questions|length|pluralize }} correctly.</p>
+{# TODO: Where should this button go after completing the quiz? #}
+<p><a href="{% url 'training' %}" class="btn btn-primary">Finish</a></p>
+
+{% else %}
+<h3>Better luck next time...</h3>
```
RickMohr: I'd ask @jfrankl and/or @dboyer for what text to use if the user fails the quiz.

src/nyc_trees/apps/quiz/templates/quiz/quiz_page.html
```diff
@@ -0,0 +1,77 @@
...
+        </div>
+    </div>
+{% endblock aside %}
+
+{% block main %}
+
+<form method="POST" action="" class="quiz">
+{% csrf_token %}
+{% for question in quiz.questions %}
+<div class="question
```
RickMohr: Simpler would be `{% if forloop.first %}active{% else %}hidden{% endif %}"`

RickMohr:
+1

src/nyc_trees/apps/quiz/templates/quiz/quiz_page.html
kdeloach: Thanks, updated.

PR #347, opened 2015-01-23
Feature/progress style

jfrankl:


jfrankl:
On mobile, modal functions like normal:

image

On desktop, it expands to a sidebar or panel that is always open:

image

src/nyc_trees/sass/partials/_globals.scss
```diff
@@ -299,3 +298,40 @@ label {
 .map-small {
   padding-top: 40%; // Makes height 40% of width
 }
+
+.map {
```
maurizi: I introduced a `partials/_map.scss` file recently, with a `map` class.  We should either move this class there or removing the `_map.scss` file.

src/nyc_trees/sass/partials/_globals.scss
```diff
@@ -299,3 +298,40 @@ label {
...
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+}
+
+@media (min-width: 768px) {
```
maurizi: Should this use a `$screen-` variable?

src/nyc_trees/sass/partials/_globals.scss
jfrankl: I added this to _map.scss.

maurizi:
When I tested this on chrome, the legend prevents clicking on the map in the area below it.

werid_slice

jfrankl:
@maurizi, I fixed that glitch. Let me know if you have anything else.

maurizi:
+1

maurizi:
I don't think it has to be fixed in this PR, but as an FYI this does mess up the styling a bit on the reservations map, which currently shares `legend.html` (though that page needs styling work right now anyways)

We'll need to find a way to change the text on the reservations page (maybe inherit from a shared template and override a block?)
screenshot from 2015-01-25 14 41 30

PR #348, opened 2015-01-23
Minor styling changes

jfrankl:


jwalgran:
On an iPhone 6 the follow buttons are floating in the middle of the ``/group/`` page. Is that by design?

screen shot 2015-01-26 at 11 06 19 am

jwalgran:
+1 if the "Follow" styling is correct

jfrankl:
Good catch. I need to add some default width styling on that element for when groups have no description or very short ones.

Issue #303, opened 2015-01-14
Add cache header to tiles


maurizi:
Tiles generated by Windshaft have a `cache_buster` argument in their URL based on the most recent updated date of the tables used to generate them.

This should allow us to cache every tile forever.

maurizi:
@hectcastro This is done right?

hectcastro:
Headers are in place. Not sure how the cache busting works though:

```bash
$ curl -I https://d3a6cg2duujf6d.cloudfront.net/1424975281/nyc_trees/progress/9/152/193.png
HTTP/1.1 200 OK
Content-Type: image/png
Content-Length: 334
Connection: keep-alive
Access-Control-Allow-Headers: X-Requested-With, X-Prototype-Version, X-CSRF-Token
Access-Control-Allow-Origin: *
Cache-Control: max-age=315360000
Date: Thu, 26 Feb 2015 19:18:40 GMT
Expires: Thu, 31 Dec 2037 23:55:55 GMT
Server: nginx/1.6.2
X-Powered-By: Express
X-Windshaft-Cache: 59491
X-Cache: Miss from cloudfront
Via: 1.1 95f35fc8c27a19d2bdc2350557a8592e.cloudfront.net (CloudFront)
X-Amz-Cf-Id: ImKzqb199pu0O1wR6g3Hx3s_ZeR7W7DO6m0PqL5m4TerkuPTMp9fZQ==
```

maurizi:
@hectcastro The first part of the URL, (what is in your example `1424975281`), is a unix-style timestamp of the maximum updated datetime of all the tables used in the SQL query used to generate the tiles.

I think we can close this.

PR #354, opened 2015-01-26
Add HTTP caching support for tiles and static assets

### LINKED ISSUE #302: Add cache headers to static files when in production mode
When in production mode we add a md5sum to the filename for all JS, CSS and image files.

This should allow us to tell browsers to cache these files forever.

hectcastro:
This changeset adds HTTP caching support to production deployments by:

- Enabling the Cache-Control, Expires, and ETag headers for static assets
- Enabling the Cache-Control and Expires headers for tiles

It is worth noting that when the tiler is fronted by CloudFront, CloudFront will choose the greater of Cache-Control vs. 24 hours, but reserves the right to purge infrequently accessed assets from edge locations to make way to more frequently accessed ones.

Attempts to resolve #302 and #303.

### DIFF
--- a/deployment/ansible/roles/nyc-trees.app/templates/nginx-nyc-trees-app.conf.j2
+++ b/deployment/ansible/roles/nyc-trees.app/templates/nginx-nyc-trees-app.conf.j2
@@ -19,10 +19,20 @@ server {
   access_log /var/log/nginx/nyc-trees-app.access.log logstash_json;
 
   location /static/ {
+    {% if ['packer'] | is_in(group_names) -%}
+    etag on;
+    expires max;
+    {% endif %}
+
     alias {{ app_static_root }};
   }
 
   location /media/ {
+    {% if ['packer'] | is_in(group_names) -%}
+    etag on;
+    expires 1h;
+    {% endif %}
+
     alias {{ app_media_root }};
   }
 
--- a/deployment/ansible/roles/nyc-trees.tiler/templates/nginx-nyc-trees-tiler.conf.j2
+++ b/deployment/ansible/roles/nyc-trees.tiler/templates/nginx-nyc-trees-tiler.conf.j2
@@ -17,6 +17,10 @@ server {
   access_log /var/log/nginx/nyc-trees-tiler.access.log logstash_json;
 
   location / {
+    {% if ['packer'] | is_in(group_names) -%}
+    expires max;
+    {% endif %}
+
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $http_host;
     proxy_redirect off;


deployment/ansible/roles/nyc-trees.app/templates/nginx-nyc-trees-app.conf.j2
```diff
@@ -19,10 +19,20 @@ server {
...
+    expires max;
+    {% endif %}
+
     alias {{ app_static_root }};
   }
 
   location /media/ {
+    {% if ['packer'] | is_in(group_names) -%}
+    etag on;
+    expires 1h;
```
jwalgran: Was a media expiration time of "1h" chosen for a specific reason (best practice, previous experience, etc.)?

jwalgran:
I tried testing this by adding ``"packer"`` to the ``ANSIBLE_GROUPS`` in my  ``Vagrantfile`` like this:

```ruby
ANSIBLE_GROUPS = {
  "app-servers" => [ "app", "packer" ],
  "tile-servers" => [ "tiler", "packer" ],
  "services" => [ "services" ],
  "monitoring-servers" => [ "services" ]
}
```

I ran a ``vagrant provision tiler`` but I did not see the additional stanza in ``/etc/nginx/sites-available/nyc-trees-tiler.conf``

What is the correct way to get ``packer`` into the ``group_names`` variable accessible by the .j2 templates?

deployment/ansible/roles/nyc-trees.app/templates/nginx-nyc-trees-app.conf.j2
hectcastro: This was just a conservative number relative to the aggressive `max`. I also think I went with something conservative because I don't have a good idea of what is going to be served out of here yet.

hectcastro:
I was testing by inverting the conditional in the templates manually.

I haven't mucked with the `Vagrantfile` to simulate Packer provisioning yet, but something like this should work:

```ruby
ANSIBLE_ENV_GROUPS = {
  "packer:children" => [
    "app-servers", "tile-servers", "services"
  ]
}
```

RickMohr:
Perhaps we should consider a more direct way of determining `deploy_target == production`? 

`if ['packer'] | is_in(group_names)` is potentially fragile, and strange to simulate for testing.

jwalgran:
+1
I had trouble simulating "packer" mode so I inverted the ``is_in`` checks and verified that configs have the additional keys.

hectcastro:
In the provisioning steps, `packer` is used to unlock the things that need to be done to produce an AMI. I went with `packer` because [Packer](https://www.packer.io/) is the tool we're using to generate the AMI. Also, the goal with these AMIs is that they can run in staging and production without being modified.

What would people like to see as a better alternative?

PR #359, opened 2015-01-26
Reservation selection

maurizi:
This is only the client-side portion of reserving blockfaces.
More work will need to be done in a future PR to post the blockface ids
to the server in order to reserve them.

Selected blockfaces will have a vector line laid overtop of them to
indicate that they are currently selected.

This was abstracted into `selectableBlockfaceLayer.js` for reuseability with
other pages where we will need to show the currently selected blockface(s)

### DIFF
--- a/src/nyc_trees/js/src/reserveBlockfacePage.js
+++ b/src/nyc_trees/js/src/reserveBlockfacePage.js
@@ -1,10 +1,20 @@
 "use strict";
 
-var MapModule = require('./map'),
+var mapModule = require('./map'),
     zoom = require('./mapUtil').zoom,
-    L = require('leaflet');
+    selectableBlockfaceLayer = require('./selectableBlockfaceLayer'),
+    L = require('leaflet'),
+    $ = require('jquery'),
 
-var reservationMap = MapModule.create({
+    dom = {
+        currentReservations: "#current-reservations",
+        totalReservations: "#total-reservations"
+    };
+
+// Extends the leaflet object
+require('leaflet-utfgrid');
+
+var reservationMap = mapModule.create({
     geolocation: true,
     legend: true,
     search: true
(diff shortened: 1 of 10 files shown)


src/nyc_trees/js/src/selectableBlockfaceLayer.js
```diff
@@ -0,0 +1,102 @@
...
+ * adding or removing geoJSON.
+ *
+ * If the callbacks do not return a truthy value, the feature will not be
+ * added/removed.
+ *
+ * Returns the created geojson layer
+ */
+function createBlockfaceLayer(options) {
+    var map = options.map,
+        gridUrl = options.gridUrl,
```
maurizi: Note to self: It is probably better to take a `UtfGrid` object, so other features on the grid can be interacted with (like showing popups for unavailable blockfaces)

src/nyc_trees/js/src/selectableBlockfaceLayer.js
```diff
@@ -0,0 +1,102 @@
...
+        onAdd = options.onAdd || alwaysTrue,
+        onRemove = options.onRemove || alwaysTrue,
```
kdeloach: Consider defaulting to `$.noop`? Doing nothing unless `onAdd`/`onRemove` are specified seems like a reasonable default to me.

src/tiler/sql/user_reservable.sql
```diff
@@ -1,24 +1,33 @@
...
-        AND reservation.canceled_at IS NULL
-        AND reservation.expires_at > now() at time zone 'utc')
-  LEFT OUTER JOIN users_trustedmapper AS trustedmapper
-    ON (turf.group_id = trustedmapper.group_id
-        AND trustedmapper.user_id = <%= user_id %>)
+    <% if (is_utf_grid) { %>
+    CASE
+      WHEN survey_type = 'available' THEN ST_AsGeoJSON(geom)
+      ELSE NULL
+    END AS geojson,
```
kdeloach: Consider moving this column to the original query instead of using a subquery.

maurizi: I tried that to begin with.

You can't reference a computed column like `survey_type` from within the same query, which was what prompted me to move the bulk of this SQL statement into a subquery.

src/nyc_trees/js/src/reserveBlockfacePage.js
```diff
@@ -13,3 +23,35 @@ var reservationMap = MapModule.create({
...
+    map: reservationMap,
+    gridUrl: config.urls.layers.reservable.grids,
+
+    onAdd: function(gridData) {
+        if (selectedBlockfacesCount < blockfaceLimit && gridData.restriction === 'none') {
+            selectedBlockfaces[gridData.id] = gridData;
+            selectedBlockfacesCount++;
+            $current.text(selectedBlockfacesCount);
+
+            return true;
```
kdeloach: Blockface selection appeared to suddenly stop working after I had chosen a few. It took me a minute to realize that I had clicked on the maximum number of blockfaces that I was allowed to reserve. Consider some kind of notification message for all cases that cause `onAdd` to return false.

src/tiler/sql/user_reservable.sql
kdeloach: Totally overlooked that this references a generated column. Makes sense.

kdeloach:
+1

kdeloach:
I'll test this on a few mobile devices tomorrow.

kdeloach:
Works great on my mobile. I have 2 observations which I will log as issues.

src/nyc_trees/js/src/reserveBlockfacePage.js
maurizi: Added issue #366 to tackle this in a later PR

PR #360, opened 2015-01-27
Whitelist views for soft launch

jwalgran:
We want to allow user registration before the census begins. The middleware in this PR allows for defining a whitelist of allowable URL regexes and redirecting any request to a non-matching URL to a common page.

We will also need to show alternate content on some pages to avoid linking to or referencing unavailable content or functionality, so the PR also includes a context processor to make a ``soft_launch`` boolean available to all templates.

screen shot 2015-01-26 at 5 38 18 pm

screen shot 2015-01-26 at 5 38 26 pm


maurizi:
+1

I'm a little concerned that `SOFT_LAUNCH_ENABLED` defaulting to `True` will get annoying during user testing (we don't really have a distinction between production and staging), but we can easily work around it if it is an issue.

PR #361, opened 2015-01-27
Add explicit method signature to custom filters

hectcastro:
This changeset adds an explicit method signature to the Ansible custom filters.


kdeloach:
+1

PR #377, opened 2015-01-27
My reservations action bar

maurizi:
An HTML partial is loaded via AJAX when a blockface is selected, and
used to populate the actionbar at the bottom of the page.
selectable

### DIFF
--- a/src/nyc_trees/apps/survey/routes.py
+++ b/src/nyc_trees/apps/survey/routes.py
@@ -24,6 +24,11 @@ reserve_blockface_page = route(
         render_template('survey/reserve_blockface.html'),
         v.reserve_blockfaces_page))
 
+reserved_blockface_popup = route(
+    GET=individual_mapper_do(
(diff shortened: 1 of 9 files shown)


maurizi:
This builds off of #369.  The only pertinent commit here is the last one.

jwalgran:
Looking at this now

jwalgran:
+1 this works well

I would dig it if the selected lines were thicker

screen shot 2015-01-28 at 11 21 31 am

screen shot 2015-01-28 at 11 21 42 am


```javascript
// Note: this must be kept in sync with src/tiler/style/progress.mss
// Line widths are purposefully 6 pixels wider than the tiler styling
function getLineWidth(zoom) {
    if (zoom >= 19) {
        return 22;
    } else if (zoom === 18) {
        return 14;
    } else if (zoom === 17) {
        return 10;
    } else {
        return 8;
    }
}

PR #378, opened 2015-01-28
Move map search control container

### LINKED ISSUE #350: Change DOM location of location search
Currently the location search is nested in a div with the classes '''.leaflet-top''' and '''.leaflet-left'''. Those classes prevent me from positioning it correctly on the page. Would it be possible to bump it up a level?

jwalgran:
Moving the search control container to be a direct child of the Leaflet control container, rather than making it an full Leaflet control, makes styling the search controls easier.

screen shot 2015-01-28 at 9 47 25 am

Fixes #350


jfrankl:
Here's the current state:

image

Looking okay. I have a request for you, Justin. I'd like for the ```location-search-status``` element to be the first sibling of the ```location-search-result``` element. Also, I want it to receive a class that states whether or not results were found (```results-found``` or ```results-not-found```).

jfrankl:
New idea: how about if I merge this and I put that previous request in an issue?

jwalgran:
I have implemented the changes you asked for and will add them to this PR ASAP. Sorry for the delay.

jwalgran:
@jfrankl better late than never.

Here is a bad search

screen shot 2015-02-02 at 10 35 02 pm

Here is a search with multiple results

screen shot 2015-02-02 at 10 34 19 pm

jfrankl:
Looks good on mobile:

image

Trouble on desktop:

image

The desktop issue is a little more difficult than a standard "put a z-index on it" issue. What's happening is that the search is a child of the map. We want the overlay to appear under the search but on top of the map. As far as I know, that's really tough to fix.

A few ideas. Wanted to see if you have others:

- We can move the search box to a different location on desktop. That's not going to be awesome visually, but it will pass.
- Maybe we could have two search box elements, one that is a child of the map and one that is not. They'd function exactly the same, but one would be hidden on desktop and one would be hidden on mobile.

jwalgran:
@jfrankl What if we take a cue from the legend and, when we have a large enough viewport, we always show the search box. Is that what you were thinking about when you said "We can move the search box to a different location on desktop"?

search-in-toolbar

jfrankl:
I moved the search to a better positioning. Still not perfect, but I think this is good enough for now. The desktop map page might change layout in the next couple of weeks, so I want to move forward for now and come back to this later if there is time.

Take a look and let me know what you think.

image

jwalgran:
+1
Looks great. I agree that we should roll with this design for now.

PR #382, opened 2015-01-28
Remove redundant css that is also in _map.scss

jfrankl:
Not sure how these got back in globals, but this all needs to get deleted.


maurizi:
+1

PR #383, opened 2015-01-28
Target OpenJDK version 7u75 across project

hectcastro:
This changeset ensures that we use OpenJDK 7u75 across the project. In addition, it should mitigate minor version package issues that crop up.


steventlamb:
+1 I checked this out and reprovisioned with success. I still have old versions of java on my `app` and `tiler` VM, but that's to be expected.

PR #384, opened 2015-01-28
Fix bug with new reservations counts

maurizi:
The "current" reservations should reflect how many blockfaces are
selected on the page, not how many blockfaces are reserved.


kdeloach:
+1

PR #385, opened 2015-01-28
Save/restore blockface reservations

kdeloach:
This creates a reusable module for saving/loading stuff to localStorage and implements saving/loading state on the reserve blockfaces page.


src/nyc_trees/js/src/lib/storage.js
```diff
@@ -0,0 +1,52 @@
+"use strict";
+
+var $ = require('jquery'),
+    storage = window.localStorage || {};
+
+
+function Storage(options) {
```
kdeloach: I will probably rename this.

kdeloach: TODO: Documentation

src/nyc_trees/js/src/lib/storage.js
```diff
@@ -0,0 +1,52 @@
...
+    }, options);
+
+    function save() {
+        storage[opts.key] = serialize();
+    }
+
+    function load() {
+        var state = deserialize(storage[opts.key]);
+        if (state) {
+            opts.validate(state);
```
maurizi: Letting the exceptions from `validate` bubble up isn't the best thing...

Technically the `LocalStorage` could be edited by a user (though it seems like a low probability), or a bug in our code could make saved state no longer validate.

It's probably more robust to wrap this in a try/catch, and if there are any exceptions, log it and delete whatever was saved.

kdeloach: I'm in favor of throwing here and letting the consumer decide how to cleanup the mess. I'll have to think about it.

kdeloach:
@azavea-bot rebuild

kdeloach:
I refactored the selectable blockface module to be its own Leaflet layer. Take another look when you get the chance.

maurizi:
Since `selectableBlockfaceLayer.js` is now a class that gets instantiated with `new`, it should be renamed to start with a capital letter (`SelectableBlockfaceLayer.js`)

maurizi:
We need to update [`reservationPage.js`](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/reservationPage.js#L33-L45) to accomodate the changes in `selectableBlockfaceLayer.js`

maurizi:
Overall this looks really nice, modulo the 2 comments above.

+1 once those are addressed.

kdeloach:
Done and done. Merging now.

Issue #367, opened 2015-01-27
Styling issue with legend


maurizi:
![legend](https://cloud.githubusercontent.com/assets/4432106/5922540/6d298ae2-a61a-11e4-8c2e-2c617952417c.png)

maurizi:
Seems to be an issue on the progress map, reservations map, and new reservations map

kdeloach:
I didn't see this issue before #353 was merged but I'm seeing it now.

PR #386, opened 2015-01-28
Feature/map action fix

jfrankl:
This styles the events and new reservation map pages. The action bar will appear fixed on the bottom of the page on mobile and in the sidebar on desktop. Also, for the event page, I needed to add some javascript to put a class on a container element based on selected tabs. There might be a better way to achieve that goal.


jfrankl:
image

image

image

image

image

maurizi:
I would squash these two commits together.

I think the JS could use some work, but one of the developers can fix it up at a later time (can you put a `TODO` comment above it?).

maurizi:
Also, you should make a note in the commit message that this fixes #367

src/nyc_trees/sass/partials/_map.scss
```diff
@@ -17,6 +47,16 @@
   height: 400px;
 }
 
+.map-event {
+  // margin-right: -$grid-gutter-width/2;
+  // margin-left: -$grid-gutter-width/2;
```
maurizi: You should remove this commented out CSS

jfrankl:
Thanks for the feedback Mike. This should be good to go.

maurizi:
I opened #412 to track the issue we discussed on the event list page.

PR #387, opened 2015-01-28
Topic/integrate quiz into training

steventlamb:


src/nyc_trees/apps/quiz/views.py
```diff
@@ -22,19 +24,28 @@ def quiz_page(request, quiz_slug):
...
 
     result, created = TrainingResult.objects.get_or_create(
-        user_id=request.user.id,
+        user_id=user.id,
         module_name=quiz_slug)
 
     result.score = max(result.score, score)
     result.save()
 
+    passed_quiz = (score >= quiz.passing_score)
```
kdeloach: Could we pass this variable to the context also?

kdeloach:
+1

src/nyc_trees/apps/quiz/views.py
steventlamb: woops! that was the point!

PR #388, opened 2015-01-29
Add glob to RELP package version

hectcastro:
This changeset sets the package version of RELP to the highest version that begins with `7.4.4`.


maurizi:
Looks good to me, though I haven't tested it.

maurizi:
I reprovisioned sucessfully and RELP was updated.
+1

PR #389, opened 2015-01-29
Fix pgweb startup sequence after boot

hectcastro:
This changeset sets the Upstart `start on` stanza for `pgweb` to depend on a `initctl` event emitted by the PostgreSQL `init.d` script.


kdeloach:
Nice. I was going to submit an issue for this but decided that the psql shell was a good enough workaround. This is more convenient though.

kdeloach:
+1

PR #390, opened 2015-01-29
Disable soft launch until better setting available

jwalgran:
We want to exercise the full application in our staging environment (which uses production settings) but can't with this flag set to True. The plan is to replace this simple settings flag with a database-backed system with finer grained control over "on" and "off."


hectcastro:
:+1:

PR #394, opened 2015-01-29
Topic/embedded training questions

steventlamb:


steventlamb:
This gets us really close to having embedded exercises. Right now question feedback is displayed in an alert, for demo purposes.

steventlamb:
In addition to finishing the feature according to the spec, when decided, documentation is also necessary.

jwalgran:
looking at this now

jwalgran:
+1 works well

hectcastro:
Can we include something like this at the bottom of `setupdb.sh` as part of this PR?

```bash
# Create training flatpages
envdir /etc/nyc-trees.d/env/ /opt/app/manage.py make_training_flatpages || true
```

steventlamb:
@hectcastro that would work, or we can just change this mgmt command to return zero when records exist, instead of failing.

steventlamb:
Pushed 3 commits to finish this feature according to @jfrankl's sketches.

kdeloach:
+1

PR #395, opened 2015-01-29
Fix SES/Boto mailer

hectcastro:
This changeset fixes the custom e-mail backend we're using to send e-mail from Django through AWS SES.


jwalgran:
+1

PR #397, opened 2015-01-29
Remove quiz listing page

kdeloach:
No need for this until we get more than one quiz.


src/nyc_trees/apps/quiz/urls.py
```diff
@@ -3,16 +3,10 @@
 from __future__ import unicode_literals
 from __future__ import division
 
-from django.conf.urls import patterns, url
-
-from apps.quiz.routes import quiz_list_page, quiz_page
+from django.conf.urls import patterns
```
steventlamb: shouldn't you kill this whole file if it doesn't add any urls?

src/nyc_trees/apps/quiz/routes.py
```diff
@@ -11,10 +11,6 @@
 from apps.quiz import views as v
 
 
-quiz_list_page = do(login_required,
-                    render_template('quiz/quiz_list_page.html'),
-                    route(GET=v.quiz_list_page))
-
 quiz_page = do(login_required,
```
steventlamb: is this route still used anywhere besides `apps/home/training/routes` ?

src/nyc_trees/apps/quiz/views.py
```diff
@@ -5,17 +5,11 @@
...
 from apps.users.models import TrainingResult
 
 
-def quiz_list_page(request):
-    return {
-        'quizzes': quizzes.iteritems()
-    }
-
-
 def quiz_page(request, quiz_slug):
```
steventlamb: I'm pretty sure this route and view are now only used by `/home/training/routes/`, but it is used in a way that no longer makes sense. The training route just calls this with a hardcoded `quiz_slug`, but if we're not going to use a view for arbitrarily named quizzes, shouldn't we flatten that plumming out by removing this view and its route and moving all that code into training?

src/nyc_trees/apps/quiz/urls.py
kdeloach: This stub existed before I started working on quiz stuff. I don't have a strong opinion about its existence.

src/nyc_trees/apps/quiz/routes.py
kdeloach: Nope.

src/nyc_trees/apps/quiz/views.py
kdeloach: @jwalgran Should we squash all this quiz stuff into `apps.home.training` since that is the only place quizzes are referenced from presently?

jwalgran:
I'm like the idea of consolidating the quiz stuff into the training app and moving it back out in the unlikely event that we need to support more quizzes.

kdeloach:
Will do, thanks.

kdeloach:
Merging as-is unless there is a public outcry.

steventlamb:
fine by me.

kdeloach:
@steventlamb Not to spend any more time on this than we should, but I managed to merge the quiz app into training... I did encounter some strange circular dependency issues because of how the urls->types->views thing works so I ended up having to move some imports inside functions. I think this is a little ugly but everything works. What do you think?

steventlamb:
+1

Issue #392, opened 2015-01-29
No DEFAULT_FROM_EMAIL setting exists


hectcastro:
In the staging environment, this e-mail address needs to have an `azavea.com` domain. I'm not sure we have an e-mail address to use in the production environment, but when we do, it will need to be set with `DEFAULT_FROM_EMAIL` and be verified with Amazon SES.

jwalgran:
The question for Parks is: what should the "from" address be for automated email sent from TreesCount?

dboyer:
Parks is working on getting us an email address. They need to finalize it.

jwalgran:
To bootstrap user testing, set ``DEFAULT_FROM_EMAIL`` to ``azaveadev@azavea.com``

steventlamb:
@hectcastro This looks fixed, please confirm.

hectcastro:
I left it open as a reminder that we still need to get an e-mail address from Parks and wire it up to the application and SES.

steventlamb:
Got it

dboyer:
Emailed confirmed -  treescount.help@parks.nyc.gov

hectcastro:
Waiting for them to confirm via SES and then I'll wire up and test.

PR #410, opened 2015-01-30
Add DEFAULT_FROM_EMAIL

hectcastro:
This changeset adds a `DEFAULT_FROM_EMAIL` setting with the Azavea domain so that it is not rejected by Amazon SES.

Attempts to resolve: https://github.com/azavea/nyc-trees/issues/392


maurizi:
Looks good to me

PR #425, opened 2015-02-03
Add homepage and dashboard styling

jfrankl:
First pass at homepage and dashboard styling.

image


jfrankl:
Still more to do on this. Here are a few issues that I'm opening. I don't think we need to close these before merging this PR. #428 #427

jwalgran:
I am seeing some funky wrapping of the contributions and counter on the ``home.html``

screen shot 2015-02-02 at 10 48 50 pm

screen shot 2015-02-02 at 10 49 00 pm

jwalgran:
I opened #430 showing how we can move these two different root pages into partials. We can either move this styling work over to that PR, or update this PR and trash the other.

PR #431, opened 2015-02-03
Add field_training_complete field

kdeloach:
This field is set to True on the user model when they have been checked
into a training event.


maurizi:
+1

RickMohr:
+1

PR #432, opened 2015-02-03
Fix errors on reserve blockface page

kdeloach:
This appeared to work when I first tested it but Objects don't have a
length property, obviously. Also, suppress an error when clicking empty
spaces on the map.

### DIFF
--- a/src/nyc_trees/js/src/reserveBlockfacePage.js
+++ b/src/nyc_trees/js/src/reserveBlockfacePage.js
@@ -77,6 +77,9 @@ var selectedLayer = new SelectableBlockfaceLayer(reservationMap, grid, {
 });
 
 grid.on('click', function(e) {
+    if (!e.data) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


src/nyc_trees/js/src/reserveBlockfacePage.js
```diff
@@ -89,10 +92,14 @@ reservationMap.addLayer(selectedLayer);
...
         selectedLayer.addBlockface(data);
     });
-    // Zoom to extent of selected blockface reservations.
-    reservationMap.fitBounds(selectedLayer.getBounds());
+    try {
+        // Zoom to extent of selected blockface reservations.
+        reservationMap.fitBounds(selectedLayer.getBounds());
+    } catch (ex) {
+        // Ignore Leaflet fitBounds error when there are no selections.
+    }
```
maurizi: Consider instead doing:
```javascript
if (state && Object.keys(state).length > 0) {
    ...
}
```

maurizi:
+1

PR #433, opened 2015-02-03
Bug/229 more events failure

steventlamb:


kdeloach:
+1

jwalgran:
The link at the bottom of event detail is still targeting ``/group/{group-slug}/event/``. It should target ``/group/{group-slug}/#group-events``

screen shot 2015-02-03 at 1 29 35 pm

steventlamb:
closing in favor of https://github.com/azavea/nyc-trees/pull/439

jwalgran:
For completenes, there was a redirect in place, but the view was wrapped with ``group_admin_do``, so non-admins got a 403 instead of the redirect.

PR #438, opened 2015-02-03
Don't require group image, contact url, admin

### LINKED ISSUE #233: Make image and contact URL not required when adding a group via the Django admin
Some groups will be too small to have websites and the census admin setting up the group  may not have access to a logo image. The group administrator can add those items via the group edit page on the public facing side of the site.

steventlamb:
fixes #233 on github


maurizi:
+1

steventlamb:
Merging because it has a +1 and PRs with migrations are best merged quickly.

PR #439, opened 2015-02-03
Bug/227 more events failure

steventlamb:


src/nyc_trees/apps/event/views.py
```diff
@@ -23,8 +23,9 @@
 
 
 def event_dashboard(request):
-    # TODO: implement
-    pass
+    base_url = reverse('group_detail',
+                       kwargs={'group_slug': request.group.slug})
+    return HttpResponseRedirect(base_url + '#group-events')
```
maurizi: I don't think this is quite right.

The original idea of the `event_dashboard` from the proposal is that it is the place for admins to manage events, and I believe it has evolved into the group tab of the event management page.
This is why the route has a `group_admin_do` decorator on it.

It makes sense for "Find more events" link to go to the events section of the publicly visible group detail, but the events section is different from the group admin's event dashboard.

steventlamb: Correct, and also I am going to update this PR to rip out event dashboard entirely, because it has been replaced by the event tab of the group admin.

steventlamb:
@maurizi @jwalgran @kdeloach This has been totally overhauled to address the following:
1. Drop the redirect indirection, link directly to the events section of group detail
2. Rip out events dashboard entirely
3. Nullify the permission problem that turned up for anonymous users

jwalgran:
+1
Works perfectly. Nice extraction of the DOM id into a shared var.

PR #440, opened 2015-02-03
Require that users complete training before they can RSVP for an event

kdeloach:
This alters the `user_must_have_online_training` decorator and redirects users to a "training instructions" page if they try to RSVP for an event before they have completed training.

Refs #405

### DIFF
--- a/src/nyc_trees/apps/core/decorators.py
+++ b/src/nyc_trees/apps/core/decorators.py
@@ -11,7 +11,9 @@ from django.shortcuts import get_object_or_404
 
 from django_tinsel.utils import decorate as do
 
-from apps.core.helpers import user_is_census_admin, user_is_group_admin
+from apps.core.helpers import (user_is_census_admin, user_is_group_admin,
+                               user_has_online_training,
+                               user_has_field_training)
 from apps.core.models import Group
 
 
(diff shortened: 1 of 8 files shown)


jwalgran:
The "Attend a training event" link on the instructions begs for a link to ``/event/?q=training`` to apply the meaningful filter automatically. If it is to far out of scope to add that to this PR, I would like to add and issue to track it as an enhancement.

kdeloach:
Good idea. I'll add ids to those tabs so that `/event/#training` will select the correct tab.

jwalgran:
The "Complete your online training" link on the instructions page should be disabled/hidden/whathaveyou when the online training is complete

jwalgran:
I registered a new user and walked though all the training. The Django Admin shows all training boxes checked.

screen shot 2015-02-03 at 5 55 12 pm

When I try to RSVP for a training event, I am redirected to the instructions page. I think this is because https://github.com/azavea/nyc-trees/pull/440/files#diff-502e026b7bf08f56e0176383e168e5ebR16 is not taking the event count and the online vs. field training into account.

kdeloach:
Did the user you created, RSVP and then check-in to a training event first?

kdeloach:
All of these guys need to be checked:
image

kdeloach:
Sorry, I should read the entire comment before responding next time. You're totally right, and I'll make these corrections tomorrow!

kdeloach:
@jwalgran Fixed the RSVP button for training events and updated the training instructions template. I would like to create a separate card for deep linking into event list tabs.

Thanks for testing this thoroughly.

kdeloach:
TODO: Create issue for design work after merging.

jwalgran:
+1

* New user could not register for any event - redirected to instructions
* After completing online training, new user could RSVP to training events. Trying to RSVP to mapping events redirected to the instructions page with the training item struck through.
* After being checked in to a training event, the new user could RSVP for mapping events.

PR #444, opened 2015-02-04
Add CloudWatch alarms to the application and tiler ELBs

hectcastro:
This changeset adds CloudWatch alarms to the application and tiler Elastic Load Balancers (ELBs). The two  metrics right now are backend 4XX and 5XX failures.

### DIFF
--- a/deployment/troposphere/app_template.py
+++ b/deployment/troposphere/app_template.py
@@ -2,6 +2,7 @@ from troposphere import Template, Parameter, Ref, Tags, Output, GetAtt, ec2
 
 import template_utils as utils
 import troposphere.autoscaling as asg
+import troposphere.cloudwatch as cw
 import troposphere.elasticloadbalancing as elb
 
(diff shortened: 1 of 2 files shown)


jwalgran:
This is a good idea. Should we also be attaching these alarms to an SNS topic and attaching an email address to that topic? Does it make sense to reuse ``AWS_SNS_TOPIC``?

kdeloach:
+1

hectcastro:
`notification_arn_param` is the SNS topic ARN passed to CloudFormation. That gets wired up to all of these alarms.

Anyone who wants to subscribe to the topic can do so via the SNS tab on the AWS console. I also think that an internal mailing list is being formed for support rotation. Once that exists, I'll subscribe it too.

jwalgran:
Right. Sorry I missed that. Perfect.

maurizi:
+1

PR #449, opened 2015-02-04
Add DNS support for monitoring server

hectcastro:
This changeset gives the monitoring server a subdomain off of the default Route 53 hosted zone.

For example, if the default hosted zone is `treescount.foo.com`, these CloudFormation changes will create an A record of `monitoring.treescount.foo.com` under the `treescount.foo.com` hosted zone that points to the public IP address of the monitoring server.


jwalgran:
I was briefly confused because I forgot that monitoring == bastion. I am not sure if and where we should add a note about this naming equivalency to our deployment documentation, or if the troposphere scripts are enough.

hectcastro:
The diagram in https://github.com/azavea/nyc-trees/pull/458 should provide more clarity.

kdeloach:
+1

PR #451, opened 2015-02-04
Add documentation for applying migrations

hectcastro:
This changeset adds documentation for the process of applying migrations to an AWS stack of this project.


jwalgran:
+1
I ran through the process as documented and the migration command ran successfully. There was an error creating the flat pages, but that is because the current version of the management command is designed to fail, rather than overwrite data.

PR #452, opened 2015-02-05
Add profile widgets to the dashboard

jwalgran:
The dashboard shows much of the same content as the user profile page. I have extracted the context generation helpers out of the users view file so they can be more easily shared with the home app.

### DIFF
--- a/src/nyc_trees/apps/home/templates/home/partials/dashboard.html
+++ b/src/nyc_trees/apps/home/templates/home/partials/dashboard.html
@@ -24,50 +24,21 @@
 
     <div class="container">
         <div class="row">
-            {% if not user.online_training_complete %}
             <main class="col-sm-12">
-                <section class="your-contributions clearfix">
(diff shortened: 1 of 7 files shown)


RickMohr:
For testing consider fast template tests (like `ProfileTemplateTests`) rather than slower ui tests.

maurizi:
I'm not really a big fan of putting actual code into `__init__.py` files.

Consider moving that code into a `helpers.py` file (and possibly exposing it in `__init__.py` if you want to import it directly from `apps.users`).

maurizi:
I agree with @RickMohr, if you're only testing markup and not behaviour, I would prefer testing the templates produced by the routed views over UI tests.

maurizi:
Otherwise, looks good to me, +1

jwalgran:
tests

maurizi:
Continued +1 (modulo my comment about `__init__.py` above, which is a matter of personal taste)

PR #454, opened 2015-02-05
Style updates and git jpg handling fix

jwalgran:
I built on top of some styling commits from Jeff, fixing the .gitattributes file so that .jpg files are not treated as text, which prevents them from being corrupted on commit when Git attempts to normalize the line endings.


jfrankl:
+1

jfrankl:
This is awesome Justin aka Scooby Doo.

Issue #453, opened 2015-02-05
Provision fails (services VM) because postgres version not found


RickMohr:
Error message:
```
msg: 'apt-get install 'postgresql-9.3=9.3.5*.pgdg14.04+1' 'postgresql-server-dev-9.3=9.3.5*.pgdg14.04+1'' failed: E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-9.3/postgresql-client-9.3_9.3.5-2.pgdg14.04+1_amd64.deb  404  Not Found [IP: 87.238.57.227 80]

E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-9.3/postgresql-9.3_9.3.5-2.pgdg14.04+1_amd64.deb  404  Not Found [IP: 87.238.57.227 80]

E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-9.3/postgresql-server-dev-9.3_9.3.5-2.pgdg14.04+1_amd64.deb  404  Not Found [IP: 87.238.57.227 80]
```

It fails to download `postgresql-9.3_9.3.5-2.pgdg14.04+1_amd64.deb` from
http://apt.postgresql.org/pub/repos/apt/pool/main/p/postgresql-9.3/

That file is in fact absent on that server. Perhaps we want 
`postgresql-9.3_9.3.6-1.pgdg14.04+1_amd64.deb`, which is there.

So in `nyc-trees/deployment/ansible/group_vars/all` I tried changing to 
`postgresql_package_version: "9.3.6*.pgdg14.04+1"`

Now the error is:

msg: 'apt-get install 'postgresql-9.3=9.3.6*.pgdg14.04+1' 'postgresql-server-dev-9.3=9.3.6*.pgdg14.04+1'' failed: E: Version '9.3.6*.pgdg14.04+1' for 'postgresql-9.3' was not found

PR #455, opened 2015-02-05
Unpin minor version of PostgreSQL related packages

hectcastro:
This changeset extends the glob for the version number of PostgreSQL related packages so that the minor version is included. Originally, the minor version was being included because the version of PostgreSQL being used in RDS is `9.5.3`.

Attempts to resolve: #453


maurizi:
+1, I trust the Postgresql folks to do a good job with semver

RickMohr:
+1 Successfully provisioned new services and app VMs. Thanks.

PR #460, opened 2015-02-05
Add polling to user event checkin page

kdeloach:
This will poll an endpoint every few seconds to determine if the user has
been checked-in by an admin yet. When the user is checked-in they will
be redirected to the event detail page.


kdeloach:
@RickMohr Updated. Though I'm still not sure where "Start Mapping" is supposed to go.

RickMohr:
+1

maurizi:
+1

kdeloach:
Thanks @RickMohr, I know you have a lot of other stuff going on this week.

PR #461, opened 2015-02-05
Fix actions in event dropdowns on group admin "Events" tab

### LINKED ISSUE #408: Some options don't work on event drop-down on groups page
Testing on staging. 
event_dropdown

At 9:20, I set up an event to run from 9:10-10:10. Since it was in progress, I assumed I could use the drop-down arrow for that event on my groups page to email attendees, RSVP, and check-in for the event. See screenshot.

First issue is that the event was put in Past even though it was still occurring. Seems like events shouldn't move to the past tab until after the end time for the event is past.

Second issue is that nothing happened when I clicked RSVP, email attendees (perhaps because email isn't set up on staging), or event check-in. It seems like I should at least be taken to the event check-in page when I click that option.

RickMohr:
Fixes #408

### DIFF
--- a/src/nyc_trees/apps/event/event_list.py
+++ b/src/nyc_trees/apps/event/event_list.py
@@ -69,6 +69,7 @@ class EventList(object):
                                         .values_list('event_id', flat=True))
 
         return [{'event': event,
+                 'event_share_url': event.get_shareable_url(request),
                  'user_is_registered': event.pk in user_registered_event_ids}
                 for event in qs]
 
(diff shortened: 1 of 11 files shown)


kdeloach:
On the Group Edit page, "RSVP to event" still exists in the event dropdown but doesn't do anything. The "Get shareable link" option works.

The "shareable like" option doesn't work for the event dropdown on the Group Detail page.

image

On the Event Detail page, consider wiring up the Share Link "Copy" button with your copy to clipboard module (unless we are planning to come up with a more elegant solution for non-admins).

image

kdeloach:
+1

RickMohr:
@azavea-bot rebuild

PR #462, opened 2015-02-05
Fix event list styling issues

maurizi:


src/nyc_trees/js/src/event_list_page.js
```diff
@@ -64,12 +66,15 @@ $(dom.mapToggle).one('shown.bs.tab', function() {
...
-        return (css.match (/(^|\s)tab-status-\S+/g) || []).join(' ');
-    });
-    $('.page-map').addClass('tab-status-'+selectedTab);
+    var selectedTab = $(e.currentTarget).attr('href');
+    $(dom.mapContainer).removeClass('tab-status-list tab-status-map-tab');
+
+    if (selectedTab === '#map-tab') {
+        $(dom.mapContainer).addClass('tab-status-map-tab');
+    } else if (selectedTab === '#list') {
+        $('.page-map').addClass('tab-status-list');
```
kdeloach: Consider using `dom.mapContainer` instead of `.page-map`.

maurizi: Oops, good catch!

kdeloach:
+1

steventlamb:
+1, tested, works

steventlamb:
Scratch that, does not work, but is not necessarily the result of this pull. I'm looking into this, and will update as I go.

steventlamb:
Opened a PR on @maurizi's branch that should make this work.

PR #463, opened 2015-02-05
Clean up z-index values to be lower than Bootstrap's

maurizi:
I was having a lot of trouble getting Bootstrap modals to work properly,
because the z-index values in our SCSS was significantly higher than
Bootstrap's.

This moves most of z-index values to be based off of variables, defined
directly below the Bootstrap z-index definitions, and lowers all z-index
values where appropriate.

I also removed several z-index properties altogether because I could not
determine what effect, if any, they had.


maurizi:
@jfrankl If you think I missed anything or need to put some of the deleted `z-index` statements back, I'll defer to your judgement.

PR #464, opened 2015-02-06
Add ability to cancel reservations

maurizi:
Clicking "Cancel This" from the action bar will open a Bootstrap modal

After clicking "Cancel Reservation" on the modal, the reservation is canceled
on the backend, and tile/grid layers are updated on the client.

There are known issues with the legend in desktop mode interfering with the cancel modal that will be fixed by #463

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/partials/reserved_blockface_popup.html
+++ b/src/nyc_trees/apps/survey/templates/survey/partials/reserved_blockface_popup.html
@@ -13,13 +13,39 @@
             </button>
             <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="reserved-blockface-popup-submenu">
                 <li>
-                    <a role="menuitem" tabindex="-1"
-                      href="{% url 'survey' blockface_id=blockface.id %}">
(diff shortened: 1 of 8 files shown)


RickMohr:
+1 Very clear.

kdeloach:
+1

src/nyc_trees/apps/survey/templates/survey/partials/reserved_blockface_popup.html
```diff
@@ -13,13 +13,39 @@
...
+         The solution I arrived at is to place the modal contents in a script tag,
+         and move them into the modal when this partial is loaded -->
+    <script type="html/template" id="cancel-reservation-template">
+    <div class="modal-dialog">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h4 class="modal-title" id="reservation-modal-label">Cancel Reservation</h4>
+            </div>
+            <div class="modal-body">
+                This reservation will expire on {{ reservation.expires_at|date }}.
```
jwalgran: From a mappers perspective, being shown the expiration date in the confirmation dialog is a little confusing. What is the situation in which knowing the expiration date would change whether or not I follow through with my cancellation?

jwalgran:
+1 I was able to add a new reservation, then cancel it

PR #466, opened 2015-02-06
Bug/400 prevent dates before now

steventlamb:


src/nyc_trees/apps/event/forms.py
```diff
@@ -48,17 +48,33 @@ class Meta:
...
+
+            # first check if the date is in the past and if so, raise on date
+            if cleaned_date < today:
+                Event.raise_date_past('date')
+            if 'begins_at_time' in cleaned_data:
+                begins_at_time = cleaned_data['begins_at_time']
+
+                # then, check if the date is the same as today, in which case
+                # time matters, and if so, raise on the time
+                if cleaned_date == today and begins_at_time < now().time():
```
maurizi: Shouldn't this be `right_now.time()`, not `now().time()`?

steventlamb: yes, good catch

maurizi:
+1

PR #467, opened 2015-02-06
Add support for checkboxes on training quiz

kdeloach:
The first question of the training quiz is now a multiple choice question to demonstrate this feature.

### DIFF
--- a/src/nyc_trees/apps/home/training/types.py
+++ b/src/nyc_trees/apps/home/training/types.py
@@ -180,18 +180,18 @@ class Quiz(object):
     @classmethod
     def extract_answers(cls, post_fields):
         """
-        Return dict of {question_index => answer_index, ...} from post fields.
+        Return dict of {question_index => answer_indexes, ...} from post fields
         post_fields - QueryDict; Probably from request.POST
         """
         result = []
-        for field_name, values in post_fields.iteritems():
+        for field_name, values in post_fields.iterlists():
             # Parse field name format like "question.1", "question.2", etc.
             parts = field_name.split('.')
             if len(parts) == 2 and parts[0] == 'question':
                 try:
-                    value = int(values[0])
+                    answers = map(int, values)
                     question_index = int(parts[1])
-                    result.append((question_index, value))
+                    result.append((question_index, answers))
                 except ValueError:
                     pass
         return dict(result)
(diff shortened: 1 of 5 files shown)


src/nyc_trees/apps/home/training/types.py
```diff
@@ -199,25 +199,66 @@ def extract_answers(cls, post_fields):
...
+    def __init__(self, text, choices, answer):
+        super(Question, self).__init__(text, choices)
         assert answer >= 0
         assert answer < len(choices)
-        self.text = unicode(text)
         self.answer = int(answer)
-        self.choices = list(choices)
+
+    def is_correct(self, answer):
+        return answer and len(answer) > 0 and answer[0] == self.answer
```
steventlamb: The docstring says answer is an int, but here you access a member of it. Am I reading this wrong, or is this a documentation error?

src/nyc_trees/apps/home/training/types.py
```diff
@@ -199,25 +199,66 @@ def extract_answers(cls, post_fields):
...
         for i, question in enumerate(self.questions):
-            if answers.get(i, -1) == question.answer:
+            answer = answers.get(i, None)
+            if question.is_correct(answer):
                 result += 1
         return result
 
 
-class Question(object):
+class AbstractQuestion(object):
```
steventlamb: would it make more sense to represent a single choice question as a special case of a multiple choice question in which the set of correct answers has one member? Then you can do similar set operations for both types and probably just use one class.

src/nyc_trees/js/src/quiz.js
```diff
@@ -94,10 +98,11 @@ function restore() {
     var state = progress.load();
     if (state && state.form) {
         // Restore radio button selections.
-        $.each(state.form, function(fieldName, value) {
-            // Assumes that only checked values are serialized.
-            var radio = $('input[name="' + fieldName + '"][value="' + value + '"]');
-            radio.prop('checked', true);
+        $.each(state.form, function(fieldName, values) {
+            $.each(values, function(__, value) {
+                var radio = $('input[name="' + fieldName + '"][value="' + value + '"]');
```
steventlamb: isn't this now sometimes not a `radio`? Consider using a more generic term like `input`.

src/nyc_trees/apps/home/training/types.py
```diff
@@ -199,25 +199,66 @@ def extract_answers(cls, post_fields):
...
+    """
+    def __init__(self, text, choices, answer):
+        super(MultiChoiceQuestion, self).__init__(text, choices)
+        for ans in answer:
+            assert ans >= 0
+            assert ans < len(choices)
+        self.answer = set(answer)
+
+    def is_correct(self, answer):
+        return answer and set(answer) == self.answer
```
steventlamb: comparing a set, in which order is not significant, to an iterable, in which order does matter, is a little spooky. Consider using sets, or `frozenset`s for everything.

steventlamb:
+1 for code review. a clean way to represent this behavior. Did not yet pull down and exercise.

src/nyc_trees/apps/home/training/types.py
kdeloach: I know this is confusing, but in this context `answer` is `iterable<int>` (as documented by `is_correct`) and `self.answer` is `int` (as documented by the constructor). I think the mistake here was using the same variable name for 2 different contexts and datatypes. Would this make more sense if this argument was named `answers` or `answer_list`?

src/nyc_trees/apps/home/training/types.py
kdeloach: This is really just comparing 2 sets. I think I'll update this to `return answer and set(answer) == set(self.answer)` if that makes it clearer.

src/nyc_trees/js/src/quiz.js
kdeloach: Totally, good catch.

src/nyc_trees/apps/home/training/types.py
kdeloach: I thought about this a lot, and the reason why I didn't do it like that now escapes me. I'll drop the abstract class and make `MultiChoiceQuestion` the base class.

src/nyc_trees/apps/home/training/types.py
steventlamb: That would be fine, but why not store `self.answer` as a set directly?

kdeloach: It is (https://github.com/azavea/nyc-trees/pull/467/files#diff-3716652adc66700d57b01586f5487dafR258).

src/nyc_trees/apps/home/training/types.py
steventlamb: Your call. Just because you can't remember doesn't mean the reason wasn't good.

src/nyc_trees/apps/home/training/types.py
steventlamb: Ah, that makes sense. Most of this comment is not relevant. However, my confusion stems from my remaining concern, why represent them as tuples in `__init__.py` just to convert them?

src/nyc_trees/apps/home/training/types.py
steventlamb: how about `def is_correct(self, candidate)` ?

src/nyc_trees/apps/home/training/types.py
kdeloach: Good observation. The idea is that `set` is an implementation detail. As long as you define the quiz answer as some iterable (tuple, array, generator, etc.) it will work.

kdeloach:
@steventlamb All changes made.

steventlamb:
+1, pulled down and tested

PR #468, opened 2015-02-09
Add support for HVM NAT instances

hectcastro:
This changeset adds HVM AMI support to the NAT instances, which allows us to use `t2.*` instance types vs. the previous `t1.micro`.


steventlamb:
+1, tested via examination of steps in the staging aws console

PR #469, opened 2015-02-09
List/Approve/Reject individual mappers

kdeloach:
This lists mappers on the group admin page and adds an Approve/Deny button next to each mapper request.

How to test:
1. Since there is currently no interface to request group mapper status, manually create an entry in the superadmin [http://localhost:8000/admin/users/trustedmapper/](http://localhost:8000/admin/users/trustedmapper/).
2. Requests with unknown (NULL) status should appear in the Mappers > Pending tab.
3. Requests with approved/denied status should appear in the Mappers > All tab.

### DIFF
--- a/src/nyc_trees/apps/survey/tests.py
+++ b/src/nyc_trees/apps/survey/tests.py
@@ -79,6 +79,7 @@ class ConfirmBlockfaceReservationTests(TestCase):
 
         self.assert_blocks_reserved(0, self.block1)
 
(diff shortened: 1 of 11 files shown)


jwalgran:
Looking at this now

jwalgran:
+1 works perfectly. I denied a request, and viewed the mappers reservation page. All blockfaces unavailable. I then went back and approved the request, then refreshed the users reservation page. Available blocks lit up. Nice.

PR #470, opened 2015-02-09
Redirect to event tab on group edit after adding

### LINKED ISSUE #401: After adding a new event the page should go to the events tab on the group's page

steventlamb:
fixes #401 on github

### DIFF
--- a/src/nyc_trees/apps/event/views.py
+++ b/src/nyc_trees/apps/event/views.py
@@ -14,7 +14,7 @@ from django.utils.timezone import get_current_timezone, now
 from apps.core.forms import EmailForm
 from apps.core.models import User
 
-from apps.users.views.group import GROUP_EVENTS_ID
+from apps.users.views.group import GROUP_EVENTS_ID, GROUP_EDIT_EVENTS_TAB_ID
 
 from apps.event.forms import EventForm
 from apps.event.models import Event, EventRegistration
@@ -29,8 +29,10 @@ def add_event(request):
     is_valid = _process_event_form(form, request)
 
     if is_valid:
+        base_url = reverse(
+            'group_edit', kwargs={'group_slug': request.group.slug})
         return HttpResponseRedirect(
-            reverse('group_edit', kwargs={'group_slug': request.group.slug}))
+            '%s#%s' % (base_url, GROUP_EDIT_EVENTS_TAB_ID))
     else:
         return {
             'form': form,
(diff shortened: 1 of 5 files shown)


src/nyc_trees/js/src/lib/tabhash.js
```diff
@@ -0,0 +1,24 @@
...
+
+    // bootstrap swallows events so the url doesn't change
+    // add our own here
+    $(tabList).on('click', 'a', function (e) {
+        $(this).tab('show');
+
+        var originalScrollPosition = $('body').scrollTop();
+        window.location.hash = this.hash;
+        $('html,body').scrollTop(originalScrollPosition);
+    });
```
maurizi: Rather than intercepting clicks and then triggering a tab to open via `$(this).tab('show')`, consider relying on Bootstrap's events.

Also, I think we can drop the scroll position stuff by using the History API, which [is supported on most recent browsers](http://caniuse.com/#search=replaceState).

This is one option for what this could look like if you follow both of those suggestions:
```javascript
$(tabList).on('show.bs.tab', function(e) {
    var newTab = e.target,
        url = window.location.pathname + '#' + newTab.hash;
    window.history.replaceState(null, null, url);
});

File: src/nyc_trees/apps/event/views.py [hunk 2]
```diff
@@ -29,8 +29,10 @@ def add_event(request):
     is_valid = _process_event_form(form, request)
 
     if is_valid:
+        base_url = reverse(
+            'group_edit', kwargs={'group_slug': request.group.slug})
         return HttpResponseRedirect(
-            reverse('group_edit', kwargs={'group_slug': request.group.slug}))
+            '%s#%s' % (base_url, GROUP_EDIT_EVENTS_TAB_ID))
```
kdeloach: I would be in favor of hard-coding `#events` throughout instead of storing it in a constant.

kdeloach:
I can't switch to the events tab on the group admin page now (I ran build-debug and restarted my debugserver).

src/nyc_trees/js/src/lib/tabhash.js
kdeloach: I agree with using the `show.bs.tab` event instead, but I would suggest that we avoid altering the scroll position. I find it a little disorienting when clicking the Details tab and having the tab bar disappear.

kdeloach: Consider displaying the correct tab when the `onhashchange` event fires (back button press).

steventlamb:
@kdeloach ah, that was a last minute breaking change where I failed to rename a variable in all places. Fixed now.

kdeloach:
+1

PR #474, opened 2015-02-10
Add support for EBS volume SSD storage

hectcastro:
This changeset switches our default EBS volume types from `standard` (AKA magnetic) to `gp2` (AKA general purpose SSD) by using an SSD based AMI.

See also: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html


hectcastro:
@azavea-bot rebuild

PR #475, opened 2015-02-10
Add conservative globs to the geospatial libraries dependencies

hectcastro:
This changeset adds globs to the versions of the geospatial libraries used by the Django application.


kdeloach:
+1

PR #476, opened 2015-02-10
Add Tasseo as a dashboard for select Graphite metrics

hectcastro:
This changeset adds support for Tasseo, which is a dashboard for highlighting select Graphite metrics. As of right now, the only metrics highlighted are:

- Django response codes
- E-mail success/failures
- Windshaft success/failures

tasseo

### DIFF
--- a/deployment/ansible/group_vars/packer
+++ b/deployment/ansible/group_vars/packer
@@ -17,3 +17,5 @@ tiler_host: "tile.service.nyc-trees.internal"
 postgresql_username: nyctrees
 postgresql_password: nyctrees
 postgresql_database: nyc_trees
+
+tasseo_graphite_url: "http://$(curl -s http://instance-data.ec2.internal/latest/meta-data/public-hostname):{{ graphite_web_port }}"
--- a/deployment/ansible/roles/nyc-trees.tasseo/templates/nyc-trees.js.j2
+++ b/deployment/ansible/roles/nyc-trees.tasseo/templates/nyc-trees.js.j2
@@ -1,0 +1,50 @@
+var period = 10;
+var metrics = [
+    {
+        "alias": "django.percent-error",
+        "target": "scale(divideSeries(transformNull(sumSeries(statsite.counts.django.response.{4,5}*),0),transformNull(sumSeries(statsite.counts.django.response.{1,2,3}*),0)),100)",
+        "unit": "%",
+        "warning": 10,
+        "critical": 15
+    },
+    {
+        "alias": "django.1-3XX",
+        "target": "transformNull(sumSeries(statsite.counts.django.response.{1,2,3}*),0)"
+    },
+    {
+        "alias": "django.4-5XX",
+        "target": "transformNull(sumSeries(statsite.counts.django.response.{4,5}*),0)"
+    },
+    {% if ['packer'] | is_in(group_names) -%}
+    {
+        "alias": "email.percent-error",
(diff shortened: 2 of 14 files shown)


hectcastro:
@azavea-bot rebuild

deployment/ansible/group_vars/packer
```diff
@@ -17,3 +17,5 @@ tiler_host: "tile.service.nyc-trees.internal"
 postgresql_username: nyctrees
 postgresql_password: nyctrees
 postgresql_database: nyc_trees
+
+tasseo_graphite_url: "http://$(curl -s http://instance-data.ec2.internal/latest/meta-data/public-hostname):8080"
```
jwalgran: It's cool to use a hard-coded 8080 because we now use unique IP/hostnames to run multiple instances, rather than changing ports. Is my understanding correct?

deployment/ansible/roles/nyc-trees.tasseo/templates/nyc-trees.js.j2
```diff
@@ -0,0 +1,50 @@
+var period = 10;
+var metrics = [
+    {
+        "alias": "django.percent-error",
+        "target": "scale(divideSeries(transformNull(sumSeries(statsite.counts.django.response.{4,5}*),0),transformNull(sumSeries(statsite.counts.django.response.{2,3}*),0)),100)",
```
jwalgran: Are 1xx responses excluded from the percent-error metric by design?

jwalgran:
I provisioned and can load tasseo, but AJAX calls to graphite are failing. Maybe a lack of data in my graphite DB, perhaps?

screen shot 2015-02-10 at 11 17 10 pm

deployment/ansible/roles/nyc-trees.tasseo/templates/nyc-trees.js.j2
hectcastro: Fixed in https://github.com/azavea/nyc-trees/commit/d1554da050a45569db959be4fca8a353ca102e1f.

deployment/ansible/group_vars/packer
hectcastro: Hm. This should be `graphite_web_port`, which ends up being whatever `apache_port` is (Graphite's web interface is a Django application that runs with `modwsgi`).

In development, we're bypassing the Vagrant port forwarding because we're using the machine's private host-only interface IP. In testing, monitoring stuff like this is disabled. In production, we just use the default ports for everything because the default ports are hardcoded in the CloudFormation.

Fixed in https://github.com/azavea/nyc-trees/commit/bd14bcd316f7fb62117c3a264377b65c83bebfdf.

hectcastro:
Unfortunately, `divideSeries` doesn't handle missing metrics well. I haven't found a good solution for it yet. I hit the progress page locally to populate metrics, which allows the page to render.

Alternatively, you can send dummy metrics to Statsite using `nc` from the `services` virtual machine:

```bash
$ echo "windshaft.tiles.failure:1|c" | nc -w 0 -u localhost 8125
```

`-w` is a timeout and `-u` tells it to use UDP.

jwalgran:
I won't be able to look at this update for a while. If you want to get it merged in soon, please pass it along.

kdeloach:
Here's what I get when I run `vagrant provision`:

image

hectcastro:
The fix for that was merged into `develop` in https://github.com/azavea/nyc-trees/commit/3ab2048c6b75ec5645088b1802dfe232e6d568e3. You should be able to get around it by rebasing the latest `develop`.

kdeloach:
Worked (more or less) after rebasing.

kdeloach:
+1

PR #487, opened 2015-02-11
Request mapper status

kdeloach:
Click on unavailable blockfaces to request group mapper status.

How to test:

1. Restart tiler and rebuild JS
2. Enable `individual_mapper` on your test user
3. Go to the "new reservations" page [http://localhost:8000/blockface/reserve/](http://localhost:8000/blockface/reserve/)
3. If a group has `allows_individual_mappers` disabled, when you click on blocks, you should get a generic error message.
4. If a group has `allows_individual_mappers` enabled, when you click on blocks, you should get a popup that asks if you would like to request permission for group mapper status.
5. The mapper status request should appear on the group admin page [http://localhost:8000/group/east-side/edit/](http://localhost:8000/group/east-side/edit/) (replace group name)

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
+++ b/src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
@@ -38,6 +38,10 @@
 
 {% endblock main %}
 
+{% block extra_content %}
+{% include "survey/partials/reserve_blockface_modals.html" %}
+{% endblock extra_content %}
+
 {% block page_js %}
     <script type="text/javascript" src="{{ "js/reserveBlockfacePage.js"|static_url }}"></script>
 {% endblock page_js %}
--- a/src/nyc_trees/apps/users/views/group.py
+++ b/src/nyc_trees/apps/users/views/group.py
@@ -206,3 +206,12 @@ def _grant_mapping_access(group, username, is_approved):
     return {
         'mapper': mapper
     }
+
+
+def request_mapper_status(request):
+    mapper, created = TrustedMapper.objects.update_or_create(
+        group=request.group,
+        user=request.user)
+    return {
+        'success': True
+    }
--- a/src/tiler/sql/user_reservable.sql
+++ b/src/tiler/sql/user_reservable.sql
@@ -7,21 +7,27 @@
     <% } %>
     *
     FROM (SELECT
-      block.geom, block.id, turf.group_id,
+      block.geom,
+      block.id,
+      turf.group_id,
+      grp.slug AS group_slug
       CASE
         WHEN (block.is_available AND reservation.id IS NULL
               AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'
         ELSE 'unavailable'
       END AS survey_type,
       CASE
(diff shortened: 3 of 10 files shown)


src/tiler/sql/user_reservable.sql
```diff
@@ -7,21 +7,28 @@
     <% } %>
     *
     FROM (SELECT
-      block.geom, block.id, turf.group_id,
+      block.geom,
+      block.id,
+      turf.group_id,
+      grp.slug AS group_slug,
+      grp.allows_individual_mappers,
```
maurizi: This isn't used in the UtfGrid anymore, so you should probably drop it.

maurizi:
Aside from my 1 comment, looks good to me (I haven't tested it)

src/tiler/sql/user_reservable.sql
kdeloach: Thanks, missed this during a rebase.

src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
```diff
@@ -38,6 +38,107 @@
 
 {% endblock main %}
 
+{% block extra_content %}
+<div class="modal fade" id="limit-reached-modal">
```
steventlamb: I think there are enough modals here to justify a modal partial template.

src/nyc_trees/apps/users/views/group.py
```diff
@@ -206,3 +206,12 @@ def _grant_mapping_access(group, username, is_approved):
     return {
         'mapper': mapper
     }
+
+
+def request_mapper_status(request):
+    mapper, created = TrustedMapper.objects.update_or_create(
```
steventlamb: Would it be better to have this view raise and return a `4xx` code if executed repeatedly?

steventlamb: this can be `mapper, _` but I guess it doesn't matter. I thought pep8 would complain about that.

kdeloach: We don't prevent users from repeatedly requesting mapper status in the UI so this is a compromise.

src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
steventlamb: I probably would have solved this problem with one modal and text mutations, but this is a little more elegant.

steventlamb:
+1 for code review. Some of it touches things I'm not familiar with, but it all looks pretty solid. Provisioning now, will follow up with test results.

src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
kdeloach: I thought this would be easier for the design team to customize rather than rebuilding the JS bundle.

steventlamb:
+1 works

src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
kdeloach: Thanks, I moved these to a partial.

PR #501, opened 2015-02-12
Bug/436 fix privacy button

steventlamb:


src/nyc_trees/apps/core/templates/core/partials/nav.html
```diff
@@ -4,7 +4,9 @@
         <span>
             <h2><a href="{% url 'user_detail_redirect' %}" class="nav-side-user-name">{{ user.username }}</a></h2>
             <div class="nav-side-options">
-                <a href="{% url 'user_profile_settings' %}" class="btn btn-primary">Settings</a> <a href="#" class="btn btn-primary">Privacy</a>
+                <a href="{% url 'user_profile_settings' %}"
+                   class="btn btn-primary">Settings</a>&nbsp;<a href="{{ user_settings_privacy_url }}"
```
kdeloach: Consider simplifying by just using `{% url 'user_profile_settings' %}#privacy-pane` here.

kdeloach:
+1

maurizi:
-1.  According to the linked issue, it should fairly easy to fix this the right way, by restoring the mistakenly deleted modal.

jwalgran:
I do want to restore the privacy pop-up, since it was mistakenly deleted, but I also very much want to keep the functionality added in this pull. On any page with tabs is it much friendlier when refreshing the page stays on the same tab.

maurizi:
Perhaps it would make more sense to move the tab handling code into [`base.js`](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/base.js) so we don't have to explicitly enable on every page?

Do we have anything else that utilizes the URL hash?

jwalgran:
Yes. The reservation page uses the hash for something other than tabs.

maurizi:
Well it's a bit bigger than this PR, but I would be in favor of changing the reservation page to use a query string parameter, and enabling the tab/hash saving in `base.js`

kdeloach:
Putting the tab hash stuff in base may be more trouble than it's worth.

steventlamb:
If we put tabhashing in base.js, we'd have to convert the selector for a tabhash list to something other than id, and we'd have to make sure it works with multiple tab lists per page, whether we already have them, or to avoid a regression in the future.

I agree with @kdeloach, we shouldn't do this unless it becomes necessary.

src/nyc_trees/apps/core/templates/core/partials/nav.html
steventlamb: I did consider it, and I strongly prefer this solution. Stringly typed URLs are a one way street, all it takes is one non-descriptive django_url or hash name, and searchability during refactors becomes very, difficult.

PR #510, opened 2015-02-12
Make location search UI take up less space

### LINKED ISSUE #507: Hide the status message when geocoder shows multiple results
Space is at a premium on a small screen and the "Choose an address or search again" message is not adding a lot of value.

screen shot 2015-02-12 at 8 39 16 am

### LINKED ISSUE #509: Hide the search bar after zooming to a location
Space is precious on a small screen so we want to show as much of the map as possible after we zoom

screen shot 2015-02-12 at 8 39 16 am

jwalgran:
I talked over our options with @jfrankl and made two changes to make location search more useful on small mobile phones.

1. The status message is hidden when there are multiple results. The
   message was not helpful and it now allows an extra result to be
   shown.

2. The search UI is hidden automatically after zooming to a result. This
   allows more of the map to be visible. It is not likely that a
   person will immediately search again and, if they do, making them tap
   the search toolbar button again is worth the extra space gained in
   trade.

Fixes #507 
Fixes #509

### DIFF
--- a/src/nyc_trees/js/src/searchController.js
+++ b/src/nyc_trees/js/src/searchController.js
@@ -106,7 +106,7 @@ function create($controlsContainer, map) {
         },
 
         showCandidates = function(candidates) {
-            updateStatus('Choose an address or search again', {resultsFound: true});
+            updateStatus('', {resultsFound: true});
             removeResults();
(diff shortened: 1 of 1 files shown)


kdeloach:
+1

PR #514, opened 2015-02-12
Add script to prune stale staging AMIs

hectcastro:
This changeset adds a script to aid in the removal of stale AMIs in staging. Right now, it keeps the 10 most recent AMIs by AMI type (app, tiler, monitoring, etc.).

The pruning script depends on the `Environment` tag to determine AMIs eligible for removal. AMIs targeted at other environments will never be pruned.

Example output of the script running under Jenkins: http://civicci01.internal.azavea.com/job/nyc-trees-packer/126/console


maurizi:
+1

PR #519, opened 2015-02-12
Fix typo in reservable SQL.  Fixes #518

### LINKED ISSUE #518: Tile layer on new reservation page is failing due to a SQL error
```
{"error":"ERROR:  syntax error at or near \"CASE\"\nLINE 9:       CASE\n              ^\n\nFull sql was: 'SELECT * FROM (SELECT\n    \n    *\n    FROM (SELECT\n      block.geom,\n      block.id,\n      turf.group_id,\n      grp.slug AS group_slug\n      CASE\n        WHEN (block.is_available AND reservation.id IS NULL\n              AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'\n        ELSE 'unavailable'\n      END AS survey_type,\n      CASE\n        WHEN reservation.id IS NOT NULL THEN 'reserved'\n        WHEN grp.id IS NOT NULL AND NOT grp.allows_individual_mappers THEN 'unavailable'\n        WHEN NOT block.is_available THEN 'unavailable'\n        WHEN turf.id IS NOT NULL AND trustedmapper.id IS NULL THEN 'group_territory'\n        ELSE 'none'\n      END AS restriction\n      FROM survey_blockface AS block\n      LEFT OUTER JOIN survey_territory AS turf\n        ON block.id = turf.blockface_id\n      LEFT JOIN core_group AS grp\n        ON grp.id = turf.group_id\n      LEFT OUTER JOIN survey_blockfacereservation AS reservation\n        ON (block.id = reservation.blockface_id\n            AND reservation.canceled_at IS NULL\n            AND reservation.expires_at > now() at time zone 'utc')\n      LEFT OUTER JOIN users_trustedmapper AS trustedmapper\n        ON (turf.group_id = trustedmapper.group_id\n            AND trustedmapper.user_id = 5\n            AND trustedmapper.is_approved)\n    ) AS subquery\n) AS query LIMIT 0'\n  encountered during parsing of layer 'survey_blockface' in Layer at line 177"}

maurizi:


kdeloach:
Thanks, I caused this during a recent rebase. Was too slow on the draw to fix this one.

kdeloach:
+1

Issue #517, opened 2015-02-12
Put new individual mapper workflow in place


dboyer:

RickMohr:
1. User completes online training and in-field training event

2. User attends another training or mapping event 

2. After attending those two events (one training and one training or mapping), a "request individual mapper status" button appears on their user profile page with some text describing what this is. We do not need to put up any notification on the dashboard about this button.

3. Users clicks the button and automatically is approved as an individual mapper

4. After clicking the button, the user is immediately taken to a page (maybe this is a pop-up?) that says "you've been granted individual mapper status! learn more!" with a page of instructions on reservations. They then can click through to the reservations page. 

5. Parks staff can rescind individual mapper status by unchecking the individual mapper box on that user's page in the admin interface.

kdeloach:
Should the reservations page be accessible for non-individual mappers? What should happen when they click Reservations in the sidebar, and what should happen when they try to create a New Reservation?

jwalgran:
I would suggest that the "Reservations" sidebar item is hidden if you are not an individual mapper (much like the reservations section on the dashboard/profile). I would want a second on that from Deb/Parks before committing to it.

jfrankl:
Number 3 on Rick's list above should also appear on the dashboard, since that is where a user will find out about ways they can engage with the tree census (follow a group, join an event, request individual mapper status).

kdeloach:
Since the dashboard shares a partial template with the user profile page, this button appears in both places.

![selection_077](https://cloud.githubusercontent.com/assets/43062/6220295/d17aee1c-b602-11e4-9bb6-11f83be37d30.png)

kdeloach:
@jfrankl Can we close this?

PR #530, opened 2015-02-13
Add individual mapper status workflow

kdeloach:
Once a user has completed online training, attended 1 training event, and 1 other event, they should see a button on their user profile to request individual mapper status. When clicked, this button will immediately grant access to the reservations page and the ability to reserve blockfaces.

Migrations must be run for testing.

See #517 for a description of the workflow that should be followed for verification purposes.

### DIFF
--- a/src/nyc_trees/apps/core/decorators.py
+++ b/src/nyc_trees/apps/core/decorators.py
@@ -12,6 +12,7 @@ from django.shortcuts import get_object_or_404
 from django_tinsel.utils import decorate as do
 
 from apps.core.helpers import (user_is_census_admin, user_is_group_admin,
+                               user_is_individual_mapper,
                                user_has_online_training,
                                user_has_field_training)
 from apps.core.models import Group
@@ -84,10 +85,13 @@ def user_must_have_field_training(view_fn):
 
 
 def user_must_be_individual_mapper(view_fn):
-    # TODO: implement
     @wraps(view_fn)
     def wrapper(request, *args, **kwargs):
-        return view_fn(request, *args, **kwargs)
+        if user_is_individual_mapper(request.user):
+            return view_fn(request, *args, **kwargs)
+        else:
+            raise PermissionDenied('%s is not an individual mapper'
+                                   % request.user)
     return wrapper
 
 ##############################################################################
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -105,6 +105,17 @@ class User(NycModel, AbstractUser):
     def training_complete(self):
         return self.online_training_complete and self.field_training_complete
 
+    @property
+    def eligible_to_become_individual_mapper(self):
+        """
+        Return True if user has completed online training, attended a mapping
+        event, and one other training/mapping event.
+        """
+        from apps.event.models import EventRegistration
+        return self.field_training_complete and \
+            EventRegistration.objects.filter(user=self,
+                                             did_attend=True).count() >= 2
+
 
 class Group(NycModel, models.Model):
     name = models.CharField(max_length=255, unique=True)
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -28,7 +28,7 @@ REFERRER_AD = (
 
 
 class User(NycModel, AbstractUser):
-    individual_mapper = models.BooleanField(default=False)
+    individual_mapper = models.NullBooleanField()
     requested_individual_mapping_at = models.DateTimeField(null=True,
                                                            blank=True)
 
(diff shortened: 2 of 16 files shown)


src/nyc_trees/apps/core/decorators.py
```diff
@@ -87,7 +88,11 @@ def user_must_be_individual_mapper(view_fn):
     # TODO: implement
```
maurizi: Shouldn't this TODO be removed now?

src/nyc_trees/apps/core/models.py
```diff
@@ -105,6 +105,20 @@ def online_training_complete(self):
...
+        Return True if user has completed field training and at least
+        one other event.
+        """
+        # Has an admin rescinded mapper status?
+        if self.individual_mapper is False:
+            return False
+        from apps.event.models import EventRegistration
+        return self.field_training_complete and \
+            EventRegistration.objects.filter(user=self,
+                                             did_attend=True).count() > 1
```
maurizi: Consider rewriting this as:
```python
return self.field_training_complete and \
    self.eventregistration_set.filter(did_attend=True).exists()
```

maurizi:
Looks good to me

kdeloach:
I just noticed that I forgot to populate the "Requested individual mapping at" field. Is this still needed or should I delete it?

jwalgran:
Reviewing now

jwalgran:
I think we should populate ``requested_individual_mapping_at`` since it is so simple to do and is a nice time series metric.

src/nyc_trees/apps/users/urls/group.py
```diff
@@ -34,11 +34,10 @@
         name='edit_user_mapping_priveleges'),
 
     # Keep in sync with src/nyc_trees/js/src/reserveBlockfacePage.js
-    url(r'^(?P<group_slug>[\w-]+)/request-individual-mapper-status/$',
+    url(r'^(?P<group_slug>[\w-]+)/request-trusted-mapper-status/$',
```
jwalgran: Does ``trusted mapper`` == ``individual mapper``? If so, this more correct name may actually cause confusion.

jwalgran:
+1 The app behavior matches the workflow

- Newly registered user gets 403 on ``/blockface``
- "Reservations" link is missing from the sidebar fro a new user
- "Reservations" remain hidden after completing online training
- Getting checked in at second event makes "Request" button appear on the dashboard and the user profile.
- Clicking the "Request" button  made "Reservations" appear in the burger menu and ``/blockface`` accessible.
- Rescinding individual mapper status through the Django admin caused "Reservations" to disappear and ``/blockface`` to become a 403.

src/nyc_trees/apps/users/urls/group.py
kdeloach: The ORM model for group "individual mappers" is called TrustedMapper so this new name should be more accurate. This particular change isn't related to the global "individual mappers" concept but I lumped it together with the other changes anyway.

src/nyc_trees/apps/core/decorators.py
kdeloach: Will do.

src/nyc_trees/apps/core/models.py
```diff
@@ -105,6 +105,20 @@ def online_training_complete(self):
     def training_complete(self):
         return self.online_training_complete and self.field_training_complete
 
+    @property
+    def qualified_as_individual_mapper(self):
```
kdeloach: I'm considering renaming this to `elligible_for_individual_mapping`.

src/nyc_trees/apps/core/models.py
kdeloach: You actually need to attend 2 events, I could clarify this by saying `count() >= 2`.

kdeloach: Although I dislike having to import `EventRegistration` here to avoid a circular-dependency issue, I still prefer doing this rather than using `xxx_set`.

src/nyc_trees/apps/core/models.py
RickMohr: That would be `eligible_...`

RickMohr:
I asked @dboyer "If individual mapper status is rescinded, should the 'request individual mapper status' button be shown on the profile or not?" She said "No, I don't think so, but I'll confirm with Parks."

If "No" is indeed the desired behavior then `requested_individual_mapping_at` will allow us to do the right thing. So, consider:

* Setting `requested_individual_mapping_at` when the request is made
* Checking it when deciding whether to display the button.

src/nyc_trees/apps/core/models.py
kdeloach: Thanks.

kdeloach:
Currently, it only shows the "Request mapper status" button if `individual_mapper` is null. If mapper status is rescinded this value is false, so the button does not appear.

RickMohr:
Subtle but simple.

dboyer:
This method should be fine since they could always shown the button again by setting the value to unknown.

PR #536, opened 2015-02-13
Don't use filename as a key in javascript

### LINKED ISSUE #484: No date or time picker when adding a new event on staging

steventlamb:
fixes #484 on github

The build tooling replaces occurrences of this filename with a modified
version that has a snippet of a sha attached to it. We use this filename
as a key in configuration and it caused a failed key lookup.

### DIFF
--- a/src/nyc_trees/apps/core/templates/core/partials/config.js
+++ b/src/nyc_trees/apps/core/templates/core/partials/config.js
@@ -2,7 +2,7 @@
 (function(window, Object) {
     window.config = {
         "files": {
-            "datetimepicker_polyfill.js": "{{ 'js/datetimepicker_polyfill.js'|static_url }}"
+            "dtpPolyfill": "{{ 'js/datetimepicker_polyfill.js'|static_url }}"
         },
         "urls": {
             "geocode": "{% url 'geocode' %}",
--- a/src/nyc_trees/js/src/event_form.js
+++ b/src/nyc_trees/js/src/event_form.js
@@ -17,7 +17,7 @@ var dom = {
 
 Modernizr.load({
     test: Modernizr.inputtypes.date && Modernizr.inputtypes.time,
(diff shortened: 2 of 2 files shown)


steventlamb:
I was not able to truly confirm that this will work, but it doesn't introduce regressions in development so is at worst a no-op. It looks like this will fix the issue, based on reasoning through it and inspecting the production js output that I built locally.

kdeloach:
+1

steventlamb:
The failed build gave a useful comment about camel case that I will follow.

maurizi:
I don't see how this could help fix the bug...  

Then again I suppose it can't hurt either.
I can help you investigate on Monday if this doesn't fix it.

kdeloach:
On our staging environment, view the source of the edit event page ([http://treescount.azavea.com/group/manhattan-tree-mappers/event/test-even-123/edit/](http://treescount.azavea.com/group/manhattan-tree-mappers/event/test-even-123/edit/)). Observe that this Django template has some inline JS cofiguration for the datetime polyfill:

    window.config = {
        "files": {
            "datetimepicker_polyfill.js": "/static/js/datetimepicker_polyfill.4d0a2496.js"
        },
        ...
    }
    
Then view the source of the bundle that is loaded on the edit event page ([http://treescount.azavea.com/static/js/event_form.884c3d42.js](http://treescount.azavea.com/static/js/event_form.884c3d42.js)). We would expect to find that the datetime polyfill key referenced in this bundle would match the one defined on the edit event page, but the key is (presently) `nope:config.files["datetimepicker_polyfill.4d0a2496.js"]`.

Part of our JS build process replaces references to files in our source code with a tagged version. Since this does not affect the Django templates, there is a mismatch on this key definition.

maurizi:
Oh of course, since the key is the same as the name of an entry bundle, it gets replaced with the tagged name...

Very tricky.  Good job tracking that down.

PR #537, opened 2015-02-13
Drastically reduce dom repainting; Fixes #528 (I think)

### LINKED ISSUE #528: Scrolling on iOS does not have smooth deceleration
The list views (groups, events, etc) on both a physical iPhone 4S and the iOS Simulator do not smoothly decelerate when scrolling.

This SO answer may be applicable http://stackoverflow.com/a/18752290

jfrankl:
Justin,

This should increase the scrolling speed on iOS (and all devices). I removed all fixed elements, which were causing the browser to repaint the entire window, since they were in moving elements.

More information here:

https://remysharp.com/2012/05/24/issues-with-position-fixed-scrolling-on-ios

I can try adding the css you suggested if this doesn't do it.

### DIFF
--- a/src/nyc_trees/sass/partials/_globals.scss
+++ b/src/nyc_trees/sass/partials/_globals.scss
@@ -37,7 +37,7 @@ body {
   width: 100%;
   transition: all 500ms $transition;
   position: relative;
-  height: 100%;
+  min-height: 100%;
   font-feature-settings: "kern" 1;
   text-rendering: geometricPrecision;
 }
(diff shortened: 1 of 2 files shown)


jwalgran:
+1 silky smooth

jfrankl:
Yay!

Also, a plug for next week's DSM: John, Christopher, and I are going to team up to talk about browser/device testing. I'm going to talk a little about how I caught this repainting issue.

Issue #491, opened 2015-02-11
Event edit: Time validation uses wrong timezone


RickMohr:
1) Edit an event
2) Change the start time to 1 hour from now, and hit "Save"

Error message: "Event must not begin in past"

You can only change the start time to something beyond 5 hours from now, making me suspect we are comparing against GMT

maurizi:
Probably related to https://github.com/azavea/nyc-trees/issues/147

PR #539, opened 2015-02-13
Validate event dates using Eastern not UTC

jwalgran:
The date fields are passed in to the ``clean`` method as they appear on the form, which means they will always be Eastern US time. We need to convert now() to Eastern US time before doing any comparisons.


kdeloach:
+1 I was able to add an event without having to add a 5 hour offset.

jwalgran:
Fixes #539

jwalgran:
Fixes #491

PR #541, opened 2015-02-15
Feature/training

jfrankl:
image

image

### DIFF
--- a/src/nyc_trees/apps/home/templates/home/quiz_page.html
+++ b/src/nyc_trees/apps/home/templates/home/quiz_page.html
@@ -16,7 +16,7 @@
             </div>
         </div>
     </div>
-{% endblock aside %}
+{% endblock aside %}x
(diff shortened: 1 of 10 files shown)


src/nyc_trees/apps/home/templates/home/quiz_page.html
```diff
@@ -16,7 +16,7 @@
             </div>
         </div>
     </div>
-{% endblock aside %}
+{% endblock aside %}x
```
maurizi: Typo?

steventlamb:
+1 tested this branch, it works and looks great. I had a little trouble telling the difference between multi-answer questions and single answer questions, we should consider adding a "Check all that apply" div that is visible for multi-answer questions too.

jfrankl:
Great idea Steve. Just added an issue for that.

PR #546, opened 2015-02-16
Implement 'This week's events' on dashboard

### LINKED ISSUE #534: "This week's events" section of the dashboard does not display events from followed groups in the near future
To reproduce:

1. As group admin: Add an event 4 days from now
1. As mapper following group the group: load the dashboard

The new event should appear under "This week's events" but it does not.

steventlamb:
fixes #534 on github


jwalgran:
Reviewing now

jwalgran:
+1 works perfectly

The event list on the Group detail page shows a "+ RSVP" label on the events to which the logged in user is signed up for. This would be nice to have on the dashboard, profile, and events pages. How much extra work it is to show this label everywhere?

Issue #442, opened 2015-02-04
Add a visual indication to the training page when a section is not yet available


jwalgran:
![screen shot 2015-02-03 at 5 44 37 pm](https://cloud.githubusercontent.com/assets/17363/6032937/862dfe80-abcc-11e4-87e0-b8c2deeafd20.png)

jfrankl:
@steventlamb, could you add a class on these elements that are not yet available?

PR #548, opened 2015-02-16
Add class for inactive training steps for styling

steventlamb:
addresses #442 on github


kdeloach:
+1 Verified that "inactive" class is applied to training steps that are not yet available.

PR #550, opened 2015-02-16
Fix 403 error when RSVP-ing for training event

### LINKED ISSUE #521: 403 Error when user RSVPs to Training Event when online_training_complete is False
The RSVP button should probably forward to either the training page or a page with instructions on how to complete field training.

kdeloach:
Users must complete online training before they can attend a training
event. Now, when the user tries to RSVP for a training event before
completing the online training, they will be forwarded to the training
instructions page.

Fixes #521 

How to test this:
1. Login as a user with no field/online training
2. RSVP to a Mapping event (should redirect to training instructions)
3. RSVP to Training event (should redirect to training instructions)
4. Complete online training steps
5. RSVP to a Mapping event (should redirect to training instructions)
6. RSVP to Training event (should RSVP you to that event)


maurizi:
+1, works

PR #552, opened 2015-02-16
Feature/549 add rsvp badge to event lists

steventlamb:


src/nyc_trees/apps/home/tests.py
```diff
@@ -31,7 +31,9 @@ def requery_user(self):
     def test_visit_marks_boolean_true(self):
         """assert that any successful view function, when wrapped with
         the decorator, will mark the associated boolean"""
-        view = lambda request: {}
+        def pedantic_flake_8_view(request):
```
jwalgran: Do we need the assignment or can just:

```python
def view(request):
    return {}
```

steventlamb: we do not need the assignment, I did this to document why this was happening in as few characters as possible

steventlamb:
@jwalgran followed your suggestion on the lintfixer

jwalgran:
+1 the +RSVP label now shows up on all the event lists.

Issue #495, opened 2015-02-11
Group detail: "Followers" count does not use live data


RickMohr:
The "Followers" count is still hard-coded to 653

PR #557, opened 2015-02-16
Include updating follow count on group detail

jwalgran:
This commit adds the correct follower count to the group detail page.

Updating the count when a person clicks the "Follow" button required changing the follow/unfollow views and responses, but the change had to be "backward compatible," since the group list page shows follow/unfollow buttons without counts. To implement this, I added a POST argument that is fed through to the template which then decides whether the button should be rendered alone, or with a count next to it.

### DIFF
--- a/src/nyc_trees/apps/users/routes/group.py
+++ b/src/nyc_trees/apps/users/routes/group.py
@@ -13,7 +13,7 @@ from apps.core.decorators import group_request, group_admin_do
 from apps.users.views import group as v
 
 
-render_follow_button = render_template('groups/partials/follow_button.html')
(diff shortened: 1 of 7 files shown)


maurizi:
Looks good to me (haven't tested it)

kdeloach:
I would lean slightly more towards creating separate routes for each follow button to eliminate the whole `render_follow_button_without_count` code path. It seems reasonable to create separate urls/routes here since the follow button returns different HTML depending on the context.

kdeloach:
+1 Tested and working.

jwalgran:
Fixes #495

PR #558, opened 2015-02-16
Link directly to training events from training instructions page

### LINKED ISSUE #523: Tab/hash list need to be implemented for "All Events" page
Clicking All, Mapping, or Training tab should result in a hash change so we can link directly to that tab.

URL: [http://localhost:8000/event/](http://localhost:8000/event/)

kdeloach:
Since we can link directly to the event tabs via query string, we don't have
to do any additional work to support tab hash on the events page.

Closes #523


steventlamb:
+1 do you think that the link from instructions should also scroll down to 'browse events' as well as activating the training filter?

steventlamb:
(tested, works)

kdeloach:
I'd like to hold off on scrolling to the Browse Events section for now.

PR #574, opened 2015-02-16
Hide group URL on detail page if not set

### LINKED ISSUE #553: Visit button shown by website field on group detail page even when there is no website
I'm fine with having the website section visible even if there isn't a URL, but a button that goes no where and just reloads the page is confusing.

jwalgran:
A button that goes nowhere is confusing to a person viewing the group detail page.

Fixes #553


kdeloach:
+1

steventlamb:
+1, tested

PR #584, opened 2015-02-16
Add more detailed tags to AMIs

hectcastro:
This changeset adds the following tags to all AMIs:

- Created (human-readable timestamp of when the AMI was created)
- Service (name of the service this AMI contains)
- Environment (target environment for this AMI)

---

**Example of AMI tagging**:

ec2_management_console


jwalgran:
+1 a definite help for future diagnostics and reporting

PR #586, opened 2015-02-16
Only initialize group detail map if it has territory.

### LINKED ISSUE #585: JS exception on group detail if the group has no territory
```
 Uncaught Error: Map container not found.leaflet-src.js:2009
L.Map.L.Class.extend._initContainerleaflet-src.js:1563
L.Map.L.Class.extend.initializeleaflet-src.js:229
NewClassleaflet-src.js:2345
L.mapmap.js:32 
creategroup_detail.js:22
require../js/src/group_detail.js../copyEventUrl_prelude.js:1
s_prelude.js:1
e_prelude.js:1 (anonymous function)
```

maurizi:
If a group has no territory, we were not creating the "map" element, but
still trying to initialize the leaflet map.

Fixes #585

### DIFF
--- a/src/nyc_trees/js/src/group_detail.js
+++ b/src/nyc_trees/js/src/group_detail.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var fetchAndReplace = require('./fetchAndReplace'),
+    $ = require('jquery'),
     mapModule = require('./map');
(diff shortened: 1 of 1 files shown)


RickMohr:
+1

PR #587, opened 2015-02-16
Don't show 'Back' button on exercise failure

### LINKED ISSUE #572: Training shouldn't show "Try Again" and "Previous" buttons
The training button area should either show:

- Previous and Next
- Previous
- Next
- Try Again

In situations where it shows Try Again and Previous, I think it should just show Try Again.

This is an example of where this is happening:

image

steventlamb:
fixes #572 on github


maurizi:
+1, works

Issue #565, opened 2015-02-16
Legend is incorrect on Reservations page


RickMohr:
On the Reservations page the legend indicates 5 categories, but we're only showing "Reserved by you". We should only show 1 category on the legend.

dboyer:
Connected to #566

kdeloach:
We should verify if this has been resolved by #594.

jwalgran:
@steventlamb Sorry I stepped on you in #594.

steventlamb:
np, reassigning to you

PR #594, opened 2015-02-16
Show all blockfaces on reservation page

### LINKED ISSUE #566: Blank "Reservations" map is confusing
When a user first sees the Reservations page it shows no blockfaces. That is confusing since it's not clear that the page shows only the user's reservations.

Possible solutions:
* Call it the "My Reservations" page or make the page intent clear in the text
* Show other blockfaces in gray
* Include a "You have no reservations yet" message

jwalgran:
An individual mapper will have at most 20 blockfaces reserved at any given time. This left us with a largely empty reservation map, so we decided to show the other blockfaces as unavailable.

Fixes #566

### DIFF
--- a/src/tiler/sql/user_reservations.sql
+++ b/src/tiler/sql/user_reservations.sql
@@ -2,11 +2,14 @@
   <% if (is_utf_grid) { %>
   ST_AsGeoJSON(block.geom) AS geojson,
   <% } %>
-  block.geom, block.id, 'reserved' as survey_type
+  block.geom, block.id, CASE WHEN reservation.user_id = <%= user_id %> THEN  'reserved' ELSE 'unavailable' END as survey_type
   FROM survey_blockface AS block
-  INNER JOIN survey_blockfacereservation AS reservation
+  LEFT OUTER JOIN survey_blockfacereservation AS reservation
     ON (block.id = reservation.blockface_id
         AND reservation.canceled_at IS NULL
+        AND reservation.user_id = <%= user_id %>
         AND reservation.expires_at > now() at time zone 'utc')
-  WHERE reservation.user_id = <%= user_id %>
+  <% if (is_utf_grid) { %>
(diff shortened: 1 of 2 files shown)


jwalgran:
Fixes #565 

screen shot 2015-02-16 at 4 28 41 pm

src/tiler/sql/user_reservations.sql
```diff
@@ -1,12 +1,12 @@
...
-  block.geom, block.id, 'reserved' as survey_type
+  block.geom, block.id, CASE WHEN reservation.user_id = <%= user_id %> THEN  'reserved' ELSE 'unavailable' END as survey_type
   FROM survey_blockface AS block
-  INNER JOIN survey_blockfacereservation AS reservation
+  LEFT OUTER JOIN survey_blockfacereservation AS reservation
     ON (block.id = reservation.blockface_id
         AND reservation.canceled_at IS NULL
+        AND reservation.user_id = <%= user_id %>
         AND reservation.expires_at > now() at time zone 'utc')
-  WHERE reservation.user_id = <%= user_id %>
```
maurizi: Since there is no interactivity for `unavailable` blockfaces, it would make sense to me to keep this `WHERE` clause in place when `is_utf_grid` is true.

That would prevent a "hand" pointer over top of blocks that you can't actually click on, and greatly slim down the grid size.
Alternatively, we could show some toast in the UI when you click on an unavailable blockface (but I'd prefer the first option much more).

maurizi:
Looks like you introduced some lint: `/opt/app/apps/survey/views.py:10:1: F401 'map_legend' imported but unused`

jwalgran:
@maurizi updated

src/tiler/sql/user_reservations.sql
jwalgran: Done

maurizi:
+2

PR #607, opened 2015-02-17
Add footer to non-map pages

### LINKED ISSUE #516: Add Parks logo in footer
From the visual designs. I realize the text associated with the logo is pending but let's add the suggested text for now.

### LINKED ISSUE #601: Footer code needed on all pages except maps
I want to have the following footer code inserted into base.html in the extra_content area (just before the closing body tag). However, I don't want it on any of the pages with full screen maps. If we show the footer on the map pages, it will take up so much of the screen that the map won't be useful.

I'm am hoping a developer can figure out the best way to do this. Ideally, I think we'd be able to set which pages to exclude instead of having to add additional code to every new non-map page that gets added. Also, this footer content will be the same on every page that has it (all non map-pages).

```
        {% block extra_content %}
            <footer>
                <div class="container">
                    <img class="section-about-logo pull-left" src="/static/img/logo-nycparks.png" srcset="/static/img/logo-nycparks@2x.png 2x">
                    <p>A project of New York City Parks</p>
                </div>
            </footer>
        {% endblock extra_content %}
```

jwalgran:
The footer is excluded from full screen map page to maximize the viewable area.

Fixes #601
Fixes #516

screen shot 2015-02-17 at 8 48 29 am

screen shot 2015-02-17 at 8 48 42 am

screen shot 2015-02-17 at 8 49 09 am

### DIFF
--- a/src/nyc_trees/apps/core/templates/base.html
+++ b/src/nyc_trees/apps/core/templates/base.html
@@ -70,6 +70,15 @@
         {# This block exists for placing elements that need to be children of the body, like bootstrap modals #}
         {% block extra_content %}
         {% endblock extra_content %}
+
+        {% block footer %}
+            <footer>
+                <div class="container">
+                    <img class="section-about-logo pull-left" src="/static/img/logo-nycparks.png" srcset="/static/img/logo-nycparks@2x.png 2x">
+                    <p>A project of New York City Parks</p>
+                </div>
+            </footer>
+        {% endblock footer %}
     </body>
     <script type="text/javascript">
       {% include 'core/partials/config.js' %}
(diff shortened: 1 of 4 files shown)


maurizi:
@jwalgran Confusingly, only the progress map extends from `base_map.html`.

The reservation maps extend from `base_reservations.html` and the event map/list page extends from `base_event.html`

We should probably remove the footer in those pages as well.

jwalgran:
More full map templates now exclude the footer

maurizi:
Looks good to me (I haven't tested it)

jfrankl:
Just tested it. Looks excellent. +1

jfrankl:
Sorry, I didn't mean to close it.

PR #608, opened 2015-02-17
Feature/style footer

jfrankl:
I wanted to get a footer that works on mobile in for user testing, but getting this to look right on desktop is going to take more effort. For now, I am just hiding the footer on desktop.


jwalgran:
+1 footer is hidden when the viewport is wide

PR #611, opened 2015-02-17
Show reservations link in side nav

kdeloach:
Add "Reservations" link back to the side nav and also adds some instructions on how to become an individual mapper.

How to test this:
1. Login as a non-individual mapper user
2. Go go Reservations page
3. Click New (should go to instructions page)
4. Grant user individual mapping status
5. Go go Reservations page
6. Click New (should go to reserve blockface page)

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -172,3 +172,7 @@ def admin_extend_blockface_reservation(request, blockface_id):
 def admin_blockface_available(request, blockface_id):
     # TODO: implement
     pass
+
+
+def reservations_instructions(request):
+    pass
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -105,6 +105,11 @@ class User(NycModel, AbstractUser):
     def training_complete(self):
         return self.online_training_complete and self.field_training_complete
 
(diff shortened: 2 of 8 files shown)


src/nyc_trees/apps/survey/views.py
```diff
@@ -172,3 +172,7 @@ def admin_extend_blockface_reservation(request, blockface_id):
 def admin_blockface_available(request, blockface_id):
     # TODO: implement
     pass
+
+
+def reservations_instructions(request):
+    pass
```
maurizi: There is no need for this "empty" view.  The `render_template` decorator can work without actually decorating anything.

maurizi:
Overall this looks good, though I haven't tested it.  You may want to squash your commits, stuff from the first commit references the second commit.

jwalgran:
Welcome back, Reservations sidebar menu item. I ran through the instruction flow and it works nicely.

Sneaky edge case. If an admin revokes access by setting ``individual_mapper`` to "No", then future clicks on the "New" reservation button redirect to the instructions and "Step 4" will forever tell them to click a button that is hidden.

We probably need to have some "Sorry" text on the instructions page when ``individual_mapper`` is "No"

I think it's cool to open a new issue to cover that edge case, or roll it into this PR.

+1

PR #614, opened 2015-02-17
Fix floppyforms checkbox attributes template

### LINKED ISSUE #593: Individual mappers checkbox selection doesn't match actual value on group admin
Seems like this checkbox value is no longer being populated correctly. If you mark it checked, then come back to this screen, it is still unchecked.

image

kdeloach:
Fixes #593


maurizi:
+1

Issue #400, opened 2015-01-29
Prevent setting any dates < now()


dboyer:
As demonstrated in sprint review

kdeloach:
This was caused by submitting an invalid start date.

jwalgran:
Changing the title to reflect the fix we need

steventlamb:
By my search, these are the fields in question:
`Event.begins_at`
`Event.ends_at`
`BlockfaceReservation.expires_at`
`BlockfaceReservation.canceled_at`
`User.requested_individual_mapping_at`

Some are more obvious candidates than others, but some are unclear. I'm also not sure if we should lock down any of these wholesale - the django admin is for trusted users. What if they have to fix something?

@kdeloach which if these fields did you mean by 'invalid start date'?

kdeloach:
I submitted a start date like `2/44/1111` or something along those lines when I was adding a new event.

steventlamb:
Got it, thanks. Does anyone know why the original title was `Error Adding a Reservation`? Is this a compound issue?

dboyer:
That was the original title because I entered the issue really quickly during/right after sprint review when we didn't have info on what caused the problem.

PR #617, opened 2015-02-17
Allow editing past or in progress events.

### LINKED ISSUE #564: Validation error when editing an event in progress
If you want to update a minor detail for an existing event that is already in progress, you get a validation error.

    begins_at_time
        - Event must not begin in past

maurizi:
Switches to testing against 'ends_at' instead of 'begins_at', to allow
editing events that are in progress.

Also allows editing past events, provided that the times have not been
changed.

Fixes #564

### DIFF
--- a/src/nyc_trees/apps/event/forms.py
+++ b/src/nyc_trees/apps/event/forms.py
@@ -49,33 +49,57 @@ class EventForm(ModelForm):
 
     def clean(self):
         cleaned_data = super(EventForm, self).clean()
-        right_now = now().astimezone(timezone('US/Eastern'))
+        est_tz = timezone('US/Eastern')
+        right_now = now().astimezone(est_tz)
         today = right_now.date()
 
         if 'date' in cleaned_data:
             cleaned_date = cleaned_data['date']
 
-            # first check if the date is in the past and if so, raise on date
-            if cleaned_date < today:
-                Event.raise_date_past('date')
-            if 'begins_at_time' in cleaned_data:
-                begins_at_time = cleaned_data['begins_at_time']
+            # first check if the date was changed to be in the past and if so,
+            # raise on date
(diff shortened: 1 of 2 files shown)


src/nyc_trees/apps/event/forms.py
```diff
@@ -49,25 +49,36 @@ class Meta:
...
 
         if 'date' in cleaned_data:
             cleaned_date = cleaned_data['date']
 
-            # first check if the date is in the past and if so, raise on date
-            if cleaned_date < today:
-                Event.raise_date_past('date')
-            if 'begins_at_time' in cleaned_data:
-                begins_at_time = cleaned_data['begins_at_time']
+            # first check if the date was changed to in the past and if so,
```
steventlamb: changed to -> changed to be

steventlamb:
This allowed me to create an event in the past, which is a regression. See https://github.com/azavea/nyc-trees/issues/400

maurizi:
@steventlamb This is fixed.

steventlamb:
I now get the wrong error message for an out of bounds issue. In this case, it does not begin in the past. It begins in the future and ends in the past.
screenshot from 2015-02-17 16 00 44

maurizi:
@steventlamb I fixed that too.

steventlamb:
submitting this:
screenshot from 2015-02-17 16 42 11

gets:

screenshot from 2015-02-17 16 42 29

maurizi:
... and that one is fixed.

kdeloach:
+1 Let's merge this.

steventlamb:
I edited a date in the past and changed the `date` but not the `time`. It saved. Is that a bug?

maurizi:
@steventlamb I can't replicate that bug...

PR #620, opened 2015-02-17
Progress page action bar

maurizi:


src/nyc_trees/apps/survey/urls/blockface.py
```diff
@@ -31,6 +34,11 @@
     url(r'^(?P<blockface_id>\d+)/reservation-popup/$',
         reserved_blockface_popup, name='reserved_blockface_popup'),
 
+    # Note: changes here must be kept in sync with
+    # src/nyc_trees/js/src/reservationPage.js
```
maurizi: Note to self, wrong JS file

jwalgran:
The "reserve it" link does not have the correct href

screen shot 2015-02-17 at 2 00 34 pm

maurizi:
@jwalgran what should it link to?  @RickMohr just had it as `#`

jwalgran:
You can zoom to it on the reservation page by using the URL hash http://localhost:8000/blockface/#24333

jwalgran:
+1

PR #623, opened 2015-02-17
Refactor quiz classes

kdeloach:
This merges Question and MultiChoiceQuestion (all questions are
multiple choice). This also adds strongly typed classes for the answer
field to differentiate between checkbox & radio button questions.

### DIFF
--- a/src/nyc_trees/apps/home/training/types.py
+++ b/src/nyc_trees/apps/home/training/types.py
@@ -169,7 +169,7 @@ class Quiz(object):
     """
     title - string
     questions - iterable<Question>
-    passing_score - int; How many correct answers are needed to pass the quiz
+    passing_score - int; Number of correct answers needed to pass the quiz
     """
     def __init__(self, title, questions, passing_score):
         assert len(questions) > 0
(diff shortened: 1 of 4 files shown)


src/nyc_trees/apps/home/training/types.py
```diff
@@ -210,38 +210,40 @@ def score(self, answers):
...
-        super(Question, self).__init__(text, choices, [answer])
+    def __init__(self, *answers):
+        self.answers = list(answers)
+
 
-    def is_multiple_choice(self):
-        return False
+class SingleAnswer(MultipleAnswer):
+    """
+    answer - int
```
kdeloach: TODO: Documentation

src/nyc_trees/apps/home/training/types.py
```diff
@@ -210,38 +210,40 @@ def score(self, answers):
...
-        self.answer = list(answer)
+        self.answer = answer
 
     def is_correct(self, candidate):
-        return candidate and set(candidate) == set(self.answer)
+        return candidate and set(candidate) == set(self.answer.answers)
 
-    def is_multiple_choice(self):
-        return True
+    def allow_multiple_selections(self):
```
steventlamb: how about `has_multiple_answers`

kdeloach: The rationale behind this name is that it's possible to allow multiple selections (checkbox) but only have 1 "answer". I think either name is an improvement though.

steventlamb:
+1 works and makes sense

PR #627, opened 2015-02-17
Redirect group admin to admin event checkin page

### LINKED ISSUE #526: Event "starting soon" notification should take group admins to admin check-in page
Currently, it takes group admins to the user check-in page.

image

kdeloach:
This fixes an edge case where, if a group admin RSVPs to their own
event, and then clicks the "Starting soon" header notification, they
are taken to the user checkin page rather than the admin checkin page.

Fixes #526


RickMohr:
+1

PR #632, opened 2015-02-18
Fix layout issue when error message is displayed

### LINKED ISSUE #492: Edit event: layout issue when error message displayed
Try to save an event with a start time in the past.

1) Layout of the input fields changes
2) Should we perhaps be showing this error message in red?

image

kdeloach:
It looks like this was somewhat fixed by recent design changes. I moved
the validation message outside of the input div so that the error
message does not increase the length of the input field.

Also, I swapped field labels/errors because IMO it makes it much clearer to
see which form fields the errors are talking about.

There is remaining work to style validation messages.

Fixes #492

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/partials/event_form.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/event_form.html
@@ -5,32 +5,32 @@
         {{ form.non_field_errors }}
 
         <div class="field">
-            {{ form.title.errors }}
             {{ form.title.label_tag }}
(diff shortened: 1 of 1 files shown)


maurizi:
+1, I was just discussing the need for this with @jfrankl yesterday.

RickMohr:
+1 

Can you add an issue for styling the validation errors (perhaps with screenshot)?

kdeloach:
Added an issue for validation errors styling.

PR #635, opened 2015-02-18
Fix error when saving user profile

### LINKED ISSUE #633: 500 server error when try to change settings on user profile page
I tried to make my groups and contributions publicly visible and got a server error when I saved.

kdeloach:
This restores hidden form fields needed on the user profile page.
Broken by 7c543f46e7f82557187e057e2c178090e46fd81c.

Fixes #633

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/partials/event_section.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/event_section.html
@@ -1,4 +1,6 @@
 {% with event=info.event %}
+{% block extra_markup %}
+{% endblock extra_markup %}
 <a href="{{ event.get_absolute_url }}" class="event block item rightarrow">
     <div class="row">
         {% include 'event/partials/event_calendar.html' %}


kdeloach:
Not sure if I solved this the best way but it seems to be working.

src/nyc_trees/apps/event/templates/event/partials/event_section.html
```diff
@@ -1,4 +1,6 @@
 {% with event=info.event %}
+{% block extra_markup %}
+{% endblock extra_markup %}
```
steventlamb: Are you adding this because you need null out what's inherited from the parent? If so, a template comment would go a long way.

kdeloach: It's actually so that templates that inherit from this one, can inject fields here. This happens on the "edit user setting page", which displays the event list with the addition of a checkbox next to each event (for the email opt-in).

steventlamb:
oh, ok, that's a pretty typical django template pattern. In that case, I'd advise adding that into the commit message.

kdeloach:
Thanks, will do.

steventlamb:
+1

PR #641, opened 2015-02-18
Fix metric name used by Tasseo for e-mail graphs

### LINKED ISSUE #636: Email success is not appearing in graphite
When registering a new account, I receive an email but the ``statsite.counts.email.message.success`` counter does not increment.

hectcastro:
This changeset fixes the metric name used by Tasseo for e-mail graphs. Because `django-statsd` handles the StatsD client setup, all metrics that come through the following import get tucked under the `django` metric namespace:

```python
from django_statsd.clients import statsd
```

Attempts to resolve #636.


jwalgran:
+1 I did not test this interactively because it relies on our email backend that only runs on EC2

hectcastro:
Just tested the registration flow with this patch on the monitoring server:

cursor_and_tasseo

jwalgran:
Love those green bars

PR #642, opened 2015-02-18
Bug/512 restrict highest map zoom level

steventlamb:


jwalgran:
+1 all maps load correctly

steventlamb:
I kept the hacky solution in VC just in case. It's not optimized.

PR #645, opened 2015-02-19
Test that logged out users can view group detail page

kdeloach:
This tests for a regression against
6d2bcb358589682c8e0d864b95449784a1e1fc73.


jwalgran:
+1 test passes

PR #655, opened 2015-02-20
Flesh out Survey and Tree models

RickMohr:
* Include fields from TreeKit app (minus those Parks doesn't want )
* Include fields and choices from the Parks "Variables" document
* Values for `Trees.problems` will be a list of one or more problems. Included choices as a comment pending implementation.


src/nyc_trees/apps/survey/models.py
```diff
@@ -55,6 +60,59 @@ class Tree(NycModel, models.Model):
...
+    #     ('MetalGrates', 'Metal grates'),
+    #     ('RootOther', 'Other'),
+    # "Trunk problems"
+    #     ('WiresRope', 'Wires or rope'),
+    #     ('TrunkLights', 'Lights'),
+    #     ('TrunkOther', 'Other'),
+    # "Branch problems"
+    #     ('BranchLights', 'Lights'),
+    #     ('Sneakers', 'Sneakers'),
+    #     ('BranchOther', 'Other'),
```
maurizi: Consider adding a clean method that ensures the `problems` field is a comma separated list containing at most 1 of each of these choice values.

We don't want incorrect fields getting in here via the Admin site.

src/nyc_trees/apps/survey/models.py
```diff
@@ -55,6 +60,59 @@ class Tree(NycModel, models.Model):
     # We do not anticipate deleting a Species, but we definitely
     # should not allow it to be deleted if there is a related Tree
     species = models.ForeignKey(Species, null=True, on_delete=models.PROTECT)
+    distance_to_tree = models.FloatField()
+    distance_to_end = models.FloatField(null=True)
```
maurizi: Shouldn't this have `blank=True` as well?

maurizi:
It looks like the new non-nullable fields broke a few unit tests.  Probably just need to set something for those fields.

maurizi:
Other than the comments I left, this looks good to me.

maurizi:
Consider adding some simple validation in the model's `clean` method to verify that incompatible fields are not present together (i.e. no circumference for "dead" trees).

Alternatively, you could add a TODO comment and a Github issue for doing so, and defer it to a later PR.

src/nyc_trees/apps/survey/models.py
RickMohr: Done.

src/nyc_trees/apps/survey/models.py
RickMohr: Done.

RickMohr:
Actually they want circumference for dead trees. I added code to clear fields collected only for living trees If the status is not "Alive".

RickMohr:
@maurizi, please take another look.

RickMohr:
Closing in favor of #675, which includes a slightly modified version.

PR #656, opened 2015-02-20
Centralize Nginx log_format directive

hectcastro:
This changeset moves the Nginx `log_format` directive into the `http` block (via an include). The newly created `nyc-trees.nginx` role handles providing the `logstash_json` format for application server and tiler `server` blocks to use.

This also prevents Nginx errors citing that `log_format` must be defined in the `http` block.


RickMohr:
+1

PR #657, opened 2015-02-20
Add HTTPS support for the application server stack

hectcastro:
This changeset adds HTTPS support to the application server stack by:

- Associating a certificate ARN with the load balancer
- Accepting connections on load balancer port 443
- Ensuring Nginx redirects inbound HTTP requests to HTTPS
- Add health-check endpoint as an exception to the line item above


kdeloach:
I can review this later today.

kdeloach:
+1

PR #661, opened 2015-02-20
Add unit tests for trusted mapper and individual mapper status workflows

kdeloach:
This tests the workflow of attaining mapper status and also fixes a few subtle bugs.


maurizi:
+1, Code looks good and the tests pass on staging.  I don't really have the time to test this out myself though, if you feel that is necessary feel free to assign this to someone else.

PR #662, opened 2015-02-20
Zoom to reservations

kdeloach:
We mentioned this enough times that it ought to be done (even if this page ends up getting remodeled).


maurizi:
Looks good to me

PR #665, opened 2015-02-23
Select the blockface present in the URL hash by default

kdeloach:
Instead of just zooming to a blockface on page load, select it as well. This change impacts the Reservations page and the Survey page.


RickMohr:
+1

PR #668, opened 2015-02-24
Add UI for blockface endpoint selection on survey page

kdeloach:
Needs design support.


RickMohr:
+1

Clear code, works well. Great to see it handle changing your mind on blockface or endpoint.

PR #669, opened 2015-02-24
Set survey bounds to reservations extent

kdeloach:
Accessing the survey page as an individual mapper will set the map
bounds to the extent of your reservations.


RickMohr:
+1

PR #671, opened 2015-02-24
Fix events in-progress appearing in the Past events tab

kdeloach:


src/nyc_trees/apps/event/event_list.py
```diff
@@ -49,10 +49,11 @@ def get_filterset(name):
             ]),
             EventList.chronoFilters: OrderedDict([
                 (_CURRENT, lambda qs: (qs
-                                       .filter(begins_at__gte=right_now)
```
steventlamb: I don't think this is quite right. I think we want 100% of events in qs to appear on at least one of two tabs. This is implied by the fact that the previous clauses were mirrored, `begins_at__lt`, `begins_at__gte`. Now, events in the past are caught correctly, events in-progress are caught correctly, but future events are dropped. I think in this case current means present+future. We should double-check that, but if so, we should drop the `begins_at__lt=right_now`.

kdeloach: Good catch, updated.

RickMohr:
+1

PR #672, opened 2015-02-24
Group admin should always be able to reserve blockfaces

kdeloach:
Group admins should not have to go through the trusted mapper workflow
in order to reserve blockfaces in their own territory.

### DIFF
--- a/src/tiler/sql/user_reservable.sql
+++ b/src/tiler/sql/user_reservable.sql
@@ -10,24 +10,28 @@
       block.geom,
       block.id,
       turf.group_id,
-      grp.slug AS group_slug,
+      core_group.slug AS group_slug,
       CASE
         WHEN (block.is_available AND reservation.id IS NULL
-              AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'
+              AND (turf.id IS NULL
+                   OR trustedmapper.id IS NOT NULL
+                   OR core_group.admin_id = <%= user_id %>)) THEN 'available'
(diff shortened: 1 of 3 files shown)


jwalgran:
Looks good. I'll pull it down and give it a run through on Wednesday.

jwalgran:
+1
I created a brand new user, ran through the training, and assigned them as the admin of one of my groups. That user saw the correct tiles on the new reservation page map and was allowed to make reservations in the group's territory.

src/tiler/sql/user_reservable.sql
```diff
@@ -10,24 +10,29 @@
...
       FROM survey_blockface AS block
+      INNER JOIN core_user ON core_user.id = <%= user_id %>
       LEFT OUTER JOIN survey_territory AS turf
         ON block.id = turf.blockface_id
-      LEFT JOIN core_group AS grp
-        ON grp.id = turf.group_id
+      LEFT JOIN core_group ON core_group.id = turf.group_id
       LEFT OUTER JOIN survey_blockfacereservation AS reservation
         ON (block.id = reservation.blockface_id
             AND reservation.canceled_at IS NULL
```
maurizi: We should drop the join to `core_user` and replace references `core_user.id` with `<%= user_id %>`

We need to update the [cache buster](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/survey/layer_context.py#L23-L25) for this query to include `Group` now.

kdeloach: Done.

PR #675, opened 2015-02-24
Survey and Tree models and views

RickMohr:


RickMohr:
@maurizi, I thought you would want first crack at this, but feel free to assign back to me if you are swamped.

src/nyc_trees/apps/survey/views.py
```diff
@@ -189,9 +195,63 @@ def start_survey_from_event(request, event_slug):
...
+            return HttpResponseForbidden(
+                "Blockface is not in group's territory.")
+    else:
+        if not _query_reservation(request.user, blockface.id).exists():
+            return HttpResponseForbidden('You have not reserved this blockface.')
+
+    if survey.has_trees and len(tree_list) == 0:
+        return HttpResponseBadRequest('Trees expected but absent')
+    if not survey.has_trees and len(tree_list) > 0:
+        return HttpResponseBadRequest('Trees not expected but present')
```
maurizi: Instead of checking what was sent for `has_trees`, I think we should just ignore and always set it ourselves:
```python
survey.has_trees = len(tree_list) > 0
```
This will simplify the client JS code.

src/nyc_trees/apps/survey/models.py
```diff
@@ -40,6 +41,11 @@ class Survey(NycModel, models.Model):
                                  related_name='surveys_as_teammate',
                                  blank=True)
     is_flagged = models.BooleanField(default=False)
+    has_trees = models.BooleanField(default=True)
+    is_mapped_in_blockface_polyline_direction = models.BooleanField(
+        default=True)
+    is_left_side = models.BooleanField(default=True)
```
maurizi: I think we should remove the `default=True` for all 3 of these fields.

If we don't get the information for these fields from the client, it's an error.
I'd rather catch the bug earlier by failing validation during testing, than just silently using an incorrect default value.

maurizi:
Looks like your PR has failing tests?

maurizi:
@azavea-bot rebuild

src/nyc_trees/apps/survey/models.py
RickMohr: That was my inclination too, but Django's default is `None` and I didn't want any `None` values in the DB. That said, It will take explicit validation code to catch omissions either way, and it's better to test for a value of `None` in the `clean` method then to test for the presence of fields in `submit_survey`, so I'll change it.

src/nyc_trees/apps/survey/views.py
RickMohr: Note the UI has a "No trees?" button that needs special handling. I put in these checks to guard against mistakes in that code, so I'd rather leave them in. The client code seems straightforward since `has_trees` will always be true for one branch and always false for the other.

RickMohr:
@maurizi, all fixed

RickMohr:
@azavea-bot rebuild

RickMohr:
@azavea-bot rebuild

RickMohr:
@azavea-bot rebuild

maurizi:
+1

RickMohr:
@azavea-bot rebuild

PR #678, opened 2015-02-25
Endpoint to flag a survey

RickMohr:


kdeloach:
What does it mean to be flagged? User is not confident with the survey?

kdeloach:
+1 Looks good and tests pass.

PR #680, opened 2015-02-25
Update MIGRATIONS.md

RickMohr:


hectcastro:
Added a section at the top to test if the agent is running and links to a guide that shows how to automatically launch `ssh-agent` on Windows if you're using Git Bash.

:+1:

PR #682, opened 2015-02-25
Add custom admin for User model

### LINKED ISSUE #130: Creating a user with the admin interface fails to set password
I made a user 'a' with `python manage.py createsuperuser` with password 'a' and if you look at the user detail in the admin interface, the password field is a crypto hash of a, as expected. This user can login with a/a.

I made a user 'b' using the admin interface with 'b', and when viewing their user detail in the admin interface, their password is *not* a crypto-hash of 'b', it *is* 'b'. This user cannot login with b/b.

I am assuming that the password field is expecting a raw crypto-hash in the admin interface. IIRC, the django admin interface handles user creation quite smoothly under normal circumstances. I also noticed that it is strange that the password field is at the top. Therefore I expect the problem is that we are accidentally bypassing the normal password workflow of the admin interface with our own field/form.

kdeloach:
This fixes an issue where you are unable to modify user passwords via
the Django admin.

Reference:
http://stackoverflow.com/questions/15012235/using-django-auth-useradmin-for-a-custom-user-model

Fixes #130

Not a priority for sprint review.

### DIFF
--- a/src/nyc_trees/apps/core/admin.py
+++ b/src/nyc_trees/apps/core/admin.py
@@ -3,7 +3,11 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from django.core.exceptions import ValidationError
+
 from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.forms import UserChangeForm, UserCreationForm
 
 from apps.core.models import User, Group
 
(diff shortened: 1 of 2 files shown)


steventlamb:
+1 NICE!!! Works great.

kdeloach:
I would like to mention that this technically removes existing functionality... New fields added to the DB won't appear in the admin unless explicitly added.

jwalgran:
There are two other validations in our user model that we need to clean

* Duplicate email is not allowed
* Minors can't have public profiles

jwalgran:
We should add an attention grabbing content at the top of the user model explaining that there is a custom admin form in place and that new fields need to be added over there.

steventlamb:
Right, that's kind of a bummer. One alternative would be to just hide the password field from the previously used form and throw a link in it.

kdeloach:
The good news is that the admin will call `clean` on all models during saving, so we don't have to duplicate this validation (for is_minor, and duplicate emails).
I will add an attention grabbing comment on the user model and then merge this in.

kdeloach:
I also added a comment explaining why we needed to reimplement `clean_username` but not any of the other validation logic.

PR #694, opened 2015-02-26
Change default from e-mail address

### LINKED ISSUE #392: No DEFAULT_FROM_EMAIL setting exists

hectcastro:
The new default from e-mail address, owned by Parks. Tested on staging after getting the e-mail account confirmed on staging and production.

Resolves #392.


kdeloach:
+1

jwalgran:
+1 confirmed that this is a verified address in the staging AWS

screen shot 2015-02-26 at 3 24 50 pm

PR #698, opened 2015-02-26
Fix error on survey page due to non-existent element

kdeloach:
Here's my proposed fix to prevent the search controller from causing an error for pages (with maps) that do not have all the correct divs/templates setup.


maurizi:
-1.

I think we should leave search and legend disabled on the survey page for now.

The other commit certainly can't hurt though.

kdeloach:
OK.

Issue #673, opened 2015-02-24
Dashboard "This week's events" shows events from last week


kdeloach:
Screenshot taken on 2/24.

![image](https://cloud.githubusercontent.com/assets/43062/6358802/4bfa7fce-bc3c-11e4-89cb-ceb5859de9cc.png)

Also affects the "Your events this week" section on `/event`.

![image](https://cloud.githubusercontent.com/assets/43062/6358849/9bc74ee2-bc3c-11e4-8f8b-f2b4a53b2b55.png)

dboyer:
I believe this is connected to #622

PR #701, opened 2015-02-27
Fix old events appearing where they shouldn't

### LINKED ISSUE #622: Old events still show on event list
There's an event from Feb 12 showing at the top of the events list. Events should be removed from the list after the date of the event has passed.

kdeloach:
This prevents old events from appearing on the Dashboard and the Events page.

Fixes #622 and #673.

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/event_list_page.html
+++ b/src/nyc_trees/apps/event/templates/event/event_list_page.html
@@ -35,20 +35,20 @@
     <div class="tab-pane active" id="list">
       <section>
         {# TODO: make event_list return a totally empty context to avoid introspection #}
+        <h4>Your Events this week</h4>
+        <div data-class="event-list-container">
         {% if immediate_events.event_infos %}
-          <h4>Your Events this week</h4>
-          <div data-class="event-list-container">
-              {% include "event/partials/event_list.html" with event_list=immediate_events %}
-          </div>
+          {% include "event/partials/event_list.html" with event_list=immediate_events %}
+        {% else %}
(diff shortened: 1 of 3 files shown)


kdeloach:
Fixing unit tests now.

kdeloach:
Fixing this up now to address some concerns.

kdeloach:
I think the issues we talked about should be resolved now.

src/nyc_trees/apps/event/templates/event/event_list_page.html
```diff
@@ -35,20 +35,20 @@
...
+        <h4>Your Events this week</h4>
+        <div data-class="event-list-container">
         {% if immediate_events.event_infos %}
-          <h4>Your Events this week</h4>
-          <div data-class="event-list-container">
-              {% include "event/partials/event_list.html" with event_list=immediate_events %}
-          </div>
+          {% include "event/partials/event_list.html" with event_list=immediate_events %}
+        {% else %}
+          <p>You haven't RSVPed to any event yet.</p>
```
steventlamb: This might be a grammatical error. I think event should be events.

steventlamb:
tested, +1 pending spec confirmation on #622

src/nyc_trees/apps/event/templates/event/event_list_page.html
kdeloach: I think you're right... that's what I had originally.

kdeloach:
Spec confirmed (never display past event on event list) so I am merging this in.

PR #702, opened 2015-02-27
Switch to using django-waffle for setting the soft launch settings.

maurizi:
~~To put the site into soft launch mode, create a "soft_launch" flag
from the Django Admin site.~~

Loading the `soft-launch-flag.json` fixture with the `loaddata` management command will create a `full_access` flag in the Django admin site.

Editing this flag will allow switching the site from soft launch mode to full access or vice versa, and it can be controlled to only allow full access for specific users.


jwalgran:
We need to enable "soft launch" for everyone _except_ for the people that we enumerate. This is the inverse of the behavior of a Waffle flag, which you can enable for specific people. To get the functionality we are looking for think we need to make a ``full_access`` flag instead of a ``soft_launch`` flag.

maurizi:
@jwalgran That makes sense, though it will make things frustrating for developers, since the default when a flag has not been defined yet is to make it inactive.  Oh well, I guess you can't have your cake and eat it too.

I'll invert the logic and reassign this back to you.

jwalgran:
I would also like to add a line to ``setupdb.sh`` to ensure that the flag is created in the DB with the name and settings that we expect, if the flag does not already exist.

maurizi:
@jwalgran I've updated this, can you take another look?

scripts/aws/setupdb.sh
```diff
@@ -18,6 +18,8 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
 envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
 # Load block face data
 envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata $DIR/../src/nyc_trees/apps/survey/fixtures/blockface.json
+# Setup soft launch flag
+envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata $DIR/../src/nyc_trees/nyc_trees/fixtures/soft-launch-flag.json
```
hectcastro: It would be nice to use the [Waffle CLI](http://waffle.readthedocs.org/en/latest/usage/cli.html) instead of Django fixtures.

maurizi: Didn't realize that existed.  I've already done it this way, so I'm not inclined to change it.

maurizi: Nevermind, I did end up doing this.

scripts/aws/setupdb.sh
```diff
@@ -19,5 +19,10 @@ envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
 # Load block face data
 envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata $DIR/../src/nyc_trees/apps/survey/fixtures/blockface.json
 
+# Setup soft launch flag if it does not already exist
+if [ envdir /etc/nyc-trees.d/env /opt/app/manage.py flag -l | grep full_access ]; then
```
jwalgran: This did not create a flag for me. I think it is because the conditional is backwards. ``grep`` will exit non-zero if it does not find the text.

maurizi: D'oh.  I'll fix that.

scripts/aws/setupdb.sh
```diff
@@ -19,5 +19,10 @@ envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
...
+    envdir /etc/nyc-trees.d/env /opt/app/manage.py flag full_access --create --deactivate
```
jwalgran: Including ``--deactivate`` sets the ``everyone`` field to ``false`` which means that the flag will never take effect for anyone. We need to omit this option so that the flag will work for the users who should have full access.

scripts/aws/setupdb.sh
maurizi: Fixed

scripts/aws/setupdb.sh
maurizi: Fixed

jwalgran:
I copied ``setupdb.sh`` into my VM and the flag create command fails when when I run it

```
vagrant@app:/opt/app$ cat setupdb.sh
#!/bin/bash

set -e
set -x

DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)

# Export settings required to run psql non-interactively
export PGHOST=$(cat /etc/nyc-trees.d/env/NYC_TREES_DB_HOST)
export PGDATABASE=$(cat /etc/nyc-trees.d/env/NYC_TREES_DB_NAME)
export PGUSER=$(cat /etc/nyc-trees.d/env/NYC_TREES_DB_USER)
export PGPASSWORD=$(cat /etc/nyc-trees.d/env/NYC_TREES_DB_PASSWORD)

# Ensure that the PostGIS extension exists
psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"

# Run migrations
envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
# Load block face data
envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata $DIR/../src/nyc_trees/apps/survey/fixtures/blockface.json

# Setup soft launch flag if it does not already exist
if [ ! envdir /etc/nyc-trees.d/env /opt/app/manage.py flag -l | grep full_access ]; then
    envdir /etc/nyc-trees.d/env /opt/app/manage.py flag full_access --create
fi

# Create training flatpages
envdir /etc/nyc-trees.d/env /opt/app/manage.py make_training_flatpages || true
```

```
vagrant@app:/opt/app$ ./setupdb.sh
+++ dirname ./setupdb.sh
++ cd .
++ pwd
+ DIR=/opt/app
++ cat /etc/nyc-trees.d/env/NYC_TREES_DB_HOST
+ export PGHOST=33.33.33.30
+ PGHOST=33.33.33.30
++ cat /etc/nyc-trees.d/env/NYC_TREES_DB_NAME
+ export PGDATABASE=nyc_trees
+ PGDATABASE=nyc_trees
++ cat /etc/nyc-trees.d/env/NYC_TREES_DB_USER
+ export PGUSER=nyc_trees
+ PGUSER=nyc_trees
++ cat /etc/nyc-trees.d/env/NYC_TREES_DB_PASSWORD
+ export PGPASSWORD=shhdonttellanyone
+ PGPASSWORD=shhdonttellanyone
+ psql -c 'CREATE EXTENSION IF NOT EXISTS postgis;'
NOTICE:  extension "postgis" already exists, skipping
CREATE EXTENSION
+ envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
Operations to perform:
  Synchronize unmigrated apps: debug_toolbar, floppyforms, gis, django_statsd, registration
  Apply all migrations: core, users, sessions, admin, sites, flatpages, waffle, contenttypes, survey, auth, event
Synchronizing apps without migrations:
  Creating tables...
  Installing custom SQL...
  Installing indexes...
Running migrations:
  No migrations to apply.
+ envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata /opt/app/../src/nyc_trees/apps/survey/fixtures/blockface.json
/usr/local/lib/python2.7/dist-packages/django/core/management/commands/loaddata.py:226: UserWarning: No fixture named 'blockface' found.
  warnings.warn("No fixture named '%s' found." % fixture_name)

Installed 0 object(s) from 0 fixture(s)
+ grep full_access ']'
+ '[' '!' envdir /etc/nyc-trees.d/env /opt/app/manage.py flag -l
grep: ]: No such file or directory
./setupdb.sh: line 23: [: missing `]'
+ envdir /etc/nyc-trees.d/env /opt/app/manage.py make_training_flatpages
CommandError: FlatPage '/getting_started/' already exists.
+ true
```

jwalgran:
+1 public and logged in users can't reach non whitelisted pages unless they are associated with the flag

PR #706, opened 2015-03-02
Split tree guard choices into two sections

maurizi:


jwalgran:
+1 I was able to save survey trees with and without guards using the new split UI.

PR #707, opened 2015-03-02
Admin page for managing group territory

RickMohr:
Show these categories of blockface:
* Available
* Reserved by this group
* Reserved by others
* Mapped by this group
* Mapped by others

These are essentially the same categories as shown on the "Reservations" page.

Initially we thought all of the following should be shown as "unavailable":
* Reserved by others
* Mapped by this group
* Mapped by others

I wanted to
* distinguish "Mapped by this group" to give context when working on a group,
  for example if you want to grow territory contiguously.
* distinguish "Reserved by others" from "Mapped by others" to allow seeing
  what yet needs to be mapped in a reapportioning situation.

With good coloring I think these will be useful and not distracting.
And it's easy to go back to 3 categories if we prefer.

Also:
* Fix omission of `BlockfaceReservation` in `get_context_for_territory_survey_layer`
* In `mapModule.addTileLayer` and `addGridLayer`, replace the unused `domId` argument with
  a `queryString` argument so the client can add query arguments to tiler URLs

### DIFF
--- a/src/nyc_trees/apps/survey/urls/census_admin.py
+++ b/src/nyc_trees/apps/survey/urls/census_admin.py
@@ -8,8 +8,9 @@ from django.conf.urls import patterns, url
 from django_tinsel.decorators import route
 
 from apps.core.decorators import census_admin_do
-from apps.survey.views import (admin_blockface_page,
-                               admin_blockface_partial,
+from apps.survey import routes as r
+
+from apps.survey.views import (admin_blockface_partial,
                                admin_blockface_detail_page,
                                admin_extend_blockface_reservation,
                                admin_blockface_available)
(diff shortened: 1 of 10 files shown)


src/nyc_trees/apps/survey/urls/census_admin.py
```diff
@@ -18,6 +20,10 @@
 # These URLs have the prefix 'census_admin/'
 urlpatterns = patterns(
     '',
+    url(r'^territory/$',
+        census_admin_do(route(GET=r.admin_territory_page)),
```
kdeloach: Consider moving to routes.py.

kdeloach:
+1 I was able to switch between groups on the census admin page and click that groups' blockfaces on the map.

src/nyc_trees/apps/survey/urls/census_admin.py
maurizi: I agree with @kdeloach here, we don't want to double decorate things.

Also, we should remove the existing `admin_blockface_page` view/route/url since we're not using them.

maurizi:
As we discussed, I think we need to make some changes to the SQL query for the tile layer.
Let me know when that is done and I'll give it a test drive.

RickMohr:
@maurizi, @kdeloach, Thanks for the comments. Updated.

maurizi:
+1, works

maurizi:
(Don't forget to squash your commits)

PR #708, opened 2015-03-02
Add expert_required to survey_blockface

jwalgran:
The block faces we will be importing include a flag that indicates when the block face may be too difficult to be mapped accurately by the general public. We will not be using this flag at "runtime" but adding it to our blockface model makes it easy to bulk-assign these blocks as territory owned by the NYC Parks group.


RickMohr:
+1

maurizi:
+1

PR #711, opened 2015-03-02
Fix bounds and location in map module

steventlamb:
Prior to this fix, bounds and location were failing for maps that didn't
have the standard dom id of 'map'.


maurizi:
+1

maurizi:
@RickMohr I'm betting @slamb is going to need to use `domId` for `addTileLayer` and `addGridLayer` for the survey page's preview map, so perhaps it would be better to change your PR to leave those in.

RickMohr:
@maurizi, I don't think the preview page will have either a tile layer or a grid layer.

maurizi:
@RickMohr Of course...

PR #713, opened 2015-03-02
Show geolocation button on the survey page.

maurizi:
Should be very useful when mapping blocks at an event.


maurizi:
This came up in a conversation between myself and @RickMohr as something that would be very useful on the initial survey map.

jwalgran:
+1

PR #718, opened 2015-03-03
Add a popup for exiting a survey without mapping

jwalgran:
This change set allows a mapper to "bail out" of a survey, and forces them to leave an explanation as to why they are quitting. I have opted to not post any tree data because it could be invalid, and it could be frustrating to a person who is trying to stop mapping if we force them to make data corrections first.

What happens after the quit has been POSTed is left as a TODO because that part of the workflow is in progress.

To test:

1. Log in as an individual mapper
1. Reserve a block face
1. Start mapping, select the block, end, and side of the street.
1. Scroll to the bottom of the survey form and click "I cannot map this block"

There should be a popup with a text area. You should not be able to click "Stop Mapping" until there is non-whitespace content in the text area.

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/survey.html
+++ b/src/nyc_trees/apps/survey/templates/survey/survey.html
@@ -36,7 +36,7 @@
         </div>
 
         <button class="btn btn-primary" id="another-tree">+ Another Tree</button>
-        <button class="btn btn-secondary" id="cant-map">I cannot map this block</button>
+        <a class="btn btn-secondary" id="cant-map" href="#quit-popup" data-toggle="modal">I cannot map this block</a>
 
         <form autocomplete="off">
             <!-- We need a clickable submit button in order to trigger browser error validation popups -->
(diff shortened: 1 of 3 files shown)


src/nyc_trees/apps/survey/templates/survey/survey.html
```diff
@@ -52,6 +52,29 @@
...
+                            <textarea id="quit-reason" style="width: 100%;"
+                                      placeholder="Offset tree bed, bad sidewalks, etc."></textarea>
+                        </form>
+                    </div>
+                    <div class="modal-footer">
+                        <button id="quit" type="submit" class="btn btn-primary">Stop Mapping</button>
+                    </div>
+            </div>
+        </div>
+    </div>
```
maurizi: To prevent frustrating styling issues, consider moving this modal into an `extra_content` block.

jwalgran: It is certainly simple to do that, so I will. I would like to explain why
it should be there. What issues does it avoid?

src/nyc_trees/js/src/lib/valIsEmpty.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+var $ = require('jquery');
+
+module.exports = function valIsEmpty(selector) {
+    return !$.trim($(selector).val());
```
steventlamb: hooray for javascript. The empty string is falsy, but the empty array isn't.

src/nyc_trees/js/src/surveyPage.js
```diff
@@ -241,6 +247,16 @@ function getTreeData(i, form) {
 
 $(dom.submitSurvey).on('click', submitSurveyWithTrees);
 
+function createSurveyData() {
+    return {
+        survey: {
+            blockface_id: blockfaceId,
+            is_left_side: $(dom.leftButton).is('active'),
+            is_mapped_in_blockface_polyline_direction: isMappedFromStartOfLine
```
steventlamb: should this have
```
quit_reason: null,
trees: null,
```
for documentation?

src/nyc_trees/apps/survey/templates/survey/survey.html
maurizi: `extra_content` is a direct child of `<body>`, which makes certain that styling changes to `base.html` don't push this modal inside of an element with a low z-index, causing this modal to appear below other elements.

src/nyc_trees/js/src/surveyPage.js
```diff
@@ -288,3 +298,55 @@ function submitSurveyWithTrees() {
...
+    // Bootstrap will handle showing the modal, but the
+    // hidden class, applied by other code, needs to be
+    // removed.
+    $(dom.quitPopup).removeClass('hidden');
+});
+
+// The quit button is not enabled until a quit reason is entered
+$(dom.quit).prop('disabled', true);
+
+$(dom.quitReason).on('change keyup paste', function() {
```
steventlamb: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/group_list.js#L28

I think I had some success with `on('input', ...)`.

steventlamb:
+1, works as described

src/nyc_trees/js/src/surveyPage.js
jwalgran: The ``'input'`` event does not have the same behavior in a ``textarea``, which is why I ended up with this collection of events suggested after googling the problem.

src/nyc_trees/js/src/surveyPage.js
jwalgran: added

src/nyc_trees/apps/survey/templates/survey/survey.html
jwalgran: moved

src/nyc_trees/js/src/lib/valIsEmpty.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
```
kdeloach: Consider consolidating with `utils.js`.

PR #719, opened 2015-03-03
Selectable layer for unmapped group territory

RickMohr:
* Endpoint `group_unmapped_territory_geojson`
* Click blockfaces to select/deselect individually
* Explain unselectable blockfaces via toast

Also:
* Sort groups by name in drop-down
* Show the first group's territory on page load


maurizi:
+1

PR #720, opened 2015-03-03
Remove is_census_admin in favor of using builtin is_superuser.

### LINKED ISSUE #75: Should we use the Django permissions system and admin flags?
We may need to create `Permission` objects if we want to use the Django Admin site.

We'll need to set either `is_staff` or `is_superuser` to allow Census admins access to the Django Admin site.  We could use one of them to indicate whether a user is a census administrator and eliminate `is_census_admin`, to avoid the two flags from getting out of sync.

maurizi:
Since Census Admins will need to have access to the Django Admin
site anyways, it makes sense to consolidate these flags into 1.

Fixes #75

### DIFF
--- a/src/nyc_trees/apps/core/helpers.py
+++ b/src/nyc_trees/apps/core/helpers.py
@@ -7,11 +7,11 @@ from apps.users.models import TrustedMapper
 
 
 def user_is_census_admin(user):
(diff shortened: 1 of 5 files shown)


kdeloach:
+1

PR #722, opened 2015-03-04
Add teammate selection to survey

jwalgran:
The teammate list is implemented as a select tag, transformed into an autocomplete by [Select2](http://select2.github.io/select2/)

I pulled in the most recent stable version of Select2
https://github.com/select2/select2/archive/3.5.2.zip
select2/select2@6633cda

The teammate list is populated with event attendees when surveying at a mapping event, and all individual mappers when surveying an individual reservation.

The list respects privacy settings and shows full names only when allowed. 

Adding another set of fields to the existing blockface selection workflow made the hide and show logic a little too messy so I added a helper that builds functions to show and hide mutually exclusive field groups.

Having the Select2 box in a section pinned to the bottom of the page does not produce the best behavior on a small mobile screen with the soft keyboard visible. This is a workable implementation, but will need some style improvement.

To test you should ideally have 2 individual mapper users and an in progress event with 3 RSVPed users (at least one of which is not an individual mapper).

* Test surveying a reservation and verify that the teammate list only contains individual mappers.
* Test surveying at an event and verify that the teammate list contains all checked in users.
* Test that the teammate selection can be cleared by clicking the "X" in the dropdown
* Test that surveys can be saved successfully both with and without teammates.


src/nyc_trees/apps/survey/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+    if can_show_full_name(user):
+        full_name = ("%s %s" % (user.first_name, user.last_name)).strip()
+    else:
+        full_name = ""
+    search_index = user.username.lower() + "".join(full_name.split()).lower()
+    return {
+        "user_id": user.id,
+        "username": user.username,
+        "full_name": full_name,
+        "search_index": search_index
```
maurizi: This `search_index` doesn't seem to be used anywhere.  What is it for?

maurizi:
When testing this out there was a 404 in the developer tools for `http://localhost:8000/static/css/select2-spinner.gif`

maurizi:
Other than the 2 issues I raised, this works pretty well.  We'll definitely need some styling work for mobile, but we can tackle that later.

+1

src/nyc_trees/apps/survey/helpers.py
jwalgran: removed

jwalgran:
I added the spinner gif.

PR #725, opened 2015-03-04
Add psql helper script

steventlamb:


hectcastro:
Worth noting that you can also run `psql` from your laptop and connect to the PostgreSQL instance inside the services VM (via port `localhost:15432`). Alternatively, you can use `pgweb` too.

:+1:

steventlamb:
Ah, I didn't realize the `15432` route was available, I'll give that a go before merging. That would really make this case moot.

steventlamb:
I no longer have a need for this because we have reasonable support for connecting with a local installation of psql across the network.

PR #728, opened 2015-03-04
Add IP address range for eth0 to PostgreSQL HBA

hectcastro:
This changeset allows Vagrant users to connect to the PostgreSQL database from the virtual machine host via the port forwarded `15432`.


jwalgran:
+1 I added this line manually to my pg_hba.conf, did a ``sudo service postgresql reload`` and was able to psql into my vm database.

PR #739, opened 2015-03-05
Feature/mw/treecorder styling

designmatty:
Styled and reorganized the tree survey page to match the look and feel across the site.


src/nyc_trees/sass/partials/_survey.scss
```diff
@@ -1,12 +1,138 @@
...
+}
+
 .tree-form {
+    label {
+        color: $gray;
+    }
     // Hide checkboxes and radios, and use their checked/unchecked to style their label
     input[type="checkbox"], input[type="radio"] {
         opacity: 0;
+        display: none;
```
maurizi: The JS for the treecorder only submits "visible" elements (anything that returns true for jQuery's `.is(":visible")`), which is why I gave this `opacity: 0`.  Setting this to `display: none` breaks actually submitting the form.

Unfortunately I can't think of a way to make jQuery think this element is "visible" without it messing up the layout.
Mind if I push a commit to this branch that will fix the JS, so we can keep the `display: none`?

maurizi:
I like being able to see the survey and the map at the same time on desktops.

However, a couple of times I tried to scroll the survey and instead I accidentally moved the map.  Should we make the map not interactive when the form is open?

src/nyc_trees/sass/partials/_survey.scss
maurizi: I pushed a commit that fixes this.

maurizi:
There are scrollbars around the survey are always visible on desktop browsers, even when there is no need for them (which is always true for the horizontal scroll)
screenshot from 2015-03-05 12 08 20

designmatty:
I'll look into these scroll bars. They weren't visible on osx which is probably why I missed them.

designmatty:
The scrollbars should no longer appear unless an actual overflow exists.

maurizi:
+1.  Let me know if you need help fixing any merge conflicts.

designmatty:
I should be fine with this but if I run into any issues I'll be sure to let you know. Thanks

PR #743, opened 2015-03-05
Update blockface fixture and add scripts for blockface management

jwalgran:
We are receiving the production blockface data in chunks. I have updated the fixture with the production Manhattan blockfaces.

``dump_blockfaces.sh`` serves to document the process of dumping the content of the survey_blockface table into a known file which is loaded by setupdb.sh

Parks wants to assign all the "difficult" blocks to a group so that only skilled users can reserve and map them. This command idempotently assigns any blockface with ``expert_required == True`` to a specified group, possibly deleting a previous assignment to another group.

screen shot 2015-03-05 at 9 45 19 am

### DIFF
--- a/scripts/dump_blockfaces.sh
+++ b/scripts/dump_blockfaces.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -e
+set -x
+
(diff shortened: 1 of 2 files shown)


RickMohr:
+1 Looks good. I was able to import the data and assign the expert blocks.

PR #744, opened 2015-03-05
Increase header z-index to appear over leaflet attribution

jfrankl:
image

Fixes #744


jwalgran:
+1 no more floaty attributions

PR #746, opened 2015-03-05
Fix survey tree fields

maurizi:


RickMohr:
Looks good...
But Parks wants us to save diameter separately rather than converting, adding a nullable `stump_diameter` field and making `circumference` nullable.

maurizi:
@RickMohr alright, I added a new `stump_diameter` field

RickMohr:
+1 thanks for the thorough validation.

PR #751, opened 2015-03-05
Topic/render survey page

steventlamb:


src/nyc_trees/js/src/surveyPage.js
```diff
@@ -379,7 +379,7 @@ function submitSurveyWithTrees() {
         })
         .done(function(content) {
             window.alert('Successfully saved survey');
-            // TODO: Go to preview page
+            window.location.href = '/survey/detail/' + content.survey_id + '/';
```
jwalgran: If you add an ``hasUnsavedData = false;`` before setting the href it will suppress the "are you sure you want to leave the page" message.

jwalgran:
I made myself a little survey:

screen shot 2015-03-05 at 3 19 32 pm

+1 after we suppress the alerts.

PR #758, opened 2015-03-06
Topic/survey detail controls

steventlamb:


jwalgran:
I think "No -Start Over" should include the blockface ID in the URL hash so the same block is selected on reload.

jwalgran:
If they say "No - Submit for review" we need to hit the ``flag_survey`` endpoint before asking what they would like to do next.

steventlamb:
@jwalgran how's that look?

jwalgran:
+1 all three options work perfectly

PR #763, opened 2015-03-06
Style reservations instructions page

jfrankl:
Fixes #613

PR #767, opened 2015-03-06
Update reservations instructions view

kdeloach:
This adds some logic to make steps inactive until you are able to start
that step.

I figured out a simpler way to accomplish this after giving it some thought. Please test to see if this behavior works as expected.


kdeloach:
Ref https://github.com/azavea/nyc-trees/pull/763.

PR #769, opened 2015-03-06
Allow saving updated group territory

RickMohr:
* "Save" button posts currently-selected blockface ids to `group_update_territory` endpoint,
  which adds/removes appropriate `Territory` records.
* We need new tiler URLs after saving. So, change to return full tiler URLs along with
  Ajax data instead of stashing them on the map object.
* "Are you sure?" confirmation if user reverts, changes group, or leaves page.
* Disable controls (save, revert, change group, click blockface) during Ajax requests
* Refactor and simplify `adminTerritoryPage.js`


maurizi:
I saw some issues with Leaflet draw on Chromium on Linux.  I don't think that should block this PR, but we should open an issue for it.

maurizi:
Otherwise looks good, +1

PR #774, opened 2015-03-06
Update migration documentation to including automated deployment

hectcastro:
This update to the migrations documentation includes steps for disabling and enabling deployments before and after applying migrations.


jwalgran:
+1 nice addition.

PR #780, opened 2015-03-07
Prevent selecting unavailable blocks on survey

### LINKED ISSUE #776: Block can be remapped after submitting data as accurate
I reserved a block, filled out a survey for the block, selected that yes the data was accurate, and then selected that I wanted to map another block. 

I was returned to the map page, and the block was shaded as unavailable. However, I was able to select the same block again and map it. The block should no longer be selectable to map.

jwalgran:
In a previous commit I changed the color of the block but forgot to extend the filtering to the UTF grid.

Fixes #776


maurizi:
+1

Issue #777, opened 2015-03-07
Species list not displaying in Treecorder


dboyer:
See screenshot. This may be the same as #756 although it looks like there's a button partially displayed as well. I see that #771 is meant to fix that but it's been merged and I'm still seeing the problem.

dboyer:
![screenshot_2015-03-06-21-37-04](https://cloud.githubusercontent.com/assets/516764/6539714/8585e624-c449-11e4-9f69-a4884f539b71.png)

PR #781, opened 2015-03-07
Override default plugin style

### LINKED ISSUE #756: List of teammates doesn't autocomplete
Perhaps this is related to design changes, but when I try to enter a teammate, I see a blue line on the left side indicating there are autocomplete options but I can't see what those options are. This happened on both desktop and phone so it's not related to screensize.

jfrankl:
Fixes #756 and #777


jwalgran:
+1

PR #785, opened 2015-03-09
Add docstring comments to popular scripts

steventlamb:
For parsing with third-party script execution tools

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 
+# Run flake8 against the Django codebase
+
 set -e
 set -x
 
-# Run flake8 against the Django codebase and output a known string so that
-# the Jenkins text finder plugin can detect a failed check and mark the build
-# unstable. This command should only fail the build if the `vagrant ssh`
-# command itself fails.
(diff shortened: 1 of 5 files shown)


scripts/check.sh
```diff
@@ -1,10 +1,8 @@
 #!/bin/bash
 
+# Run flake8 against the Django codebase
+
 set -e
 set -x
 
-# Run flake8 against the Django codebase and output a known string so that
```
jwalgran: I don't want to loose the comment explaining why ``|| echo flake8`` is there. We could breakup the comment and move the details below the ``set -x`` to clean up the help output.

steventlamb: I think that that comment is only important in `test.sh`, where it was copied from. I think `check.sh` is just being used by developers for a quick lint.

steventlamb: In other words, when I made `check.sh`, I should have deleted that comment.

jwalgran: You are totally right. In fact, the ``|| echo flake8...`` shoud be removed entirely (but that is out of the scope of this change)

jwalgran:
+1

Issue #784, opened 2015-03-09
Treecorder - validation no longer works for checkboxes or radio buttons


maurizi:
This is because the checkbox elements are hidden, which prevents the browser from validating them.

maurizi:
Closed by #789

PR #789, opened 2015-03-09
fixes form validation due to hidden checkboxes and radios #784

designmatty:


maurizi:
+1

src/nyc_trees/sass/partials/_buttons.scss
```diff
@@ -5,6 +5,55 @@
...
+
+  .btn {
+    display: table-cell;
+    float: none;
+    vertical-align: middle;
+    padding: 8px;
+
+    // This fixes the white space between buttons caused by the checkbox/radio btns
+    // Bad Practice
+    // Look into a better solution for this fix
```
maurizi: Since we missed the boat on getting this fixed in time for user testing, it might be best to clean this up now.

Issue #788, opened 2015-03-09
Can't scroll treecorder form on non-Retina iPad


jwalgran:
Movie from iOS simulator in landscape orientation. Confirmed on a v1 iPad mini

http://cl.ly/0i2A1i2L3Z3T/2015_03_09_treecorder_ipad_cant_scroll.mov

Issue #787, opened 2015-03-09
Treecorder on non-retina iPad does not show selection prompts


jwalgran:
Screenshot from iOS simulator. Verified on v1 iPad mini.

![screen shot 2015-03-09 at 9 15 38 am](https://cloud.githubusercontent.com/assets/17363/6558998/05964828-c63d-11e4-88db-4a5d20672b1c.png)

jwalgran:
Rotating to landscape orientation shows the content

![screen shot 2015-03-09 at 9 18 05 am](https://cloud.githubusercontent.com/assets/17363/6559022/3a620984-c63d-11e4-8363-775e9297db47.png)

designmatty:
This might require adjusting the media queries. I'll take a look

PR #792, opened 2015-03-09
fixed media query and scroll issues for #787 and #788

designmatty:
I was able to fix two issues with this. 

The media queries were hiding the map-sidebar on 767px width only.
There was an overflow-x: hidden; which for some reason was disabling all scrolling functionality for y as well.


jwalgran:
Taking a look at this now

jwalgran:
+1 both scrolling and content visibility issues fixed

PR #793, opened 2015-03-09
Add events feed endpoint

steventlamb:


src/nyc_trees/apps/event/views.py
```diff
@@ -166,8 +166,38 @@ def events_list_page_partial(request):
...
+            "address": e.address,
+            # NOT POSSIBLE WITHOUT REVERSE GEOCODING:
+            # "zip": None,
+            # WHERE WOULD WE GET THESE? PROBABLY NEED TO DROP:
+            # "park_id": None,
+            # "borough": None,
+            # "accessible": None,
+        }],
+        # EVENT HAS NO LINKS. WALK UP TO event.group.contact_url ?:
+        # 'links': [],
```
maurizi: I think we should link to the URL for the event detail here.

src/nyc_trees/apps/event/views.py
```diff
@@ -166,8 +166,38 @@ def events_list_page_partial(request):
...
+        # NOT INCLUDED IN THIS APP. DROPPED:
+        # 'phone': None,
+        # 'cost_free': None,
+        # 'cost_description': None,
+        # 'categories': [],
+    } for e in Event.objects.order_by('-begins_at')]
```
maurizi: I think we should exclude private events from this list.

dboyer:
Yes, we should exclude private events.

maurizi:
I think we should change the URL for this view to end in `.json`

maurizi:
+1, after fixing my the comment about the URL

PR #795, opened 2015-03-09
updated the requested text changes and fixed #778

### LINKED ISSUE #778: Text changes on Treecorder
1. Tree Bed is missing from the tree location question. The original variable list has "Tree bed" in front of the "along the curb" and "offset from the curb" buttons. If there's no room on the buttons, tree bed needs to go in the header. It could change to "Location of Tree Bed."

2. Tree Species Confidence can be relabeled as "Are you confident in this answer?"

3. Evidence of Stewardship should change to "How many stewardship practices do you see?"

designmatty:


maurizi:
+1

PR #796, opened 2015-03-09
Disable map activity during survey form

### LINKED ISSUE #766: Disable survey map interactions while filling out tree details
When using a large screen size, the survey form is shown over a fullscreen map. It is possible to interact with the map by accident, which could lead to applying a survey to the wrong block, or in the wrong direction. 

screen shot 2015-03-06 at 11 40 23 am

steventlamb:
fixes #766 on github


jwalgran:
Testing now

jwalgran:
+1 I verified that I could not click on another blockface or click the other "end circle" while the survey was in process.

PR #799, opened 2015-03-10
Add group affiliation field and use it for searching on groups page

### LINKED ISSUE #234: Need Affiliation field on the groups model
There should be a field called Affiliation that is an open text field. The group admin can enter the name of a larger group with which their group is affiliated (i.e. Brooklyn Maple Lovers enters Maple Lovers of Greater New York in the affiliation field). The affiliation field will be searchable from the groups page.

This is a new item that may not be represented on the wireframes.

maurizi:
Fixes #234

### DIFF
--- a/src/nyc_trees/js/src/group_list.js
+++ b/src/nyc_trees/js/src/group_list.js
@@ -2,12 +2,17 @@
 
 var $ = require('jquery'),
     fetchAndReplace = require('./fetchAndReplace'),
-    attrs = {name: 'data-group-name',
-             description: 'data-group-description'},
-    dom = {searchInput: '[data-class="group-search"]',
-           viewAll: '[data-class="group-view-all"]',
-           searchNoneFound: '[data-class="group-search-none-found"]',
-           groups: '[' + attrs.name + ']'},
+    attrs = {
+        name: 'data-group-name',
+        description: 'data-group-description',
+        affiliation: 'data-affiliation'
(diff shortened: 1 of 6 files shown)


src/nyc_trees/js/src/group_list.js
```diff
@@ -2,12 +2,17 @@
...
-    attrs = {name: 'data-group-name',
-             description: 'data-group-description'},
-    dom = {searchInput: '[data-class="group-search"]',
-           viewAll: '[data-class="group-view-all"]',
-           searchNoneFound: '[data-class="group-search-none-found"]',
-           groups: '[' + attrs.name + ']'},
+    attrs = {
+        name: 'data-group-name',
+        description: 'data-group-description',
+        affiliation: 'data-affiliation'
```
steventlamb: this should be `data-group-affiliation` for consistency.

src/nyc_trees/js/src/group_list.js
```diff
@@ -26,7 +31,7 @@ $(dom.viewAll).on('click', function () {
 });
 
 $(dom.searchInput).on('input', function (e) {
-    var query = $(e.target).val(),
+    var query = $(e.target).val().toLowerCase(),
```
steventlamb: this is not necessary because the template tags call `.lower`

steventlamb:
+1

PR #803, opened 2015-03-10
Redirect to previous page when clicking on the login link.

### LINKED ISSUE #649: Add the current page as the ?next= parameter of the "Login" link
I think this flow 

Search Events->Event 1->Login -> Event 1

Is probably better than this one

Search Events->Event 1->Login->Profile->"Wait, how do I get back to that event page?"

maurizi:
Fixes #649


jwalgran:
+1 verified

PR #804, opened 2015-03-10
Fix event attendance counts.  Fixes #251

### LINKED ISSUE #251: Fix event attendance counts on Group Detail
When a single user attends 2 events held by the same group, should that count as 1 "event attendee" or 2?

maurizi:


kdeloach:
+1 Tests pass.

PR #806, opened 2015-03-10
Blur the Select2 element when it is closed.

maurizi:
Needed for better UX on mobile devices with soft keyboards


designmatty:
+1 works perfectly.

PR #810, opened 2015-03-10
Add support for a Redis backed cache

### LINKED ISSUE #169: Select a cache for the production Django settings
Currently set to `django.core.cache.backends.locmem.LocMemCache`. This is not desirable. Candidates are `memcached` and Redis.

hectcastro:
This changeset adds a Redis backed cache to all environments other than testing. Non-default settings include:

- The Hiredis parser is being used
- Socket timeout is explicitly set to 3
- Connection failure exceptions are suppressed
- The Redis database at index 1 is being used

Attempts to resolve #169.

### DIFF
--- a/deployment/ansible/roles.txt
+++ b/deployment/ansible/roles.txt
@@ -1,4 +1,4 @@
-azavea.apache2,0.2.1
+azavea.apache2,0.2.2
 azavea.ntp,0.1.0
 azavea.pip,0.1.0
 azavea.nodejs,0.3.0
--- a/src/nyc_trees/nyc_trees/settings/base.py
+++ b/src/nyc_trees/nyc_trees/settings/base.py
@@ -40,6 +40,7 @@ STATSD_PREFIX = 'django'
 STATSD_HOST = environ.get('NYC_TREES_STATSD_HOST', 'localhost')
 # END STATSD CONFIGURATION
 
+
 # EMAIL CONFIGURATION
 DEFAULT_FROM_EMAIL = 'treescount.help@parks.nyc.gov'
 # END EMAIL CONFIGURATION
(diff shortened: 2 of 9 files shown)


deployment/ansible/roles.txt
```diff
@@ -1,4 +1,4 @@
-azavea.apache2,0.2.1
+azavea.apache2,0.2.2
```
hectcastro: Just realized this doesn't have anything to do with this PR. The version bump adds a version glob for Apache.

src/nyc_trees/nyc_trees/settings/base.py
```diff
@@ -40,10 +40,33 @@
...
+CACHES = {
+    'default': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        # The Redis database at index 0 is used by Logstash/Beaver
+        'LOCATION': 'redis://{0}:{1}/1'.format(
+            environ.get('NYC_TREES_CACHE_HOST', 'localhost'),
+            environ.get('NYC_TREES_CACHE_PORT', 6379)),
+        'OPTIONS': {
+            'PARSER_CLASS': 'redis.connection.HiredisParser',
+            'SOCKET_TIMEOUT': 3,
```
jwalgran: Is "3" a best practice for socket timeout?

src/nyc_trees/nyc_trees/settings/base.py
```diff
@@ -40,10 +40,33 @@
...
+        }
+    }
+}
+
+# Don't throw exceptions if Redis is down.
+DJANGO_REDIS_IGNORE_EXCEPTIONS = True
```
jwalgran: Are we going to lean on CloudWatch and or logstash to keep an eye in Redis?

hectcastro: Plan is to depend on Watchman to detect that connections from the application server to the cache  aren't working, which fires off a Pingdom alert.

src/nyc_trees/nyc_trees/settings/base.py
hectcastro: That was the number we settled on for BCG when the application was blowing up because it couldn't talk to Redis. Default is to hang, which I want to avoid.

jwalgran:
+1

I provisioned my VMs and ran ``redis-cli monitor`` on the ``services`` VM to watch the traffic. I then edited  the group list template and wrapped the group listing with ``{% cache 5 grouplist %}``. I saw the template fragment get added to redis, saw immediate refreshes hit the cache, and then saw the template fragment get re-cached when the page was reloaded after 5 seconds.

PR #814, opened 2015-03-10
Limit map another block prompt

### LINKED ISSUE #802: User can be in Treecorder without having reservations.
If a user surveys the last block they have reserved and selects "Yes" to mapping more, they are taken back to the first survey page where they are told to select a block to map. No blocks are selectable since they have no reservations. They should probably go to the reservations page instead with a message about needing to reserve more blocks.

steventlamb:
fixes #802

Only show map another block prompt when there are remaining reservations.

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/partials/map_another_popup.html
+++ b/src/nyc_trees/apps/survey/templates/survey/partials/map_another_popup.html
@@ -2,15 +2,15 @@
     <div class="modal-dialog">
         <div class="modal-content">
             <div class="modal-body">
-                <div>Map another block?</div>
+                <div class="hidden" data-class="map-another-possible">Map another block?</div>
+                <div class="hidden" data-class="map-another-impossible">You've mapped all your reserved blocks!</div>
             </div>
             <div class="modal-footer">
-                <a id="map-another-no" href="{% url 'progress_page' %}" class="btn btn-default">
-                    I'm Done For Now
+                <a class="btn btn-default" href="{% url 'progress_page' %}">
+                  <span data-class="map-another-possible">I'm Done For Now</span>
+                  <span data-class="map-another-impossible">OK</span>
                 </a>
-                <a id="map-another-yes" href="{% url 'survey' %}" class="btn btn-primary">
-                    Yes!
-                </a>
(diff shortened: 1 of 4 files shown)


src/nyc_trees/apps/survey/templates/survey/partials/map_another_popup.html
```diff
@@ -2,15 +2,23 @@
...
-                <a id="map-another-yes" href="{% url 'survey' %}" class="btn btn-primary">
-                    Yes!
+                <a id="map-another-no"
+                   class="btn btn-default"
+                   data-class="map-another-possible"
+                   href="{% url 'progress_page' %}">I'm Done For Now</span>
+                <a id="map-another-no"
+                   class="btn btn-default"
+                   data-class="map-another-impossible"
+                   href="{% url 'progress_page' %}">OK</span>
```
steventlamb: @dboyer @jfrankl  This needs content review. How does this look?

screenshot from 2015-03-10 18 59 26

src/nyc_trees/apps/survey/templates/survey/partials/map_another_popup.html
```diff
@@ -2,15 +2,23 @@
...
-                <a id="map-another-no" href="{% url 'progress_page' %}" class="btn btn-default">
-                    I'm Done For Now
-                </a>
-                <a id="map-another-yes" href="{% url 'survey' %}" class="btn btn-primary">
-                    Yes!
+                <a id="map-another-no"
+                   class="btn btn-default"
+                   data-class="map-another-possible"
+                   href="{% url 'progress_page' %}">I'm Done For Now</span>
+                <a id="map-another-no"
```
maurizi: This id is the same as the element above.  We shouldn't have duplicate ids

maurizi:
I don't see anything in the view code that differentiates between surveys for events and surveys for reservations.

We shouldn't be checking remaining reservations when the survey was submitted for an event.

steventlamb:
@maurizi ready for another look.

maurizi:
@azavea-bot rebuild

dboyer:
Let's add a note: You've mapped all your reserved blocks!

PR #819, opened 2015-03-11
Set PostgreSQL log_min_duration_statement setting

hectcastro:
This changeset sets the `log_min_duration_statement` to `500` milliseconds within the PostgreSQL instance on Vagrant and RDS.

### DIFF
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -7,6 +7,7 @@ postgresql_listen_addresses: "*"
 postgresql_username: nyc_trees
 postgresql_password: nyc_trees
 postgresql_database: nyc_trees
+postgresql_log_min_duration_statement: 500
 postgresql_hba_mapping:
   - { type: "host", database: "all", user: "all", address: "33.33.33.1/24", method: "md5" }
   - { type: "host", database: "all", user: "all", address: "10.0.2.0/24", method: "md5" }
--- a/deployment/troposphere/data_store_template.py
+++ b/deployment/troposphere/data_store_template.py
@@ -195,6 +195,15 @@ database_server_subnet_group = t.add_resource(rds.DBSubnetGroup(
     Tags=Tags(Name='dbsngDatabaseServer')
 ))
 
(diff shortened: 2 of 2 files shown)


hectcastro:
Example output of an offending query from staging:

```
2015-03-11 18:47:40 UTC:10.0.1.77(33798):nyctrees@nyc_trees:[10269]:LOG: duration: 515.088 ms execute <unnamed>: SELECT ST_AsBinary("geom") AS geom,"geojson","group_id","id","survey_type" FROM -- The 'DISTINCT ON' and 'ORDER BY' clause are needed so that we only use
-- the latest survey for a blockface
(SELECT
DISTINCT ON (block.id)

ST_AsGeoJSON(block.geom) AS geojson,

block.geom, block.id, turf.group_id,
CASE
WHEN (survey.user_id IS NOT DISTINCT FROM 3
OR survey.teammate_id IS NOT DISTINCT FROM 3) THEN 'surveyed-by-me'
WHEN survey.id IS NOT NULL THEN 'surveyed-by-others'
WHEN reservation.user_id IS NOT DISTINCT FROM 3 THEN 'reserved'
WHEN (block.is_available AND reservation.id IS NULL
AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'
ELSE 'unavailable'
END AS survey_type
LEFT OUTER JOIN survey_territory AS turf
ON block.id = turf.blockface_id
LEFT OUTER JOIN survey_survey AS survey
ON block.id = survey.blockface_id
LEFT OUTER JOIN survey_blockfacereservation AS reservation
AND reservation.expires_at > now() at time zone 'utc')
LEFT OUTER JOIN users_trustedmapper AS trustedmapper
ON (turf.group_id = trustedmapper.group_id
AND trustedmapper.user_id = 3
AND trustedmapper.is_approved)
ORDER BY block.id, survey.created_at DESC NULLS LAST
) AS query WHERE "geom" && ST_SetSRID('BOX3D(-76.02539062453434 39.84228602029377,-74.44335937453434 41.046216814071)'::box3d, 4326)
```

maurizi:
+1

PR #820, opened 2015-03-11
Remove device settings from Collectd df plugin

hectcastro:
This changeset removes the `Device` setting from the Collectd plugin because Collectd intelligently collects metrics from all mounted devices automatically.

### DIFF
--- a/deployment/ansible/roles/nyc-trees.collectd/templates/df.conf.j2
+++ b/deployment/ansible/roles/nyc-trees.collectd/templates/df.conf.j2
@@ -1,4 +1,3 @@
 <Plugin df>
-	Device "/dev/sda1"
 	ReportInodes true
 </Plugin>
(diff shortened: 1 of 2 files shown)


hectcastro:
Before, only metrics for `df-root` would come through:

graphite_browser

jwalgran:
I did a ``provision`` and ``reload`` with this branch, and I see the change to my ``df.conf``

```
$ cat /etc/collectd/collectd.conf.d/df.conf 
<Plugin df>
        ReportInodes true
</Plugin>
```

However, my graphite still only shows ``df-root``

screen shot 2015-03-16 at 9 36 14 am

hectcastro:
A change in the `collectd` configuration files should have triggered a restart of the service. If there was a failure during provisioning or something, the restart trigger could have gotten dropped, but a reload of the `app` virtual machine should have picked up the change.

I just provisioned this branch locally and the new file systems came up. Anything uncommon or unexpected when you provisioned? Did you refresh the Graphite UI after provisioning?

jwalgran:
I don't recall and problems. I will try a fresh checkout.

jwalgran:
+1 provisioning from a fresh checkout resulted in a Graphite dashboard with the additional df-* items.

PR #822, opened 2015-03-11
Add updated Amazon RDS certificate bundle to all instances

hectcastro:
This changeset adds the updated Amazon RDS certificate bundle to all instances so that they can securely communicate with RDS after 23 Mar 2015.

See also: https://forums.aws.amazon.com/ann.jspa?annID=2889

### DIFF
--- a/deployment/ansible/roles.txt
+++ b/deployment/ansible/roles.txt
@@ -28,3 +28,4 @@ azavea.swapfile,0.1.0
 azavea.build-essential,0.1.0
 azavea.tasseo,0.1.1
 azavea.phantomjs,0.1.0
+azavea.rds-ca-bundle,0.1.0
--- a/deployment/ansible/roles/nyc-trees.common/meta/main.yml
+++ b/deployment/ansible/roles/nyc-trees.common/meta/main.yml
@@ -1,6 +1,7 @@
 ---
 dependencies:
   - { role: "azavea.ntp" }
+  - { role: "azavea.rds-ca-bundle", when: "['packer'] | is_in(group_names)" }
   - { role: "azavea.git" }
   - { role: "nyc-trees.nginx", when: "['monitoring-servers'] | is_not_in(group_names)" }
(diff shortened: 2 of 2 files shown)


hectcastro:
This is pretty complicated to test beyond removing the `packer` guard and seeing if it installs the certificate bundle cleanly within the Vagrant virtual machines:

```bash
vagrant@app:~$ ls -l /etc/ssl/certs/ | grep rds
lrwxrwxrwx 1 root root     26 Mar 11 19:49 b1d2b355.0 -> rds-combined-ca-bundle.pem
lrwxrwxrwx 1 root root     26 Mar 11 19:49 cbbf81bb.0 -> rds-combined-ca-bundle.pem
lrwxrwxrwx 1 root root     59 Mar 11 19:49 rds-combined-ca-bundle.pem -> /usr/local/share/ca-certificates/rds-combined-ca-bundle.crt
```

To test this more thoroughly, I had to spin up a new RDS database with an updated certificate authority setting of `rds-ca-2015`. I then used `psql` to connect to it from machines with this certificate bundle installed, and ones without it installed:

```bash
$ export PGSSLROOTCERT="/etc/ssl/certs/ca-certificates.crt"
$ export PGSSLMODE="verify-full"
$ psql -h <DB_ENDPOINT> -U <DB_USER>
```

In addition to `verify-full`, I also tested `require` and `verify-ca`. As expected, certificate validation failed on instances without the bundle installed.

The next bit of testing is to make sure Django, Windshaft, and others make use of the certificate bundle. This will probably consist of a update to the staging RDS instance's certificate authority setting with an eye on failures if anything breaks after the bundle is installed.

/cc @notthatbreezy @ddohler @moradology because we probably need to do this for HL too.

RickMohr:
+1 I read the announcement, the ansible role, and this code, and everything looks sensible. Testing strategy sounds good, thanks for writing it out. I did not duplicate your test.

PR #824, opened 2015-03-11
Allow a view to return a PDF version of a webpage

RickMohr:
To test, browse to an event detail page and append "/printable-map/" to the URL
(e.g. `/group/east-side-mappers/event/map-a-maple/printable-map/`).

Currently as a proof of concept the test endpoint returns a PDF showing the progress page.
That page has some problems (image doesn't fill the page; scrollbar shown),
but those problems go away when rendering a map which fills the entire page.

We will need to design print-oriented views for the maps we want to print.


maurizi:
+1, beautifully simple!

PR #827, opened 2015-03-11
Fix incorrect querying of current blockface reservations.

maurizi:


steventlamb:
I think this branchname is misleading

steventlamb:
+1

PR #828, opened 2015-03-11
Collapse tree sections for trees that are done.

### LINKED ISSUE #816: Setup bootstrap Accordion Group for Treecorder trees form
When there is only one tree the Accordion should always be expanded and the user should not be able to collapse. Once a new tree is added the ability to collapse the accordion should be enabled. 
screen shot 2015-03-11 at 11 37 21 am

maurizi:
This requires special handling for showing error messages for sections
which are currently collapsed, and we need to wait until after
the section is open to show error messages

Fixes #816


maurizi:
@jfrankl @designmatty I need some extra Fontello icons added for this.

I used `icon-left-open-big` and `icon-down-open-big`, but 1) we don't *have* `icon-down-open-big`, and 2) the wireframes show this as a triangle, not a chevron.

I'd have added them myself but I can't find the Fontello config file anywhere...

jfrankl:
Requested icon added here: #829

maurizi:
@jfrankl Thanks for the help!

RickMohr:
+1 works nice.

Open sections are not collapsed on submit. It appears that is not as intended, but it's debatable which way it should work.

PR #836, opened 2015-03-12
Send all Django logs to the console

hectcastro:
This changeset overrides the default Django log settings so that everything funnels through to the console (which ends up going to `/var/log/nyc-trees-app.log`). In development, the log level defaults to `DEBUG`, but `INFO` everywhere else.


maurizi:
+1

jwalgran:
Does this resolve the log rotation issue we saw on the staging server today (2015-03-16)?

hectcastro:
It does not. I am working on a fix for it locally and haven't pushed any code up yet.

PR #861, opened 2015-03-12
Change database storage type to SSD

hectcastro:
This changes the database storage type to SSD (`gp2` for general purpose SSD storage).

Before applying this to staging:

- [x] Take a `pg_dump` of the staging database
- [x] Pull database backup down locally


hectcastro:
This change was successfully applied to the staging database. Evidence exists in the CloudFormation and RDS tabs.

PR #862, opened 2015-03-13
removed unnecessary classes - fixed 1px gap in btn-groups - fixes #832

### LINKED ISSUE #832: Styling issue on treecorder - 1 pixel gap for 2 button radio group
pixel-off

Not a problem for the 3 or 4 groups

designmatty:
Apparently by adding a width of 100% the pixel gap is removed and the buttons still retain appropriate width. This is expected behavior, but was not working initially when first working through the issue.

### DIFF
--- a/src/nyc_trees/sass/partials/_buttons.scss
+++ b/src/nyc_trees/sass/partials/_buttons.scss
@@ -5,22 +5,20 @@
   }
 }
 
-@for $i from 2 through 4 {
-  .btn-group-#{$i}, .btn-group-toggles {
-    vertical-align: middle;
-    display: table;
-    width: 100%;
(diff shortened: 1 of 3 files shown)


src/nyc_trees/sass/partials/_buttons.scss
```diff
@@ -6,7 +6,7 @@
 }
 
 @for $i from 2 through 4 {
-  .btn-group-#{$i} {
+  .btn-group-#{$i}, .btn-group-toggles {
```
maurizi: I don't see any HTML changes to use this new `btn-group-toggles` class...

maurizi:
+1, works great!

PR #871, opened 2015-03-16
Printable event and reservations maps

RickMohr:
Previously the `printable_event_map` endpoint generated a PDF.
I've removed the PDF generation code for now;
it will be used to save a PDF to S3 when the event location changes.

Now instead `printable_event_map` is a view which draws a Leaflet map:
* Center map on event location and show a marker there
* Show group's territory blockfaces

Likewise `printable_reservations_map` draws a Leaflet map:
* Fit map to bounds of the user's reserved blockfaces
* Show user's reserved blockfaces (but not available blockfaces)


maurizi:
+1

PR #872, opened 2015-03-16
Add Celery support to application server

### LINKED ISSUE #851: Setup Celery infrastructure

hectcastro:
This changeset adds Celery support to the application servers by wiring up a Redis broker and launching a Celery worker on each application server. The workers aren't tied to the application server they colocate with; tasks still flow through a broker that attempts to evenly distribute work.

In addition, Celery signals are configured to emit metrics about job count and duration to Graphite via StatsD/Statsite.

Attempts to resolve #851.

### DIFF
--- a/src/nyc_trees/nyc_trees/settings/base.py
+++ b/src/nyc_trees/nyc_trees/settings/base.py
@@ -45,9 +45,11 @@ STATSD_HOST = environ.get('NYC_TREES_STATSD_HOST', 'localhost')
 DEFAULT_FROM_EMAIL = 'treescount.help@parks.nyc.gov'
 # END EMAIL CONFIGURATION
 
+
 # FILE STORAGE CONFIGURATION
 DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
 # END FILE STORAGE CONFIGURATION
+
 
 # CACHE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#caches
@@ -68,6 +70,19 @@ CACHES = {
 # Don't throw exceptions if Redis is down.
 DJANGO_REDIS_IGNORE_EXCEPTIONS = True
 # END CACHE CONFIGURATION
+
+
+# CELERY CONFIGURATION
+BROKER_URL = 'redis://{0}:{1}/2'.format(
+    environ.get('NYC_TREES_CACHE_HOST', 'localhost'),
+    environ.get('NYC_TREES_CACHE_PORT', 6379))
+
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TASK_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
+
+STATSD_CELERY_SIGNALS = True
+# END CELERY CONFIGURATION
 
 
 # DATABASE CONFIGURATION
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -3,6 +3,8 @@ django_settings_module: "nyc_trees.settings.development"
 
 redis_bind_address: "0.0.0.0"
 
+celery_log_level: "debug"
+
 postgresql_listen_addresses: "*"
 postgresql_username: nyc_trees
 postgresql_password: nyc_trees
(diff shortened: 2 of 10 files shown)


src/nyc_trees/nyc_trees/settings/base.py
```diff
@@ -70,6 +72,19 @@
 # END CACHE CONFIGURATION
 
 
+# CELERY CONFIGURATION
+BROKER_URL = 'redis://{0}:{1}/2'.format(
+    environ.get('NYC_TREES_CACHE_HOST', 'localhost'),
+    environ.get('NYC_TREES_CACHE_PORT', 6379))
+
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TASK_SERIALIZER = 'json'
```
steventlamb: In the past we've always used `pickle` because we don't communicate with other languages and don't accept tasks from untrustworthy sources. It's faster and supports more data types.
http://celery.readthedocs.org/en/latest/userguide/calling.html#serializers

I'm not sure of the reach of what we'll lose, feature-wise, but we should keep pickle serialization in mind if it becomes a problem.

deployment/ansible/roles/nyc-trees.celery/templates/upstart-nyc-trees-celery.conf.j2
```diff
@@ -0,0 +1,14 @@
...
+{% else %}
+start on (local-filesystems and net-device-up IFACE!=lo)
+{% endif %}
+stop on shutdown
+
+respawn
+setuid nyc-trees
+chdir {{ app_home }}
+
+exec envdir /etc/nyc-trees.d/env celery -A "nyc_trees.celery:app" worker -l {{ celery_log_level }} >> {{ celery_log }} 2>&1
```
steventlamb: I was under the impression that the preferred way of running celery with django is to use some form `python manage.py celery`. This might also obviate the need for `nyc_trees/celery.py`.

steventlamb:
+1, this sits well with what I know about deploying celery. I tested the branch and celery is running after provisioning.

One thing to be missed by configuring celery outside of django: it's nice to be able to log into a VM, kill the celery process, and then run `python manage.py celery worker -l info` interactively. We can still do this with `cd /opt/app && envdir /etc/nyc-trees.d/env celery -A "nyc_trees.celery:app" worker -l info` (confirmed working), but it's different than how we do it in OTM2.

src/nyc_trees/nyc_trees/settings/base.py
hectcastro: I settled on JSON because of general recommendations around keeping data being passed around non-opaque (strings, numbers, arrays, lists, dicts). Obviously you can pass those around with Pickle too, but seemed like JSON would help enforce it.

Also, Celery throws a bunch of scary warnings if you start it without a default serializer that seem to want you to avoid Pickle:

```
Starting from version 3.2 Celery will refuse to accept pickle by default.

The pickle serializer is a security concern as it may give attackers
the ability to execute any command.  It's important to secure
your broker from unauthorized access when using pickle, so we think
that enabling pickle should require a deliberate action and not be
the default choice.

If you depend on pickle then you should set a setting to disable this
warning and to be sure that everything will continue working
when you upgrade to Celery 3.2::

    CELERY_ACCEPT_CONTENT = ['pickle', 'json', 'msgpack', 'yaml']

You must only enable the serializers that you will actually use.


  warnings.warn(CDeprecationWarning(W_PICKLE_DEPRECATED))
```

deployment/ansible/roles/nyc-trees.celery/templates/upstart-nyc-trees-celery.conf.j2
hectcastro: The `celery.py` file is a recommendation for [integrating with multiple Django applications](http://celery.readthedocs.org/en/latest/django/first-steps-with-django.html).

hectcastro:
Going to hold on merging this PR until I can extract Celery into its own Ansible role.

hectcastro:
@steventlamb I just added a commit that uses an Ansible role to install and configure Celery.

steventlamb:
@hectcastro tested again with your latest changes, provisions and celery runs. +1

PR #888, opened 2015-03-16
Use S3 for persistent file storage on production.

### LINKED ISSUE #352: Add support for housing image uploads in a central store - Select/write appropriate Django storage backend - Probably not default to s3 storage
It would be great to use the Boto backed Amazon S3 storage support referenced [here](https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html). It would be even better if the plugin works with IAM role credentials (since it is backed by Boto).

maurizi:
The VPC id is pulled from the instance metadata and used as the S3 bucket
name, which ensures that staging and production use different S3 buckets
but have identical settings files.

Since we can set the default file storage separately from the storage used
for static assets (JS, CSS, etc.), I decided to make S3 storage the
default storage mechanism for production.

Fixes #352

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -136,6 +136,10 @@ class User(NycModel, AbstractUser):
             not user_is_trusted_mapper(self, group)
 
 
+def _generate_image_filename(group, filename):
+    return 'group_images/{}/{}'.format(group.slug, filename)
+
+
 class Group(NycModel, models.Model):
     name = models.CharField(max_length=255, unique=True)
     # blank=True is valid for 'slug', because we'll automatically create slugs
@@ -149,7 +153,8 @@ class Group(NycModel, models.Model):
     # delete is allowed.
     admin = models.ForeignKey(User, null=True, blank=True,
                               on_delete=models.PROTECT)
-    image = models.ImageField(null=True, blank=True)
+    image = models.ImageField(null=True, blank=True,
+                              upload_to=_generate_image_filename)
     is_active = models.BooleanField(default=True)
     allows_individual_mappers = models.BooleanField(default=False)
     affiliation = models.CharField(max_length=255, default='', blank=True)
--- a/src/nyc_trees/nyc_trees/settings/production.py
+++ b/src/nyc_trees/nyc_trees/settings/production.py
@@ -2,8 +2,8 @@
 
 
 from os import environ
-from urllib2 import urlopen, URLError
 from dns import resolver, exception
+from boto.utils import get_instance_metadata
 
 from base import *  # NOQA
 
@@ -20,6 +20,13 @@ def get_env_setting(setting):
         error_msg = "Set the %s env variable" % setting
         raise ImproperlyConfigured(error_msg)
 
+
+instance_metadata = get_instance_metadata(timeout=5)
+
+if not instance_metadata:
+    raise ImproperlyConfigured('Unable to access the instance metadata')
+
+
 # HOST CONFIGURATION
 # See: https://docs.djangoproject.com/en/1.5/releases/1.5/#allowed-hosts-required-in-production  # NOQA
 ALLOWED_HOSTS = [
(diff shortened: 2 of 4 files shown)


src/nyc_trees/nyc_trees/settings/production.py
```diff
@@ -82,3 +79,22 @@ def get_env_setting(setting):
...
+
+# Django Storages CONFIGURATION
+try:
+    mac_metadata = get_instance_metadata()['network']['interfaces']['macs']
+    vpc_id = mac_metadata.values()[0]['vpc-id']
+
+    # The VPC id should stay the same for all app servers in a particular
+    # environment and remain the same after a new deploy, but differ between
+    # environments.  This makes it a suitable S3 bucket name
+    AWS_STORAGE_BUCKET_NAME = vpc_id
```
hectcastro: I'd append or prepend something to this that indicates what we're using it for. Seeing a bucket with just a VPC ID name is hard to make sense of. Perhaps removing the `vpc-` from the VPC ID and doing something like `nyc-trees-storage-<ID>` would be better.

src/nyc_trees/nyc_trees/settings/production.py
```diff
@@ -32,11 +33,7 @@ def get_env_setting(setting):
 # ELBs use the instance IP in the Host header and ALLOWED_HOSTS checks against
 # the Host header.
 try:
-    ALLOWED_HOSTS.append(
-        urlopen(
-            'http://instance-data.ec2.internal/latest/meta-data/local-ipv4'
-        ).readline()
-    )
+    ALLOWED_HOSTS.append(get_instance_metadata()['local-ipv4'])
 except URLError:
```
hectcastro: It looks like the underlying function here traps the `URLError` and returns an [empty string](https://github.com/boto/boto/blob/44d873d97bb08a09832b21dcaa0b2d47fe59411b/boto/utils.py#L198-L230).

src/nyc_trees/nyc_trees/settings/production.py
maurizi: Good idea.  I'd prefer to not remove the `vpc-` part though.  How about `django-storages-<vpc-id>`?

src/nyc_trees/nyc_trees/settings/production.py
maurizi: I'll play with this a bit and figure out what will work best.

src/nyc_trees/nyc_trees/settings/production.py
hectcastro: Sure.

maurizi:
@hectcastro I've made the changes you suggested.

src/nyc_trees/apps/core/models.py
```diff
@@ -136,6 +136,10 @@ def eligible_to_become_trusted_mapper(self, group):
             not user_is_trusted_mapper(self, group)
 
 
+def _generate_image_filename(group, filename):
+    return 'group_images/{}/{}'.format(group.slug, filename)
```
hectcastro: `AWS_S3_FILE_OVERWRITE` is `True` by default, which seems to mean that similar file paths would overwrite each other. It might be good to think through the scenarios for generating the paths on S3 to ensure that multiple users can't overwrite each others files.

src/nyc_trees/nyc_trees/settings/production.py
```diff
@@ -82,3 +82,20 @@ def get_env_setting(setting):
...
+
+# Django Storages CONFIGURATION
+mac_metadata = instance_metadata['network']['interfaces']['macs']
+vpc_id = mac_metadata.values()[0]['vpc-id']
+
+# The VPC id should stay the same for all app servers in a particular
+# environment and remain the same after a new deploy, but differ between
+# environments.  This makes it a suitable S3 bucket name
+AWS_STORAGE_BUCKET_NAME = 'django-storages-{}'.format(vpc_id)
+
```
hectcastro: Setting `AWS_QUERYSTRING_AUTH` to `False` should remove all of the additional query string parameters, but [doesn't appear to work](https://github.com/boto/boto/issues/1477) when a security token is used to establish the S3 connection. Workarounds exist, but not sure they're worth chasing.

hectcastro:
:+1:

PR #891, opened 2015-03-17
Add celery debug script

steventlamb:


jwalgran:
I like the style of giving the commands literate names, and then chaining them together

hectcastro:
:+1:

PR #894, opened 2015-03-17
Fix Gunicorn lost file handles caused by logrotate

hectcastro:
`logrotate` runs daily to rotate the application server logs in a way that causes Gunicorn to think that they've been deleted. In order to resolve this issue, after a rotate occurs, an application server restart is triggered to force Gunicorn to reestablish a handle to the file.

### DIFF
--- a/deployment/ansible/roles/nyc-trees.app/templates/logrotate-nyc-trees-app.j2
+++ b/deployment/ansible/roles/nyc-trees.app/templates/logrotate-nyc-trees-app.j2
@@ -3,5 +3,8 @@
   {{ app_log_rotate_interval }}
   compress
   missingok
+  postrotate
+    restart nyc-trees-app
+  endscript
   notifempty
 }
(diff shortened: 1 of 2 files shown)


jwalgran:
The Internet tells me that this is the best way to handle it. I'll force a rotation and check the behavior.

jwalgran:
I am running into permission issues trying force a log rotation

```
logrotate --force /etc/logrotate.d/nyc-trees-app
error: skipping "/var/log/nyc-trees-app.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
```

I found [this page](https://linuxslut.net/logrotate-parent-directory-has-insecure-permissions/) which suggests adding an ``su`` stanza to the logrotate conf. I did that, and the ``--force`` command worked, but my services started running into permission errors trying to write a fresh log file.

Do you have any guidance on how to verify this change?

hectcastro:
I think you are hitting an old Ubuntu [bug](https://bugs.launchpad.net/ubuntu/+source/logrotate/+bug/1278193).

There might be a better way, but I was testing by editing the last line of `/etc/cron.daily` to:

```diff
14c14
< /usr/sbin/logrotate /etc/logrotate.conf
---
> /usr/sbin/logrotate -f -v /etc/logrotate.conf
```

Then running:

```bash
$ sudo /etc/cron.daily/logrotate
```

jwalgran:
+1
I added an exception to the app that I could easily trigger, tailed the log, and saw the errors appear. I then ran the modified  ``/etc/cron.daily/logrotate`` , reran the tail, and saw the errors continue to appear.

PR #896, opened 2015-03-17
cant map block button moved - fix #881. No trees button adjusted - fix #877

### LINKED ISSUE #877: Make the No Trees button more obvious
Several users reported difficulty seeing the No Trees button in the Treecorder. I received the following feedback/suggestions:

1. Needs to be larger
2. Move to the left and place below Treecorder and above Tree 1
3. Place on separate line so doesn't seem like part of the Treecorder header
4. Change text to No Trees on This Block.

### LINKED ISSUE #881: Move "I cannot map this block" button
During Treecorder testing, users reported they accidentally clicked "I cannot map this block" instead of "Submit" at the bottom of the data entry page. The suggested change was to move the button below Submit so that it is the last possible thing the user could click.

designmatty:
...77

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/partials/tree_form.html
+++ b/src/nyc_trees/apps/survey/templates/survey/partials/tree_form.html
@@ -26,7 +26,7 @@
     <div data-status="Stump" class="hidden">
         <fieldset>
             <legend>Stump Size</legend>
-            <div class="stump_diameter">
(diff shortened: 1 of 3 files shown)


designmatty:
@jfrankl I made that change we discussed about moving the no trees button out of the header.

jfrankl:
+1

image

PR #899, opened 2015-03-17
Send email on RSVP

### LINKED ISSUE #856: Send email after RSVP

jwalgran:
When registering for an event the details are emailed to the user.

The text of the email is draft and subject to change and approval.

Fixes #856


src/nyc_trees/apps/mail/templates/mail/rsvp.txt
```diff
@@ -0,0 +1,8 @@
+We have received your RSVP to {{ event.title }}, hosted by {{ event.group.name }}. Here are the details:
+
+When: {{ event.begins_at|date:"l, F jS fA" }} to {{ event.ends_at|date:"fA" }}
+Where: {{ event.address }} — {{ event.location_description }}
```
maurizi: `location_description` is not a required field, so this will look a little odd if it is empty.

Consider changing it to:
```
Where: {{ event.address }}{% if event.location_description %} — {{ event.location_description }}{% endif %}
```

maurizi:
+1, I RSVPed to an event and received an email

PR #900, opened 2015-03-18
Add custom Django storage backend to work around boto issue

### LINKED ISSUE #897: Create wrapper class for S3BotoStorage to remove query string auth
There is a bug in Boto (https://github.com/boto/boto/issues/1477) that prevents disabling query string authentication for S3 URLs if the S3 connection was established with a security token. Disabling query string authentication at the `S3BotoStorage` backend makes URLs 403 because they still include a security token, but no access key. Enabling query string authentication allows things to work, but adds a lot of extra query strings to the URL, and eventually expires.

A proposed workaround to the issue: https://github.com/boto/boto/issues/1477#issuecomment-38759048

maurizi:
Fixes #897

### DIFF
--- a/src/nyc_trees/nyc_trees/settings/production.py
+++ b/src/nyc_trees/nyc_trees/settings/production.py
@@ -83,7 +83,7 @@ vpc_id = mac_metadata.values()[0]['vpc-id']
 AWS_STORAGE_BUCKET_NAME = 'django-storages-{}'.format(vpc_id)
 
 AWS_AUTO_CREATE_BUCKET = True
-DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+DEFAULT_FILE_STORAGE = 'libs.custom_storages.PublicS3BotoStorage'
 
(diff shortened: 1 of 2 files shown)


src/nyc_trees/nyc_trees/settings/production.py
```diff
@@ -83,7 +83,7 @@ def get_env_setting(setting):
 AWS_STORAGE_BUCKET_NAME = 'django-storages-{}'.format(vpc_id)
 
 AWS_AUTO_CREATE_BUCKET = True
-DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+DEFAULT_FILE_STORAGE = 'libs.custom_storages.PublicS3BotoStorage'
```
hectcastro: In testing, I followed @jwalgran's existing convention of placing this custom backend in `apps/core/storage/backends/s3boto.py`.

hectcastro:
Tested this patch on EC2, on top of `ami-8aba98e2`. :+1:

PR #901, opened 2015-03-18
Add missing migrations for changed image field

maurizi:
Commit fda90cf modified the image field of Group, but neglected to inlude
migrations.


RickMohr:
+1

PR #913, opened 2015-03-19
Add a management command for starting a scheduled Celery task

maurizi:
Uses DB locking via `select_for_update` to ensure that the task is only
started once per day.

To test, run migrations, and then run the following command at the same
time in two separate shells:
  'manage.py start_daily_celery_task debug_task'

The task should only be triggered once, which can be verified by looking
in the Celery logs.

### DIFF
--- a/src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
+++ b/src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
@@ -39,8 +39,6 @@ class Command(BaseCommand):
         # which have also selected this table for update.
         task_runs = TaskRun.objects.select_for_update().filter(name=task_name)
 
-        import time; time.sleep(10)
-
         if task_runs.filter(date_started=today).exists():
             self.stdout.write(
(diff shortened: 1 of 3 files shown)


src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
```diff
@@ -0,0 +1,55 @@
...
+                    tasks.keys()))
+
+        task_name = args[0]
+        today = now().date()
+
+        # select_for_update will block if there are other transactions running
+        # which have also selected this table for update.
+        task_runs = TaskRun.objects.select_for_update().filter(name=task_name)
+
+        import time; time.sleep(10)
```
steventlamb: lint

steventlamb:
+1 works and makes sense

steventlamb:
This is limited to day-length locks in a few places to the extent that that's not likely to change. Consider renaming `TaskRun` to `DailyTaskRun`.

Issue #846, opened 2015-03-12
Printable map failure due to PhantomJS exit status


hectcastro:
Relevant traceback from staging:

```
Internal Server Error: /group/harlem-arboreal-society/event/sprint-review/printable-map/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app/apps/core/decorators.py", line 32, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 69, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/opt/app/apps/event/views.py", line 271, in printable_event_map
    pdf_bytes = url_to_pdf(url)
  File "/opt/app/libs/phantomjs.py", line 8, in url_to_pdf
    ['phantomjs', 'js/backend/url2pdf.js', url, str(zoom_factor)])
  File "/usr/lib/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
CalledProcessError: Command '['phantomjs', 'js/backend/url2pdf.js', u'http://localhost/blockface/progress/', '0.75']' returned non-zero exit status 2
```

PR #914, opened 2015-03-19
Generate PDF maps for events 

RickMohr:
Generate PDF maps for events and for user reservations using PhantomJS.

In order to unblock dependencies I'm leaving as a TODO:
Add PhantomJS resource timeout and celery retry logic.

### DIFF
--- a/src/nyc_trees/apps/event/models.py
+++ b/src/nyc_trees/apps/event/models.py
@@ -51,6 +51,8 @@ class Event(NycModel, models.Model):
     is_canceled = models.BooleanField(default=False)
     is_private = models.BooleanField(default=False)
 
+    map_pdf_filename = models.CharField(max_length=255, default='', blank=True)
+
     objects = models.GeoManager()
 
     @staticmethod
--- a/src/nyc_trees/apps/event/views.py
+++ b/src/nyc_trees/apps/event/views.py
@@ -71,7 +71,7 @@ def _process_event_form(form, request, event=None):
         event = form.save()
 
         if needs_pdf_map:
-            create_event_map_pdf(event)
+            create_event_map_pdf(request, event)
 
     return is_valid
 
--- a/src/nyc_trees/apps/event/views.py
+++ b/src/nyc_trees/apps/event/views.py
@@ -28,6 +28,7 @@ from apps.event.helpers import (user_is_rsvped_for_event,
 from apps.mail.views import notify_rsvp
 
 from apps.survey.layer_context import get_context_for_territory_layer
+from libs.pdf_maps import create_event_map_pdf
 
 from libs.data import merge
 
@@ -48,7 +49,7 @@ def add_event(request):
         }
 
 
-def _process_event_form(form, request):
+def _process_event_form(form, request, event=None):
     form.data['group'] = request.group.pk
     try:
         form.data['location'] = Point(float(request.POST['lng']),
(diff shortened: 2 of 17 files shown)


RickMohr:
To test:

1) Make a new event or change the location of an existing event
2) `vagrant ssh app`
3) `cd /var/www/nyc-trees/media`

After 10-15 seconds you should see an `event_map` folder containing a PDF. Copy it to `/opt/app` and view from your computer.

src/nyc_trees/apps/event/models.py
```diff
@@ -51,6 +51,8 @@ class Event(NycModel, models.Model):
     is_canceled = models.BooleanField(default=False)
     is_private = models.BooleanField(default=False)
 
+    map_pdf_filename = models.CharField(max_length=255, default='', blank=True)
```
maurizi: Consider making this be a `FileField` or a `FilePathField`

src/nyc_trees/apps/event/views.py
```diff
@@ -52,9 +53,22 @@ def _process_event_form(form, request):
...
+    if event:
+        event_location = event.location
+
     is_valid = form.is_valid()
+
     if is_valid:
-        form.save()
+        needs_pdf_map = (
+            event is None or
+            not event_location.equals_exact(form.data['location']))
```
maurizi: Should we be checking to see if the event slug has changed here too?

RickMohr: We don't need to because changing the slug doesn't change the map, and the PDF path is already stored. No harm in returning a PDF whose filename contains an old slug.

jwalgran:
When I tested this, I was getting blank PDFs. I traced to the fact that the url being requested included port 8000. You can see in this example that removing the port from the URL produces a non-blank PDF.

```
vagrant@app:/opt/app$ phantomjs js/backend/url2pdf.js http://localhost:8000/group/nyc-parks/event/new-event/printable-map/ 1.0 > /tmp/map.pdf
vagrant@app:/opt/app$ ls -lah /tmp/map.pdf
-rw-r--r-- 1 vagrant nyc-trees 1.3K Mar 19 22:20 /tmp/map.pdf
vagrant@app:/opt/app$ phantomjs js/backend/url2pdf.js http://localhost/group/nyc-parks/event/new-event/printable-map/ 1.0 > /tmp/map.pdf
vagrant@app:/opt/app$ ls -lah /tmp/map.pdf
-rw-r--r-- 1 vagrant nyc-trees 917K Mar 19 22:21 /tmp/map.pdf

RickMohr:
I used `http://localhost:8000` to resolve #846. @hcastro says for production we've configured nginx to reject HTTP traffic, since we expect everything to be HTTPS. But PhantomJS needs HTTP. Hitting 8000 bypasses nginx, and worked for me because I was using the Django server. But with gunicorn we get 404's on requests for CSS files etc. hence the blank PDFs.

jwalgran:
It may be a good I idea to ask JEST about this. I know they use SSL everywhere possible, so their phantom-based solution probably knows how to deal with SSL.

RickMohr:
@jwalgran, @maurizi, please take another look. I've resolved the blank PDF issue and added a second commit to make reservations maps.

RickMohr:
Close/reopen so waffle will link to issues.

maurizi:
+1 (Haven't tested it)

jwalgran:
+1
I ran celery from the console and watched as event creation, location editing, group territory editing, and saving reservations triggered PDF generation. All PDFs were non-blank.

PR #917, opened 2015-03-19
Try to send reservation reminder emails daily

steventlamb:
This changeset introduces a management command to be executed via cron
for emailing users about their blockface reservations that will expire
soon. It has a configurable window of time for sending requests and will
keep trying to send notifications once daily for until it succeeds. Can
only be run once daily.

### DIFF
--- a/src/nyc_trees/apps/mail/templates/mail/reservation_reminder.txt
+++ b/src/nyc_trees/apps/mail/templates/mail/reservation_reminder.txt
@@ -1,0 +1,8 @@
+The following blockface reservations for user '{{ user.username }}' will expire soon:
+{% for reservation in reservations %}
+* 'id: {{ reservation.blockface.id }}' at {{ reservation.expires_at }}
+{% endfor %}
+
+Please visit the reservations page to start mapping:
+
+{{ reservations_url }}
--- a/src/nyc_trees/apps/mail/views.py
+++ b/src/nyc_trees/apps/mail/views.py
@@ -3,6 +3,8 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from urlparse import urljoin
+
 from django.conf import settings
 from django.core.mail import EmailMessage
 from django.core.urlresolvers import reverse
@@ -10,14 +12,16 @@ from django.shortcuts import get_object_or_404
 from django.template.loader import get_template
 from django.template import Context
 from django.utils.timezone import now
+from django.contrib.sites.models import Site
 
 from django_statsd.clients import statsd
 
 from apps.core.models import User
 
 # Message Types
-GROUP_MAPPING_APPROVED = 'group_mapping_approved'
-RSVP = 'rsvp'
+_RSVP = 'rsvp'
+_GROUP_MAPPING_APPROVED = 'group_mapping_approved'
+_RESERVATION_REMINDER = 'reservation_reminder'
 
 
 def _send_to(user, message_type, *args, **kwargs):
@@ -57,7 +61,7 @@ def notify_group_mapping_approved(request, group, username):
     user = get_object_or_404(User, username=username)
     reservations_url = request.build_absolute_uri(reverse('reservations'))
     return _send_to(user,
-                    GROUP_MAPPING_APPROVED,
+                    _GROUP_MAPPING_APPROVED,
                     group=group,
                     reservations_url=reservations_url)
 
@@ -69,6 +73,16 @@ def notify_rsvp(request, user, event):
     })
     event_url = request.build_absolute_uri(relative_event_url)
     return _send_to(user,
-                    RSVP,
+                    _RSVP,
                     event=event,
                     event_url=event_url)
+
+
+def send_reservation_reminder(user_id, **kwargs):
+    user = get_object_or_404(User, id=user_id)
+    reservations_url = urljoin(
+        'http://%s' % Site.objects.get_current().domain,
+        reverse('reservations'))
+    return _send_to(user, _RESERVATION_REMINDER,
+                    reservations_url=reservations_url,
+                    **kwargs)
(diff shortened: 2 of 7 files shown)


src/nyc_trees/apps/mail/templates/mail/reservation_reminder.txt
```diff
@@ -0,0 +1,4 @@
+The following blockface reservations for user '{{ user.username }}' will expire soon:
```
jwalgran: We will get "real" copy from Parks, but we will most definitely want to include an absolute link to the reservations page in the reminder to the mapper can act on it right away.

src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
```diff
@@ -0,0 +1,55 @@
...
+_COMMAND_NAME = 'apps.survey.management.commands.send_reminder_emails'
+_ALREADY_SENT_MESSAGE = ('WARNING: Reservation emails sent already today. '
+                         'Doing nothing!')
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        today = now().date()
+        _, unlocked = TaskRun.objects.get_or_create(name=_COMMAND_NAME,
+                                                    date_started=today)
```
maurizi: As we discussed, this does *not* provide a locking mechanism.  We'll need to lock the entire taskrun table explicitly.  I'm working on a PR for that now.

src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
```diff
@@ -0,0 +1,55 @@
...
+            user_data.append(reservation)
+            user_reservations[reservation.user_id] = user_data
+
+        for user_id, reservations in user_reservations.items():
+            try:
+                send_reservation_reminder(user_id, reservations=reservations)
+            except smtplib.SMTPException:
+                continue
+            self.stdout.write("Email sent to user_id: '%s'" % user_id)
+            reservation.reminder_sent_at = now()
```
jwalgran: At this point we have a ``reservations`` list, not a single ``reservation``. I don't ``reservation`` is in scope here.

We need to update ``reminder_sent_at`` for all the ``reservations``, preferably in a single update query.

steventlamb: whoa, I think what's happening is reservation is always equal to the last member of the previous forloop. This code did not fail with an undefined global, so that or something similar must be the case. Nice catch, I will fix.

src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
jwalgran: Because we are tracking successful notifications directly on the reservation row, we don't need to enforce that this command runs only once per day, just that it does not run more than once at any given time.

If it makes things easier to enforce the once per day rule, that is fine, and we don't have to rip out an already working thing. In the unlikely event that we would ever want to force it to run more than once per day, we could manipulate the taskrun table.

steventlamb: With the way `TaskRun` is designed, our current choices are to enforce once a day or not at all.

hectcastro: > Because we are tracking successful notifications directly on the reservation row, we don't need to enforce that this command runs only once per day, just that it does not run more than once at any given time.

This was my understanding too. It is also possible that I'm not following what `TaskRun` is meant to do correctly, either.

In RRP, I mentioned [`pg_try_advisory_lock`](http://www.postgresql.org/docs/9.3/static/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS-TABLE) as a potential solution to achieving the "it does not run more than once at any given time" objective. It is not tied to a table, and returns `t` immediately if the lock was successfully acquired, `f` otherwise. Closing the connection also releases the lock (good when a failure occurs while you have the lock, but unclear how that works with Django connection pooling). `pg_try_advisory_lock` also has 2 different signatures, but both provide a way to namespace the lock to a specific task type.

This may or may not be useful, but I prototyped a tool that uses the `pg_try_advisory_lock` mechanism after discussing several approaches the JEST has taken in the past to solve this same problem. Happy to discuss that too if anyone wants to go through it.

steventlamb: `pg_try_advisory_lock` looks like a nice solution, making an arbitrary identifier is definitely better than trying to lock down aspects of a table for a given feature. I think we should reach for it if it turns out we need to run tasks more than once a day. Otherwise, we should stay with `TaskRun`.

steventlamb:
@jwalgran fixed up your comments

src/nyc_trees/apps/mail/views.py
```diff
@@ -56,6 +58,16 @@ def notify_group_mapping_approved(request, group, username):
...
     return _send_to(user,
-                    GROUP_MAPPING_APPROVED,
+                    _GROUP_MAPPING_APPROVED,
                     group=group,
                     reservations_url=reservations_url)
+
+
+def send_reservation_reminder(user_id, **kwargs):
+    user = get_object_or_404(User, id=user_id)
+    reservations_url = ('http://%s/%s/' %
```
jwalgran: I can't thing of a way to avoid hard-coding the protocol, so I would prefer we use "https://" which is more correct, and works everywhere except dev, where we understand how to deal with it.

jwalgran: When I tested this, the URL in the email had extra slashes:

```
...

Please visit the reservations page to start mapping:

http://localhost:8000//blockface//
```

jwalgran:
Command is reporting both email sent and no email sent. 

```
vagrant ssh app -c 'cd /opt/app && envdir /etc/nyc-trees.d/env ./manage.py send_reminder_emails'
Email sent to user_id: '1'
no Emails sent
```

I think it it because the ``else`` clause of a for loop fires unless you call break within the loop. https://docs.python.org/2/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops

steventlamb:
@jwalgran comments addressed

jwalgran:
+1

PR #918, opened 2015-03-19
Fix locking of DB for starting Celery tasks

maurizi:
Row-level locking is not adequate to ensure that only a single task is
queued, since we add new rows for each task.

This is fixed by acquiring an exclusive lock on the entire table,
rather than a subset of row.

### DIFF
--- a/src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
+++ b/src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
-from django.db import transaction
+from django.db import transaction, connection
 from django.core.management.base import BaseCommand
 from django.utils.timezone import now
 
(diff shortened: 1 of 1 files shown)


src/nyc_trees/apps/core/management/commands/start_daily_celery_task.py
```diff
@@ -35,11 +35,15 @@ def handle(self, *args, **options):
...
-        # select_for_update will block if there are other transactions running
-        # which have also selected this table for update.
-        task_runs = TaskRun.objects.select_for_update().filter(name=task_name)
+        # Attempt to acquire a lock to the taskrun table, blocking if another
+        # transaction already has a lock.
+        # Since we create new rows to indicate when a task has been run, we
+        # need to lock the entire taskrun table, not just a subset of rows.
+        # This prevents us from using the builtin 'select_for_update'
+        with connection.cursor() as cursor:
+            cursor.execute('LOCK TABLE core_taskrun IN EXCLUSIVE MODE')
```
steventlamb: Consider 
```
                cursor.execute('LOCK TABLE %s IN EXCLUSIVE MODE'
                               % TaskRun._meta.db_table)
```

steventlamb:
+1

steventlamb:
tested, works

steventlamb:
woops, assigned to wrong person. changing.

PR #919, opened 2015-03-19
Add support for Route 53 failover to static S3 website

hectcastro:
If all of the nodes behind the application load balancer happen to fail, Route 53 will begin passing requests through to the failover DNS endpoint. In this case, the failover endpoint is a static S3 website, fronted by CloudFront, with SSL certificates that match the application load balancers.

The CloudFront dependency is necessary because we're steering users through HTTPS to interact with the application. The S3 website endpoints are unable to handle HTTPS, which could lead to users seeing connection errors when they should really be seeing the failover website. CloudFront supports HTTPS connections with our SSL certificates via SNI.


maurizi:
Looks good to me.
Where will the static S3 site come from?  Something setup manually?

hectcastro:
Buckets already existed on both accounts and were setup as websites (their names have to match the domains). No useful static site exists for the failover, yet: http://treescount.azavea.com.s3-website-us-east-1.amazonaws.com/

hectcastro:
@notthatbreezy noted that the failover site returns a 200, which may not be the status we want to signal in the event of a failure. A failover event should return the static site content, but come along with an HTTP 503 error.

hectcastro:
Pushing Chris' suggestion off into https://github.com/azavea/nyc-trees/issues/921.

PR #925, opened 2015-03-20
Add navlink to main site from admin

steventlamb:
This is very useful during development and can only improve the end user
experience. You can use this link to get back to the site, from the admin.

screenshot from 2015-03-20 15 54 41


maurizi:
Looks good to me

jwalgran:
+1

PR #926, opened 2015-03-20
Add celery tasks for dumping tree and blockfaces tables to shapefiles

### LINKED ISSUE #864: Management command to dump tree data with geometry to shapefile

maurizi:
To test, run `./scripts/manage.sh start_daily_celery_task dump_trees` and `./scripts/manage.sh start_daily_celery_task dump_blockfaces` then verify that the shapefiles files are dumped to `/var/www/nyc-trees/media/dump`

This is a preview PR, since it still needs some additional work to ensure the dumps are stored in a secure fashion when storing on S3.

Fixes #864

### DIFF
--- a/deployment/ansible/roles/nyc-trees.app/tasks/dependencies.yml
+++ b/deployment/ansible/roles/nyc-trees.app/tasks/dependencies.yml
@@ -5,6 +5,7 @@
     - "binutils=2.24*"
     - "libproj-dev=4.8.0*"
     - "gdal-bin=1.10.1*"
+    - "libgdal1-dev=1.10.1*"
 
 - name: Configure the main PostgreSQL APT repository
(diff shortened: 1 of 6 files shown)


jwalgran:
Is there a way to share ``trees.sql`` with the visual survey rendering? If we discover a bug, I would hate to fix it one place and forget to fix it in the other.

maurizi:
@jwalgran I'm not too happy with with how similar the two SQL files are but I couldn't think of easy ways to refactor them to share code without significantly impacting readability.  If you have some suggestions I welcome them.

jwalgran:
I am cool with adding big scary comments to each one, pointing to the other. We have done that for hard to sync resources in the past and it gives us some coverage in the case of future fixes.

jwalgran:
+1

dumped data in qgis next to the survey:

screen shot 2015-03-23 at 2 24 12 pm

PR #929, opened 2015-03-23
Add autoreload option to debugcelery.sh script

maurizi:


steventlamb:
+1 this works and is useful

PR #933, opened 2015-03-24
Delay s3 tool import to prevent dev celery crash

jwalgran:
Dev VMs do not have s3 tooling installed. Celery startup was failing because background data dump tasks that make use of the PrivateS3BotoStorage class were trying to import unavailable tools.


maurizi:
+1

RickMohr:
Merging so I can use it.

PR #937, opened 2015-03-24
[On Hold] Attach an event PDF to RSVP confirmation email

### LINKED ISSUE #856: Send email after RSVP

jwalgran:
Since the meeting point for an event may not be easily expressed as an address, it helps to send the registrant a map centered on the event location.

I made the ``_send_to`` helper expect and handle a list of attachments because reservation confirmation emails will need to include both a static data collection template and the dynamic reservation map PDF.

Test by running a devserver in a console then RSVPing to an event for which a PDF has been generated. You should see a multipart email appear in the console.

Fixes #856


RickMohr:
+1 So simple!

RickMohr:
... and now I'll test it...

RickMohr:
Works, I saw a giant attachment scroll by.

src/nyc_trees/apps/mail/views.py
```diff
@@ -72,10 +82,16 @@ def notify_rsvp(request, user, event):
         'event_slug': event.slug
     })
     event_url = request.build_absolute_uri(relative_event_url)
+    if event.map_pdf_filename:
```
maurizi: I think this will fail if an email is sent while the map is being (re)built.

The filename gets set at the start of the map generation process, so there is a window of time when the file path points to a non-existent file.
The answer is to also check if the file exists, like so: `default_storage.exists(event.map_pdf_filename)`

RickMohr: @maurizi is correct, we should make sure the file exists.

jwalgran:
Superseded by #1035

PR #941, opened 2015-03-25
Add cron job for reservation reminder task

steventlamb:
This adds a cron job that successfully runs this task *every minute*, which is useful for debugging purpose. It will be switched to daily in the early AM before merging.

Opening this PR to open discussion, I have some questions.

### DIFF
--- a/deployment/ansible/roles/nyc-trees.app/tasks/cron.yml
+++ b/deployment/ansible/roles/nyc-trees.app/tasks/cron.yml
@@ -1,0 +1,8 @@
+---
+- name: Send reservation reminder emails at 1AM EST
+  cron: name="send-reminder-emails"
+        user="nyc-trees"
+        job="cd /opt/app && envdir /etc/nyc-trees.d/env ./manage.py send_reminder_emails"
+        hour="5"
+        minute="0"
+        state=present
--- a/deployment/ansible/roles/nyc-trees.app/tasks/cron.yml
+++ b/deployment/ansible/roles/nyc-trees.app/tasks/cron.yml
@@ -2,7 +2,7 @@
 - name: Send reservation reminder emails at 1AM EST
   cron: name="send-reminder-emails"
         user="nyc-trees"
-        job="cd /opt/app && envdir /etc/nyc-trees.d/env ./manage.py send_reminder_emails"
+        job="cd /opt/app && envdir /etc/nyc-trees.d/env ./manage.py send_reminder_emails >> {{ app_cron_log }} 2>&1"
         hour="5"
         minute="0"
         state=present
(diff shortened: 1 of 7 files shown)


steventlamb:
@hectcastro How do you want to to handle logging? After provisioning, I get the following output, because cron is trying to send log output by mail:
```
root@app:/home/vagrant# tail -f /var/log/syslog 
Mar 25 15:00:01 app CRON[2768]: (nyc-trees) CMD (cd /opt/app && envdir /etc/nyc-trees.d/env ./manage.py send_reminder_emails)
Mar 25 15:00:02 app CRON[2767]: (CRON) info (No MTA installed, discarding output)
```

hectcastro:
I created a PR (https://github.com/steventlamb/nyc-trees/pull/1) with support for logging after pulling down this branch.

Looking at the messages that end up coming out, I think it would be helpful to prefix them with the management command that it is executing.

So instead of:

```
vagrant@app:~$ tail -f /var/log/nyc-trees-cron.log
Using json JSON module.
WARNING: Reservation emails sent already today. Doing nothing!
```

Have something like:

```
vagrant@app:~$ tail -f /var/log/nyc-trees-cron.log
Using json JSON module.
send_reminder_emails: Reservation emails sent already today. Doing nothing!
```

Unrelated, I noticed that this management command does a try/catch on `SMTPException`. Not sure how that would ever get thrown given that we're using the SES API for sending e-mail in AWS.

deployment/ansible/roles/nyc-trees.app/tasks/cron.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Send reservation reminder emails
+  cron: name="send-reminder-emails"
+        user="nyc-trees"
+        hour="1"
```
hectcastro: This is going to fire at 01:00 UTC.

deployment/ansible/roles/nyc-trees.app/tasks/log-rotation.yml
```diff
@@ -1,29 +1,26 @@
...
-           creates={{ app_log }}
+  command: touch {{ item }}
+           creates={{ item }}
+  with_items:
+    - "{{ app_log }}"
+    - "{{ app_gunicorn_log }}"
+    - "{{ app_cron_log }}"
 
 - name: Set application log file permissions
   file: path={{ app_log }}
```
hectcastro: This should be `{{ item }}`.

hectcastro:
Tested the cron task, logging to the file, and checking to see if it flowed into Kibana.

:+1: after fixing the `app_log` reference I introduced.

steventlamb:
@hectcastro rebased, can you take one more quick look?

I'm blocked on a seemingly unrelated kibana issue:
screenshot from 2015-03-25 18 18 01

Otherwise, this should be ready to merge. I tested and verified that log entries are landing in the correct log file.

steventlamb:
After provisioning, I'm able to see log messages in Kibana.

hectcastro:
@azavea-bot rebuild

hectcastro:
Ran through this again on a fresh provision. Everything looks good.

PR #953, opened 2015-03-26
Add log shipping support for Celery

hectcastro:
These changes wire up log shipping for the Celery logs into Logstash/Kibana.


maurizi:
Looking at this now

maurizi:
+1, I triggered an error and it showed up in Kibana

PR #979, opened 2015-03-27
Case-insensitive logins

### LINKED ISSUE #945: Make user IDs not case sensitive

RickMohr:
Fixes #945

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.db import models
+from django.contrib.auth.forms import AuthenticationForm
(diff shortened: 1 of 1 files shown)


steventlamb:
This looks like the right way to do this, but I can't find a canoncal reference to this technique. This is all I found, which is only tangentially related:
https://docs.djangoproject.com/en/1.7/topics/serialization/

For my education, can you provide some background into the decision process?

steventlamb:
+1 tested, works

jwalgran:
Merging for Rick while he is out of the office

RickMohr:
@steventlamb, I got the idea to override `get_by_natural_key` from a comment in https://djangosnippets.org/snippets/1368/ and to change
`AuthenticationForm.error_messages['invalid_login']` from reading the `django-registration` code.

PR #982, opened 2015-03-29
Add search and filters to survey admin

### LINKED ISSUE #963: Add search box to django admin survey page

steventlamb:
fixes #963 on github

this adds search and filters as pictured:
screenshot from 2015-03-29 16 18 36

It's clear that these filters would be useful, but I'm not sure what we're looking for as far as search fields. Anything that is walkable in standard orm style (ie. `user__username`, as included) is readily available.

@dboyer thoughts?

### DIFF
--- a/src/nyc_trees/apps/survey/admin.py
+++ b/src/nyc_trees/apps/survey/admin.py
@@ -14,6 +14,10 @@ class TreeInline(admin.TabularInline):
 
 class SurveyAdmin(admin.ModelAdmin):
     inlines = [TreeInline]
+    search_fields = ('id', 'quit_reason', 'user__username')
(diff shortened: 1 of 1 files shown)


jwalgran:
Adding "id" to the list of search fields will allow Parks to run arbitrary QA SQL queries agains the SQL dumps and then pull up problematic surveys by id through the admin.

```
search_fields = ("id", "quit_reason", "user__username")
```

screen shot 2015-03-30 at 9 44 38 am

jwalgran:
+1 searching and filtering work well.

dboyer:
I'm guessing ID and user are the most likely ways they will search this info.

PR #983, opened 2015-03-30
Align staging and production environments

hectcastro:
The following changes are aimed at aligning the staging and production environments:

- Use private hosted zone IDs vs. private hosted zone names
- Add additional instances types
- Label Route 53 hosted zones with VPC ID
- Remove specific data store names
- Switch VPC availability zones

AWS availability zones names don't translate across accounts, so in this case all we're doing is using the same labels.

### DIFF
--- a/deployment/troposphere/template_utils.py
+++ b/deployment/troposphere/template_utils.py
@@ -11,20 +11,21 @@ EC2_REGIONS = [
     'us-east-1'
 ]
 EC2_AVAILABILITY_ZONES = [
-    'b',
+    'a',
     'd'
 ]
 EC2_INSTANCE_TYPES = [
-    't1.micro',
     't2.micro',
     't2.medium',
-    'm3.medium'
+    'm3.medium',
+    'm3.large'
 ]
 RDS_INSTANCE_TYPES = [
     'db.t2.micro',
     'db.m3.large'
 ]
 ELASTICACHE_INSTANCE_TYPES = [
+    'cache.t2.small',
     'cache.m1.small'
 ]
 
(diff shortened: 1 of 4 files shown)


deployment/troposphere/template_utils.py
```diff
@@ -11,20 +11,21 @@
     'us-east-1'
 ]
 EC2_AVAILABILITY_ZONES = [
-    'b',
+    'a',
```
jwalgran: I don't understand the change from 'b' to 'a'. I see your commit message about them not translating across accounts. Does this mean that 'b' is not available in the staging account, but 'a' and 'd' are available in both?

hectcastro: There is no `b` in the production account.

jwalgran:
Do I understand this correctly: from now on we will be passing a ``PrivateHostedZoneId`` as part of our Jenkins deploy, so we can set the zone correctly depending on whether we are in staging or production?

It looks like a good idea to let AWS assign values to the ``DBInstanceIdentifier`` and ``ClusterName`` parameters.

deployment/troposphere/template_utils.py
jwalgran: Got it. Thanks.

hectcastro:
> Do I understand this correctly: from now on we will be passing a PrivateHostedZoneId as part of our Jenkins deploy, so we can set the zone correctly depending on whether we are in staging or production?

This is a precautionary measure to support more than one hosted zone of the same name under one account. This is primarily for the `*.service.nyc-trees.internal` hosted zones so that we can have something like two `database.service.nyc-trees.internal` records with different hosted zone IDs.

The separation between staging and production would happen by using the different access and secret key pairs.

jwalgran:
Got it. I did not run deployment_helper.sh, but the changes look good to me.
+1

PR #985, opened 2015-03-30
Make `debugcelery` stop the correct upstart job

steventlamb:
fixes an issue @jwalgran and I ran into during development where the
script will start more celery workers that don't seem to get jobs if
there is already a celery process in place.


jwalgran:
+1 this correctly shuts down the service

PR #986, opened 2015-03-30
Limit survey length to blockface length

### LINKED ISSUE #838: Crash if distance to tree is near or beyond end of block
Create a survey with one tree, whose distance is large (e.g. 1234), and submit.
500 Error:
```
line_interpolate_point: 2nd arg isnt within [0,1]
CONTEXT:  SQL function "st_line_interpolate_point" statement 2
PL/pgSQL function layoutboxes(geometry,boolean,double precision[],double precision[],double precision[],double precision[]) line 46 at assignment
SQL function "layoutboxes" statement 1
```

steventlamb:
attempts to resove #838, but does not fix the problem

This check was added in an attempt to resolve #838, but something else
is contributing to the problem. This check will make sure that the total
sum of tree distances is less than the length of the blockface, but for
total tree lengths that are *slightly less than* the length of the
blockface, the issue raised in #838 still occurs.


maurizi:
It seems less than ideal to only perform this validation server-side, since the client doesn't have a good way of reporting server-side validation.  Is there a way we can precompute the blockface length and do this check on the client?

ebarry:
Piping up to say that because of user error, and measuring wheel error,
limiting survey length to blockface length will not improve data quality.
Advise to not disrupt user entered data before QAQC on Parks side.

steventlamb:
@ebarry Thanks for your comments. I've reflected your comments here as well:
https://github.com/azavea/nyc-trees/issues/838

PR #991, opened 2015-03-30
Hide progress counts in soft launch mode / show map links in full access mode

jwalgran:
There will be no progress until the full launch, so these numbers will always be zero.

Progress map links were removed in 8b2edea, but we want them to be visible when the site is in full access mode.


steventlamb:
+1, works

PR #996, opened 2015-03-31
Fix filter on private hosted zone ID

hectcastro:
Stack outputs contain the VPC ID, and the VPC ID is needed to filter the Route 53 hosted zones to narrow down the ID.


jwalgran:
:+1:

PR #1000, opened 2015-03-31
Add Typography.com production fonts

hectcastro:
Adding production font assets to the repository in order to satisfy requirements for "production mode".

See also: http://www.typography.com/cloud/user-guide/moving-to-production


steventlamb:
+1

I found this helpful:
http://www.typography.com/cloud/user-guide/development-and-production
> Moving your project into Production doesn’t require any changes to your code. You’ll use the same CSS Key to engage the delivery agent, which will continue to determine what kinds of fonts each visitor needs, and direct the request to the correct files on your server.

PR #1009, opened 2015-03-31
Set sidebar link hrefs

jwalgran:
Links provided by Parks.

### DIFF
--- a/src/nyc_trees/apps/core/templates/core/partials/nav.html
+++ b/src/nyc_trees/apps/core/templates/core/partials/nav.html
@@ -36,8 +36,10 @@
     <div class="nav-side-sub">
         <h3>NYC</h3>
         <h6>Copyright 2015, The City of New York</h6>
-        <a href="#">Contact Us</a>
-        <a href="#">Privacy Policy</a>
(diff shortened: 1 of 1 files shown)


jwalgran:
@steventlamb rebased with a late-breaking change from Parks

src/nyc_trees/apps/core/templates/core/partials/nav.html
```diff
@@ -36,8 +36,10 @@
     <div class="nav-side-sub">
         <h3>NYC</h3>
         <h6>Copyright 2015, The City of New York</h6>
-        <a href="#">Contact Us</a>
-        <a href="#">Privacy Policy</a>
+        <a title="treescount.help@parks.nyc.gov" href="mailto:treescount.help@parks.nyc.gov">Contact Us</a>
```
hectcastro: We have this e-mail address as `DEFAULT_HELP_EMAIL` in the settings file if we want to reuse it.

steventlamb:
+1 links work

src/nyc_trees/apps/core/templates/core/partials/nav.html
jwalgran: It's embedded in emails too. Building a context processor and DRYing out the templates is a good idea but a little more than I want to tackle right now.

PR #1010, opened 2015-03-31
Differentiate 'map another' go url

### LINKED ISSUE #826: Map another block link should differentiate based on source (event vs individual)
```
maurizi added a note 2 hours ago
... [] ... this is another survey for reservations vs. survey for event bug: the URL for this href should be different depending on how you arrived at / submitted this survey
```
https://github.com/azavea/nyc-trees/blob/ae8d148691b406def166b5bcd37abf6f41dd826d/src/nyc_trees/apps/survey/templates/survey/partials/map_another_popup.html#L11

steventlamb:
fixes #826 on github

Because we don't store source information the survey / blockface data,
we have to rely on the network workflow to pass along the necessary
parameters between two pages.


maurizi:
I went through a survey for a event, and chose "No, restart" when I got I got the confirmation page.
It put me at `/survey/`, not the event survey page.

When I went back and chose "Yes", I saw a popup saying "You've mapped all your reserved blocks!".

maurizi:
You may be able to remove the JS code by adding a second URL for the "confirmation" page at `/survey/<group_slug>/event/<event_slug>/confirm/`, similar to how the survey submission views work.

Then the links in the Django template can be `{{ request.get_full_path }}/confirm`.

steventlamb:
@maurizi these are outside the scope of #826 and the map another popup, but I'll try to tack on these fixes.

In #826, you (transitively) said 'this is another survey for reservations vs. survey for event bug'. I can't find another issue for survey source bugs, does this cover it?

PR #1012, opened 2015-03-31
Add affiliation to group detail page

### LINKED ISSUE #844: Display affiliation on group detail page

steventlamb:
fixes #844 on github

### DIFF
--- a/src/nyc_trees/apps/users/templates/groups/detail.html
+++ b/src/nyc_trees/apps/users/templates/groups/detail.html
@@ -67,6 +67,16 @@
                 </div>
             </div>
         {% endif %}
+        {% if group.affiliation %}
+            <div class="description block item">
+                <div class="row">
+                    <div class="col-xs-8">
+                        <h5>Affiliation</h5>
+                        <div class="h6">{{group.affiliation}}</div>
+                    </div>
+                </div>
+            </div>
+        {% endif %}
     </section>
     <section class="contributions">
         <h4 class="section-heading">Contributions</h4>


src/nyc_trees/apps/users/templates/groups/detail.html
```diff
@@ -67,6 +67,14 @@ <h4 class="section-heading">Description</h4>
...
             </div>
         {% endif %}
+        <div class="description block item">
+            <div class="row">
+                <div class="col-xs-8">
+                    <h5>Affiliation</h5>
+                    <div class="h6">{{group.affiliation}}</div>
+                </div>
+            </div>
+        </div>
```
maurizi: `affiliation` is not a required field, so we should wrap this in an `{% if group.affiliation %}` block.

steventlamb: I didn't do that because I wasn't sure if it's the best UX to automatically hide fields that don't exist. @jfrankl @dboyer thoughts?

maurizi: Well we are doing it for the field above this one, `group.contact_url`

¯\_(ツ)_/¯

dboyer: I'm torn on this but inclined to say it should be hidden if not completed. I'll defer to @jfrankl though.

jfrankl: I think we should hide fields that are empty.

PR #1022, opened 2015-04-01
linked the navbar logo to home - fixes #1013 and fixes #441

### LINKED ISSUE #441: Clicking the header links to root page
It may be handy to jump to the dashboard from anywhere.

### LINKED ISSUE #1013: No way to return to home page while in soft launch mode
If the user clicks the about page in the side navigation or the login or register buttons, there's no way to get back to the home page other than clicking the back button in the browser.

designmatty:


maurizi:
+1

PR #1026, opened 2015-04-01
fixed linebreaks on the homepage - fixes #1014

### LINKED ISSUE #1014: Change line breaks in text in hero image
The current arrangement looks odd since "York City" ends on it's own line. Ideally, the lines would be

We're mapping
every street tree in
New York City.

In could also be on the third line, if that's better for alignment purposes.

designmatty:


maurizi:
+1

PR #1032, opened 2015-04-01
Feature/974/show events on dashboard

steventlamb:


jwalgran:
Giving this a runthrough. Looks like readable solution to a complicated workflow.

jwalgran:
The "Recommended" list is showing events that the person has RSVPed to. I think it makes sense for the lists to be mutually exclusive. "Here are the events you are going to, and here are events you could go to."

jwalgran:
We probably should not "Recommend" mapping events to users who do not have field training. It would save some frustration and should be straightforward to conditionally add a filter when ``request.user.field_training_complete`` is falsy.

jwalgran:
We need to filter out events in the past from both lists.

steventlamb:
It turns out we already filter past events. Is it possible that you have events from early 2016 in your dev db?

jwalgran:
Maybe it is my test. I just did this to one of my events

```python
e.begins_at = datetime(2015,3,31)
e.save()
```

And it shows up at the top of the list

screen shot 2015-04-02 at 10 09 22 am

jwalgran:
Yeah, I set ``ends_at`` and it disappeared. Bad test on my part.

jwalgran:
"Show more" chunking and empty list content works well on both tabs

steventlamb:
@jwalgran I believe all comments are addressed and that this is working. I also tested some of the other event lists for regressions.

jwalgran:
Crash when logged out

```
Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py" in routed
  33.             return req_method(request, *args2, **kwargs2)
File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py" in wrapper
  79.                 params = callable_or_dict(request, *args, **wrapper_kwargs)
File "/opt/app/apps/home/views.py" in home_page
  56.         .as_context(request))
File "/opt/app/apps/event/event_list.py" in as_context
  270.                     if k.should_show(request)])
File "/opt/app/apps/event/event_list.py" in <lambda>
  46.                       lambda request: request.user.field_training_complete)
File "/usr/local/lib/python2.7/dist-packages/django/utils/functional.py" in inner
  225.         return func(self._wrapped, *args)

Exception Type: AttributeError at /
Exception Value: 'AnonymousUser' object has no attribute 'field_training_complete'

jwalgran:
When I remove the field training flag from a user, the "Recommended" section disappears

Here is what a user with field training sees

screen shot 2015-04-02 at 11 34 50 am

I unchecked the field training checkbox

screen shot 2015-04-02 at 11 35 09 am

Recommended is now gone

screen shot 2015-04-02 at 11 35 27 am

steventlamb:
@jwalgran I think it works the way you requested now, sorry for the mixup

jwalgran:
+1 works nicely. Recommended contents change based on field training status.

PR #1034, opened 2015-04-01
Pin Django dependency to 1.7.x

jwalgran:
The Internet suggested this range syntax for safety

https://groups.google.com/forum/#!msg/python-virtualenv/DvG1InRGdR0/14b9CISO2AcJ


hectcastro:
Lol. I was just figuring out a fix for this after hitting it with MMW.

:+1:

PR #1044, opened 2015-04-02
Add task, management command and cron job for zipping up DB dumps

### LINKED ISSUE #866: Automate dumps and make zip available to parks only - cron scheduled
Parks has requested embedding the dump date in the name of the zip file

maurizi:
Refactors the 'dump_model' function to be a Celery task, and moves it
into the 'census_admin' app alongside the Tree and Blockface dumping tasks.

Since only the Tree & Blockface tasks were using 'start_daily_celery_task'
I've removed that management command in favor of one that only deals with
full DB dumps

Fixes #866


jwalgran:
+1 works. I was able to dump to S3, download and unpack the zip, and open the tree and blockface shapefiles.

screen shot 2015-04-02 at 3 31 39 pm

Using the broker for the results backend seems like a good idea since the arguments and return values are short strings and we will only be running tasks that use the backend a few hundred times.

PR #1056, opened 2015-04-02
Quote values in CSV dumps

jwalgran:
[This PR depends on #1044.  983a4b5 is the only new commit]

Prevents the dump task from producing invalid CSVs when a text value contains a comma.


jwalgran:
Closing to get a proper waffle hookup

PR #1057, opened 2015-04-02
Quote values in CSV dumps

### LINKED ISSUE #934: Specify quoting for non-numeric fields when exporting CSVs
Text field values need to be wrapped in quotes.  djqscsv supports passing arbitrary args to the CSV writer.

https://docs.python.org/2/library/csv.html#csv.QUOTE_NONNUMERIC

jwalgran:
Prevents the dump task from producing invalid CSVs when a text value contains a comma.

Test by making sure Celery is running and then running the ``dump_db`` command

```
./scripts/manage.py dump_db
```

In dev, the files will be written to ``/var/www/nyc-trees/media/dump``

If you want to test this more than once you will need to delete records from ``core_taskrun``.

Fixes #934

### DIFF
--- a/src/nyc_trees/apps/census_admin/tasks.py
+++ b/src/nyc_trees/apps/census_admin/tasks.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+import csv
 import os
 import tempfile
 import json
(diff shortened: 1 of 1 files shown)


maurizi:
+1.

I haven't tested it, but can if you feel it is necessary

PR #1059, opened 2015-04-03
Add more flexible way of overriding passwords with Packer

hectcastro:
This change threads through an override for `postgresql_password` with the `NYC_TREES_DB_PASSWORD` environment variable at AMI creation time.


hectcastro:
@azavea-bot rebuild

jwalgran:
So, before this commit, we could only write the static ``postgresql_password`` value from ``ansible/group_vars/packer`` into ``/etc/nyc-trees.d/env`` on the AMI. After this commit, we set the ``NYC_TREES_DB_PASSWORD`` env var on the packing machine, which initializes the packer variable, which gets passed in to Ansible, overriding the static group_var. Is that a correct restating of the solution?

hectcastro:
Yep.

jwalgran:
+1 I did not packer a new AMI with this, but it looks good

PR #1061, opened 2015-04-03
Bug/826/differentiate take 2

steventlamb:


maurizi:
The "No, restart" link works correctly.

The "No, Submit for Review" and "Yes" links still give me a popup that says "You've mapped all your reserved blocks!"

steventlamb:
Oh, are you saying that when you complete the survey from an event, you get the reserved blocks issue, which is not connected to event surveys?

maurizi:
@steventlamb Yes.

steventlamb:
@jwalgran @maurizi updated

jwalgran:
Testing this now

src/nyc_trees/js/src/surveyPage.js
```diff
@@ -458,8 +458,12 @@ function submitSurveyWithTrees() {
             data: JSON.stringify(data)
         })
         .done(function(content) {
+            // make sure not to take the hashstring along for the ride
+            var href = window.location.origin +
+                    window.location.pathname +
+                    'confirm/' + content.survey_id + '/';
```
jwalgran: Whenever we embed paths in JS, we have been leaving comments for future refactorers in the js and py files that say something like "this URL needs to be kept in sync with such/and/such/view".

jwalgran:
+1 event mapping and free mapping both "loop around" to the correct pages when a survey is finished.

PR #1064, opened 2015-04-03
Try to send emails multiple times before failing 

steventlamb:
@jwalgran @hectcastro @maurizi what do you think of this technique?

things of note:
- I added 1/2 second sleeping to stay under request limits. This is irrespective of retries - we should do this anyway
- I gave each email 4 attempts to send. It'll loop through the list of emails over and over until they've all succeeded or tried 4 times.
- It's going to be really hard to verify that this works in production.
- Because this isn't using celery, I didn't add it in just for retries. If this command was using celery, I would've gone that route. I'm not opposed to it now.

### DIFF
--- a/src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
+++ b/src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
@@ -6,6 +6,8 @@ from __future__ import division
 import operator
 import smtplib
 
+import time
+
 from datetime import timedelta
 
 from boto.ses.exceptions import SESError
@@ -22,8 +24,14 @@ from apps.survey.models import BlockfaceReservation
 from apps.mail.views import send_reservation_reminder
 
 _COMMAND_NAME = 'apps.survey.management.commands.send_reminder_emails'
-_ALREADY_SENT_MESSAGE = ('WARNING: Reservation emails sent already today. '
+_ALREADY_SENT_MESSAGE = ('WARNING: reservation emails sent already today. '
                          'Doing nothing!')
+
+_RESERVATIONS = 'reservations'
+_REMAINING_RETRIES = 'remaining_retries'
+_RETRY_COUNT = 4
+
+_DONE = -1
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
```diff
@@ -60,25 +66,50 @@ def handle(self, *args, **options):
...
+                except (smtplib.SMTPException, SESError):
+                    user_remaining_retries -= 1
+                    user_data[_REMAINING_RETRIES] = user_remaining_retries
+                    if user_remaining_retries > 0:
+                        remaining_retries = True
+                    continue
+                finally:
+                    # sleep to avoid hitting request-per-second limits in AWS
+                    # given that we have plenty of time for this tasks to run
+                    time.sleep(0.5)
```
jwalgran: Good catch. I had previously not read the doc that states in addition to the 50k/24h limit there is a 14/second limit.

It would be helpful to drop a link to http://docs.aws.amazon.com/ses/latest/DeveloperGuide/limits.html in the comment

src/nyc_trees/apps/survey/management/commands/send_reminder_emails.py
```diff
@@ -60,25 +66,50 @@ def handle(self, *args, **options):
...
+
+                self._log("Email sent to user_id: '%s'" % user_id)
+                ids = map(operator.attrgetter('id'), reservations)
+                (BlockfaceReservation
+                 .objects
+                 .filter(id__in=ids)
+                 .update(reminder_sent_at=now()))
```
jwalgran: I am just now realizing that management tasks run in an implicit transaction. We could send 1000 emails, mark them as sent, and then, if the process crashes, the transaction will roll back and we will resend all the emails the next day. If we don't notice the crash and it is persistent, we will start spamming people day after day.

If we stick with a bulk transaction, we are going to need to know, immediately, if the transaction rolls back for some reason. It may be less nerve wracking to fire off a task for each email.

maurizi: @jwalgran I've never heard of management commands running in an implicit transaction.  I'd be curious to read the docs on that for educational purposes.

hectcastro:
@azavea-bot rebuild

steventlamb:
I added this to the try block:
```python
                    if randint(0, 10) < 8:
                        raise smtplib.SMTPException
```
and a log to the catch:
```
                    print("triggering a retry for: %s - %s left"
                          % (user_id, user_remaining_retries))

```
And got:
```
triggering a retry for: 1 - 3 left
apps.survey.management.commands.send_reminder_emails: Email sent to user_id: '2'
triggering a retry for: 4 - 3 left
triggering a retry for: 5 - 3 left
apps.survey.management.commands.send_reminder_emails: Email sent to user_id: '1'
apps.survey.management.commands.send_reminder_emails: Email sent to user_id: '4'
triggering a retry for: 5 - 2 left
apps.survey.management.commands.send_reminder_emails: Email sent to user_id: '5'
```

Which, other than the surprising number of successes given the 20% probability, looks like it worked.

steventlamb:
@jwalgran awaiting further review on implicit transactions, but otherwise I think this is ready to merge after a git cleanup.

jwalgran:
I think I was conflating management commands with tests and incorrectly assuming that you got a transaction by default.

jwalgran:
+1

steventlamb:
@hectcastro updated as per our discussion. Should behave slightly better if things go wrong.

hectcastro:
:+1:

PR #1077, opened 2015-04-06
Change all variations to block edges

### LINKED ISSUE #1076: Block sides vs block faces
@dboyer: there are a few files that refer to "block sides". @RickMohr suggested that these should be replaced with "block faces". Just wanted to confirm with you that this is the correct move.

jfrankl:
All user-visible "block faces" and "blockfaces" and "block sides" have become "block edges".

Fixes #1076.


RickMohr:
Otherwise +1. Definitely an edgey PR.

Issue #959, opened 2015-03-26
Implement training material text and images


dboyer:
Use existing mgmt command to load flatpages - update content
Images as django static files (which get served by nginx)
Raw content is available in the project file share under ``data\Training_Materials``

dboyer:
Training content sent to @steventlamb and @jwalgran so I'm removing the blocked label. Can send text to others as necessary.

PR #1080, opened 2015-04-08
Topic/959/update training content

steventlamb:


maurizi:
The photo assets added here are going to be brought into our Gulp pipeline, subjecting them to versioning via `gulp-rev-all`.

Because of the, referring to them via `` won't work in production.

We may be able to sidestep the issue by ignoring the entire folder, like we do for our font assets: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/gulpfile.js#L96-L97

The only downside to this is that nginx will still mark them to be cached forever.
We can probably work around that if need be by just changing the filenames whenever we need to change a training image.

jwalgran:
Using a subfolder and adding an additional ``location`` to nginx, in combination with disabling versioning, may get us where we need to be. I am not sure if the location matching will work as I am assuming it will.
```
  location /static/content/ {
    {% if ['packer'] | is_in(group_names) -%}
    expires 2h;
    {% endif %}
 
    alias {{ app_static_root }}content/;
  }

  location /static/ {
    {% if ['packer'] | is_in(group_names) -%}
    etag on;
    expires max;
    {% endif %}

    alias {{ app_static_root }};
  }

steventlamb:
@maurizi @jwalgran updated to solve image problem.

jwalgran:
+1 I deleted all my flatpages, ran the management command, and then walked through the training process, viewing the new content and answering the real quiz questions. The images copied to ``/var/www/static/img/training`` did not get versioned file names.

steventlamb:
fixes #959

Issue #849, opened 2015-03-12
Upload blockfaces for remaining 4 boroughs


dboyer:

dboyer:
We have everything but the Bronx if you want to upload these. Bronx expected in the next few days but maybe not before sprint's end.

PR #1121, opened 2015-04-10
Add Brooklyn, Queens, and Staten Island blocks

jwalgran:
This were imported from shapefiles using the method documented in README.md.

screen shot 2015-04-10 at 9 32 33 am

Connects to #849


RickMohr:
`scripts/manage.sh reload_dev_data`
is giving me a memory error:
```
...
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/loaddata.py", line 142, in load_label
    for obj in objects:
  File "/usr/local/lib/python2.7/dist-packages/django/core/serializers/json.py", line 72, in Deserializer
    stream_or_string = stream_or_string.decode('utf-8')
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
MemoryError: Problem installing fixture '/opt/app/apps/survey/fixtures/blockface.json':
```

jwalgran:
@RickMohr I have split the fixture into chunks. I tried using https://github.com/fastinetserver/django-dumpdata-chunks but it threw a strange error, so I just used UNIX tools to chop up the file.

jwalgran:
@RickMohr ``reload_dev_data`` now uses the chunked files

scripts/aws/setupdb.sh
```diff
@@ -17,7 +17,12 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
 # Run migrations
 envdir /etc/nyc-trees.d/env /opt/app/manage.py migrate
 # Load block face data
-envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata $DIR/../../src/nyc_trees/apps/survey/fixtures/blockface.json
+
+for i in `seq 1 7`;
```
hectcastro: The `seq` part here can be done without a subshell:

```bash
for i in {1..7};
```

Alternatively, `ls` with wildcard matching can catch all of the `blockface_N.json` files:

```bash
ls /opt/app/apps/survey/fixtures/blockface_*.json
```

Combined with `xargs`, you can execute the management command for each match by piping the output:

```bash
ls /opt/app/apps/survey/fixtures/blockface_*.json | xargs -n1 envdir /etc/nyc-trees.d/env /opt/app/manage.py loaddata
```

The `-n1` argument just tells `xargs` to do one at a time vs. its default of all at once.

README.md
```diff
@@ -135,6 +135,35 @@ the appropriate data, then exporting the fixture:
...
+chunks. This process is not fully automated.
+
+This pipline will take a fixture created with ``dumpdata`` on stdin
+and output one object per line.
+
+```
+# convert a JSON array to linewise objects with trailing commas
+# embedded newline is IMPORTANT
+# sed -e 's/ *$//' trims whitespace
+sed -e 's/{"fields/\
```
hectcastro: Going to mess with an alternative to this, just because the literal newline seems like it could cause issues, but not going to let that hold things up.

hectcastro:
I was able to reproduce the map above using both the `reload_dev_data` and `loaddata` management commands.

:+1:

jwalgran:
We will need to update the fixture at least once more when we get the Bronx data. We can work on improving my hack job then.

RickMohr:
`reload_dev_data` worked for me with  `DEBUG=False`. With `DEBUG=True` I got
```
Deleting all existing survey and blockface data
Importing blockface fixture
Installed 20000 object(s) from 1 fixture(s)
bash: line 1: 28522 Killed                  envdir /etc/nyc-trees.d/env ./manage.py reload_dev_data
```
presumably due to  memory leaks.

+1  Great to finally see a whole city of data!!

Issue #1098, opened 2015-04-09
Update species list


dboyer:
We've received a new species list from Parks.

PR #1124, opened 2015-04-10
Update species fixture

jwalgran:
Parks provided us with a new species CSV. I formatted the columns to match our existing schema, and then used [csv2json.py](https://djangosnippets.org/snippets/1680/) to convert it to a Django fixture.

Connects to #1098


steventlamb:
+1, tested via visual inspection of diff, then by truncating species table, loading this fixture, and completing a survey with a random tree and species.

PR #1127, opened 2015-04-10
Add script to open interactive shell in vm

steventlamb:
handy for lots of stuff.


hectcastro:
To me this script seems pretty similar to the functionality provided by `vagrant ssh`, with the exception of the `envdir` call. If it helps people be more effective, I don't have a problem with it. My only concern is that this opens the door a little for stuff working when invoked via this shell vs. when invoked by Upstart.

:+1:

PR #1130, opened 2015-04-13
Allow override of instance types with deployment helper

hectcastro:
The application and tiler CloudFormation templates for have a parameter for instance types, but until now the deployment helper did not have a way to override it. These changes provide a variable to override the instance type for each stack.

In addition, more instance type options were added to the CloudFormation templates for more flexibility.


kdeloach:
+1

PR #1131, opened 2015-04-13
Change CloudWatch alert thresholds based on past alerts

hectcastro:
These changes adjust the CloudWatch alert thresholds based on alerts we've received from the staging, training, and production environments.


jwalgran:
+1

Issue #1128, opened 2015-04-11
Poor tiler query performance


hectcastro:
I pulled a query from the PostgreSQL slow query log that consistently produced long duration times after interacting with the freshest batch of blockface data. Given that it is taking a handful of seconds, there may be some room for optimization:

```
SELECT ST_AsBinary("geom") AS geom,"geojson","group_id","id","survey_type" FROM -- The 'DISTINCT ON' and 'ORDER BY' clause are needed so that we only use
	-- the latest survey for a blockface
	(SELECT
	  DISTINCT ON (block.id)
	  
	  ST_AsGeoJSON(block.geom) AS geojson,
	  
	  block.geom, block.id, turf.group_id,
	  CASE
	    WHEN (survey.user_id IS NOT DISTINCT FROM 3
	          OR survey.teammate_id IS NOT DISTINCT FROM 3) THEN 'surveyed-by-me'
	    WHEN survey.id IS NOT NULL THEN 'surveyed-by-others'
	    WHEN reservation.user_id IS NOT DISTINCT FROM 3 THEN 'reserved'
	    WHEN (block.is_available AND reservation.id IS NULL
	          AND (turf.id IS NULL OR trustedmapper.id IS NOT NULL)) THEN 'available'
	    ELSE 'unavailable'
	  END AS survey_type
	  FROM survey_blockface AS block
	  LEFT OUTER JOIN survey_territory AS turf
	    ON block.id = turf.blockface_id
	  LEFT OUTER JOIN survey_survey AS survey
	    ON block.id = survey.blockface_id
	  LEFT OUTER JOIN survey_blockfacereservation AS reservation
	    ON (block.id = reservation.blockface_id
	        AND reservation.canceled_at IS NULL
	        AND reservation.expires_at > now() at time zone 'utc')
	  LEFT OUTER JOIN users_trustedmapper AS trustedmapper
	    ON (turf.group_id = trustedmapper.group_id
	        AND trustedmapper.user_id = 3
	        AND trustedmapper.is_approved)
	  ORDER BY block.id, survey.created_at DESC NULLS LAST
	) AS query WHERE "geom" && ST_SetSRID('BOX3D(-74.190673828125 40.5722401175879,-73.992919921875 40.72228267265037)'::box3d, 4326)
```

jwalgran:
I turned on ``\timing`` in psql and ran the inner query on my vm

```
SELECT
      DISTINCT ON (block.id)

[...]

PR #1132, opened 2015-04-13
Remove GeoJSON from UTF grids

jwalgran:
Removing the GeoJSON speeds up grid tile queries by 4x on a dev VM. Fetching the GeoJSON for a feature via an AJAX request when it is clicked takes less than 50ms in dev.

Connects to #1128

### DIFF
--- a/src/nyc_trees/js/src/lib/SelectableBlockfaceLayer.js
+++ b/src/nyc_trees/js/src/lib/SelectableBlockfaceLayer.js
@@ -54,15 +54,21 @@ module.exports = BlockfaceLayer.extend({
     },
 
     addBlockface: function(data, latlng) {
-        if (data && data.geojson) {
-            var geom = mapUtil.parseGeoJSON(data.geojson);
-            if (this.options.onAdd(data, geom, latlng)) {
-                this.addData({
-                    "type": "Feature",
-                    "geometry": geom,
-                    "properties": data
-                });
-            }
+        var self = this;
+        if (!data || !data.id) {
+            return;
         }
+        mapUtil.fetchBlockface(data.id).then(function(blockfaceData) {
+            if (blockfaceData && blockfaceData.geojson) {
+                var geom = mapUtil.parseGeoJSON(blockfaceData.geojson);
+                if (self.options.onAdd(data, geom, latlng)) {
+                    self.addData({
+                        "type": "Feature",
+                        "geometry": geom,
+                        "properties": data
+                    });
+                }
+            }
+        });
     }
 });
--- a/src/nyc_trees/js/src/lib/SelectableBlockfaceLayer.js
+++ b/src/nyc_trees/js/src/lib/SelectableBlockfaceLayer.js
@@ -8,12 +8,15 @@ var $ = require('jquery'),
 
     color = '#FFEB3B';
 
-/* Adds a geojson layer to the given map, based on the geojson data of a UtfGrid.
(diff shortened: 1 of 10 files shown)


maurizi:
Looking at this now

src/nyc_trees/js/src/lib/SelectableBlockfaceLayer.js
```diff
@@ -54,15 +54,21 @@ module.exports = BlockfaceLayer.extend({
...
+                var geom = mapUtil.parseGeoJSON(blockfaceData.geojson);
+                if (self.options.onAdd(data, geom, latlng)) {
+                    self.addData({
+                        "type": "Feature",
+                        "geometry": geom,
+                        "properties": data
+                    });
+                }
+            }
+        });
```
maurizi: I'm seeing network requests when selecting blockfaces, even for ones we don't want to highlight (like clicking unavailable blocks on the reservations page).

Looking at the uses of `SelectableBlockfaceLayer.onAdd`, it looks like the only use of the `geom` parameter of `onAdd` is in [`surveyPage.js`](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/surveyPage.js#L133-L156).

I think we can modify that to move most of the logic into `onAdded` and drop the `geom` parameter from `onAdd` (which I should have probably named `shouldAdd`) by modifying `surveyPage.js` like so:
```javascript
    selectedLayer = new SelectableBlockfaceLayer(blockfaceMap, grid, {
        onAdd: function() {
            return true;
        },

        onAdded: function(feature, layer)
            var latLngs = mapUtil.getLatLngs(feature.geometry);

            blockfaceId = feature.properties.id;

            selectedLayer.clearLayers();
            endPointLayers.clearLayers();

            var startCircle = L.circleMarker(latLngs[0], defaultStyle),
                endCircle = L.circleMarker(latLngs[latLngs.length - 1], defaultStyle);

            startCircle.isStart = true;
            endCircle.isStart = false;

            endPointLayers.addLayer(startCircle);
            endPointLayers.addLayer(endCircle);

            showSelectStart();

            mapUtil.zoomToBlockface(blockfaceMap, blockfaceId);
        }
    }),
```

This will let us move the network request to *after* the call to `onAdd`, so we won't have to do it if `onAdd` returns false, like so:
```javascript
    addBlockface: function(data, latlng) {
        var self = this;
        if (!data || !data.id) {
            return;
        }
        if (self.options.onAdd(data, geom, latlng)) {
            mapUtil.fetchBlockface(data.id).then(function(blockfaceData) {
                if (blockfaceData && blockfaceData.geojson) {
                    var geom = mapUtil.parseGeoJSON(blockfaceData.geojson);
                    self.addData({
                        "type": "Feature",
                        "geometry": geom,
                        "properties": data
                    });
                }
            });
        }
    }
```

maurizi:
This works pretty well.  I set the "Network Throttling" in the Chrome dev tools to 2G and while there is a noticeable lag in highlighting the selected blockface, I don't think it's unacceptably slow (and there was a noticeable improvement in  the time it took for the blockfaces to show up).

It would be nice to see the issue I mentioned fixed before this is merged, but I think overall this is a good improvement.

maurizi:
+1

jwalgran:
@maurizi Updated with the onAdd/onAdded modifications

maurizi:
+1

PR #1133, opened 2015-04-13
Expose tiler metatile count as config option

jwalgran:
The metatile setting has a dramatic effect on both real world and perceived tile rendering speed. A larger value may be appropriate for a single tile rendering process but a smaller one be appropriate for
multiple tile rendering processes running in a load balanced cluster. Modifying the server to read the metatile count from the environment allows for experimentation and performance tuning without
code modification or redeployment.


RickMohr:
In OTM we set `metatile: 1` (turned it off). Discussion at https://github.com/OpenTreeMap/OTM2-tiler/pull/46

hectcastro:
Clicking link after link for information on metatiles distracted me from actually looking at the code changes in this PR. Since the default is 4, and everything provisions and works even with the environment variable absent, seems perfectly fine to merge.

:+1:

PR #1139, opened 2015-04-13
Fix unable to RSVP from event if it is full

### LINKED ISSUE #1045: Can't unregister for an event if it is full
1) Create an event with a capacity of 1
2) RSVP for the event

Expected: Button changes to "Attending" and is clickable
Actual: Button changes to "At Capacity" and is not clickable

screen shot 2015-04-02 at 8 52 21 am

screen shot 2015-04-02 at 8 52 27 am

kdeloach:
Previously, one could not unregister from an event that had reached
maximum capacity, because the RSVP button would become disabled and
read "At Capacity".

This fixes that by updating the RSVP button so that it cannot be disabled for
users that have registered for an event.

Fixes #1045


jwalgran:
The button now says "Attending" if I am RSVPd to a full event, and I can click it to give up my registration.

+1

Issue #576, opened 2015-02-16
Training page sidebar text does not match content


maurizi:
When all online training is complete, the main content show "Training Complete", but the sidebar indicates there is more training to be done
![almost_there](https://cloud.githubusercontent.com/assets/4432106/6219191/52580fce-b5f4-11e4-993f-cfbe2c19e82f.png)

dboyer:
Let's leave this for now as we're waiting on text from Parks and TreeKIT.

jfrankl:
Will need a developer to write conditional statements to show correct text based on conditions.

steventlamb:
Changing back to design due to #1143

steventlamb:
Assigning to @dboyer because we're just waiting on content.

dboyer:
Is this a duplicate of #443 ?

PR #1143, opened 2015-04-13
Hide 'you're almost there' when training complete

steventlamb:
addresses #576 on github

This commit changes #576 to a design/content issue and hides some incorrect text for the time being.


jfrankl:
Can you make the appropriate text appear in the left column instead of the right column? It should be in the element with ```.pageheading-description-detail```.

jfrankl:
Actually, never mind. I can take this from here.

jfrankl:
+1

PR #1148, opened 2015-04-13
Use real data for dashboard header

### LINKED ISSUE #714: Use real counts in the dashboard header text
It is currently mocked up with static data

screen shot 2015-03-02 at 4 04 56 pm

maurizi:
Also fixes the tree ticker on the public homepage, which was not being
shown due to a typo.

Fixes #714


maurizi:
Closing this because it is no longer needed

PR #1149, opened 2015-04-13
Show styled footer on desktop

### LINKED ISSUE #609: Show styled footer on desktop
In #608 the footer was hidden for wide viewports because it needs some style fixes

jfrankl:
Fixes #609.

image


jwalgran:
Breaking in the middle of the bold text on a 3.5" phone looks a little awkward. We could try adding non-breaking spaces like we do in the header.

screen shot 2015-04-14 at 9 07 03 am

jwalgran:
There is a common footer at all resolutions

+1

jfrankl:
@jwalgran Good call, just added in that detail d4bab27

Merging now.

PR #1152, opened 2015-04-13
Add default ordering for several models.

### LINKED ISSUE #692: Need to add default sort orders for the Django admin site
We should do this for most of the models in the site, but we definitely need it for:

 - User
 - Group
 - Event

### LINKED ISSUE #1083: Alphabetize the group list in drop down on optional information registration page

maurizi:
Fixes #692, Fixes #1083


RickMohr:
+1

PR #1155, opened 2015-04-13
Add progress circles to group page. Fixes #1100

### LINKED ISSUE #1100: Switch group contributions to progress circles
Now:

image

Should look like:

image

jfrankl:
image

I removed the data about number of attendees because I felt like the number of followers, which we already show on this page, is a better representation of that activity. Plus, this allows us to use the standard 3 circle pattern that is on the dashboard.

### DIFF
--- a/src/nyc_trees/apps/users/templates/groups/detail.html
+++ b/src/nyc_trees/apps/users/templates/groups/detail.html
@@ -83,7 +83,8 @@
         <div class="block nopadding--top clearfix">
             {% include "home/partials/circle.html" with label="Trees" n=3 color="primary" url=progress_page_url %}
             {% include "home/partials/circle.html" with label="Blocks" n=counts.block color="primary" url=progress_page_url %}
-            {% include "home/partials/circle.html" with label="Events" n=counts.event color="primary" modal="#users-modal" %}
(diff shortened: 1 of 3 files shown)


dboyer:
Parks asked for number of attendees. It's a more accurate representation of the number of people participating in a group's activities and how frequently they participate. Someone could follow a group and never attend an event or attend several events and never follow the group. Parks is most interested in the number of people at events and the attendees count included all people at all events for that group (not just unique attendees). It's the equivalent to number of volunteer hours which is an important metric for citizen science projects.

RickMohr:
+1 New style looks good. 

It does lose some information compared to the previous one, for example the previous "Mapped in Area" makes it somewhat clearer that this is a percentage of the group's total area then the current "Blocks". In the previous "Events Held" makes it clear that this count doesn't include events planned for the future, while the current "Events" leaves that in question.

jfrankl:
@dboyer, of these four data points, do you think we could get away with displaying only three? If so, which three do you think are most important?

- Trees
- % Mapped in Area
- Events Held
- Event Attendees

dboyer:
I'd have to ask Parks since they confirmed the four data points. If we dropped one, my vote would be for events. Let me try to get a quick answer.

dboyer:
@jfrankl Parks confirmed that we need to track and display all four items. They use those four metrics in various places throughout their TreesCount work.

PR #1157, opened 2015-04-13
Topic/1084/remove groups to follow

steventlamb:


jwalgran:
Reviewing

jwalgran:
I registered a new user and was correctly able to RSVP for a training event after the quiz, without needing to visit the groups page.

+1

PR #1158, opened 2015-04-13
Blockface to blockedge

### LINKED ISSUE #1042: Change reservations page URL to /blockedge
The proper terminology is block edge rather than block face so we need to change the URL.

maurizi:
Change blockface to blockedge in URLs and text

Fixes #1042


RickMohr:
+1

PR #1160, opened 2015-04-14
Codify AMI promotion process

hectcastro:
These changes support AMI promotion from one account to another. Sharing is facilitated by giving launch permissions to another account ID. Once shared, the AMI tags are updated to reflect the promotion and avoid any image pruning processes.

### DIFF
--- a/scripts/aws/promote-ami.sh
+++ b/scripts/aws/promote-ami.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -e
+
+if env | grep -q "NYC_TREES_DEPLOY_DEBUG"; then
+  set -x
+fi
+
+function packer_isotime() {
+   date -u +"%Y-%m-%dT%H:%M:%SZ"
+}
+
+if (env | grep -q "NYC_TREES_PROMOTE_AMI" && env | grep -q "NYC_TREES_PROMOTE_ACCOUNT_ID"); then
+    # Grant launch permissions to provided account ID
+    aws ec2 modify-image-attribute \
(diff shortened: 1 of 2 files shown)


scripts/aws/promote-ami.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if env | grep -q "NYC_TREES_DEPLOY_DEBUG"; then
+  set -x
+fi
+
+function packer_isotime() {
+   date -u +"%Y-%m-%dT%H:%M:%SZ"
+}
+
+if (env | grep -q "NYC_TREES_PROMOTE_AMI" && env | grep -q "NYC_TREES_PROMOTE_ACCOUNT_ID"); then
```
kdeloach: So in the scenario where we are deploying an AMI from staging to production, `NYC_TREES_PROMOTE_AMI` represents an AMI ID that exists on the staging account (or whatever account is configured as `AWS_PROFILE`) and `NYC_TREES_PROMOTE_ACCOUNT_ID` represents the AWS User ID for the production account?

kdeloach:
+1

scripts/aws/promote-ami.sh
hectcastro: Correct.

jwalgran:
Looks good to me. Would it make a mess to actually test promote a random AMI from staging?

hectcastro:
We could and then revoke the launch permissions after validating that it actually works.

hectcastro:
Staging after running:

ec2_management_console

Production after running:

ec2_management_console

jwalgran:
Nice. +1

PR #1165, opened 2015-04-14
Add missing migrations

jwalgran:
I could not run ``migrate`` on the develop branch because it reported that there were changes to the models not reflected in the migrations. I ran ``manage.py makemigrations`` and added all the generated migrations.


maurizi:
+1

PR #1168, opened 2015-04-14
Add final borough blockface fixtures

jwalgran:
This commit replaces the chunked blockface fixtures with one fixture per borough, which lines up with how we received the data. I needed to add the final, approved Bronx blockfaces, so I took the opportunity to fix the messy load scripts I initially committed. Adding a ``source`` column to the model is purely a data management helper.

Both the ``reload_dev_data`` management command and the ``setupdb.sh`` deployment script have been updated to load any ``blockface_*.json`` fixtures.

Connects to #849


jwalgran:
Reassigning because Steve is out and we want to get these blocks out to prod today

RickMohr:
Testing now...

RickMohr:
+1 Worked beautifully.

image

PR #1174, opened 2015-04-14
Don't ignore event related HTTP errors on survey pages

kdeloach:
The `_validate_event_and_group` function returns an HTTP error if the
user is not checked into the current event. Consumers of this helper function
erroneously assume that a valid event is returned. The result is that some
survey endpoints can be triggered by users even if they have not been
checked into the event.

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -12,7 +12,7 @@ from celery import chain
 
 from django.conf import settings
 from django.contrib.gis.geos import GeometryCollection
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, PermissionDenied
 from django.core.urlresolvers import reverse
 from django.db import transaction, connection
 from django.db.models import Prefetch
@@ -389,7 +389,7 @@ def blockface(request, blockface_id):
 def _validate_event_and_group(request, event_slug):
     event = get_object_or_404(Event, group=request.group, slug=event_slug)
     if not user_is_checked_in_to_event(request.user, event):
-        return HttpResponseForbidden('User not checked-in to this event')
+        raise PermissionDenied('User not checked-in to this event')
     return event
 
 


maurizi:
I think it would be better to change `_validate_event_and_group` to raise an error instead of returning one

kdeloach:
Raising throws a TypeError "exceptions must be old-style classes or derived from BaseException, not HttpResponseForbidden".

steventlamb:
Ah, the problem with raising is that `HttpResponseForbidden` is not an exception. You can raise `PermissionDenied`. 

https://bitbucket.org/slav0nic/djangobb/pull-request/31/dont-return-httpresponseforbidden/diff
https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.HttpResponseForbidden

You'll lose the error text though, not sure if we're ever using that. Here's an example diff:

```
diff --git a/src/nyc_trees/apps/survey/views.py b/src/nyc_trees/apps/survey/views.py
index 36ed9c9..f32871a 100644
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -11,7 +11,7 @@ from pytz import timezone
 from celery import chain
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, PermissionDenied
 from django.core.urlresolvers import reverse
 from django.db import transaction, connection
 from django.http import (HttpResponse, HttpResponseForbidden,
@@ -347,7 +347,7 @@ def blockface(request, blockface_id):
 def _validate_event_and_group(request, event_slug):
     event = get_object_or_404(Event, group=request.group, slug=event_slug)
     if not user_is_checked_in_to_event(request.user, event):
-        return HttpResponseForbidden('User not checked-in to this event')
+        raise PermissionDenied('User not checked-in to this event')
     return event
 
 
@@ -367,8 +368,6 @@ def start_survey_from_event(request, event_slug):
     group = request.group
     event = _validate_event_and_group(request, event_slug)
 
-    if not isinstance(event, Event):
-        return event
     if not event.in_progress():
         return HttpResponseForbidden('Event not currently in-progress')
 
@@ -421,8 +420,6 @@ def submit_survey(request):
 
 def submit_survey_from_event(request, event_slug):
     event = _validate_event_and_group(request, event_slug)
-    if not isinstance(event, Event):
-        return event
     return _create_survey_and_trees(request, event)
 
 
@@ -514,9 +511,7 @@ def _survey_detail(request, survey_id):
 
 
 def survey_detail_from_event(request, event_slug, survey_id):
-    event = _validate_event_and_group(request, event_slug)
-    if not isinstance(event, Event):
-        return event
+    _validate_event_and_group(request, event_slug)
     return _survey_detail(request, survey_id)
 ```

steventlamb:
+1 works

kdeloach:
Will update this to use `PermissionDenied`.

PR #1175, opened 2015-04-14
Add Windshaft request parsing support

hectcastro:
These changes break individual Windshaft tile requests into specific attributes that are indexed in ElasticSearch for querying.

1. Windshaft emits logs as JSON
2. Beaver picks them up and ships them to Logstash
3. Logstash converts the `timestamp` field into `@timestamp` with the proper formatting
4. Resulting JSON blob is shipped to ElasticSearch

---

discover_-_kibana_4


maurizi:
+1

PR #1179, opened 2015-04-15
Bump pip role version

hectcastro:
This change fixes a scenario during AMI creation where Packer installs the `python-pip` and `python-dev` packages prior to running Ansible because Python is an Ansible dependency. The version of the `python-pip` package installed when no version number is specified (as is the case in the Packer template) conflicts with the version installed by this role because it prompts the user for dependent package downgrades.

See also: https://github.com/azavea/ansible-pip/pull/1


RickMohr:
+1 pip updated when provisioning

PR #1195, opened 2015-04-15
Prevent registration of usernames that conflict with URLs

### LINKED ISSUE #1171: Prevent using usernames that conflict with URLs, or make sure there are no conflicting URLs
It is currently possible to create a user with a username of `profile`.

If you login as that user and try to go to your user profile, you'll go into a redirect loop.

The culprit is here: https://github.com/azavea/nyc-trees/blob/a3f503f817c43df6d4fc2c4729942de7da48b275/src/nyc_trees/apps/users/urls/user.py#L12-L16

We should either 1. prevent usernames of `profile`, `achievements`, and `settings`, or 2. Change our URL patterns so using a username as a slug is unambiguous

kdeloach:
Add form validation error to prevent registration of usernames that
would conflict with any of our special `/user/*` URLs.

Fixes #1171

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -100,6 +100,12 @@ class User(NycModel, AbstractUser):
                       'contributions_are_public': 'contributions',
                       'achievements_are_public': 'achievements'}
 
+    # Please keep in sync with special `/user/*` endpoints.
+    # Ref: src/nyc_trees/apps/users/urls/user.py
+    reserved_usernames = ('profile',
+                          'settings',
+                          'achievements')
+
     def clean(self):
         if ((User.objects.exclude(pk=self.pk)
              .filter(email__iexact=self.email).exists())):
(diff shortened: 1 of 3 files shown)


src/nyc_trees/apps/login/forms.py
```diff
@@ -39,6 +40,12 @@ class NycRegistrationForm(RegistrationFormUniqueEmail):
         required=False
     )
 
+    def clean_username(self):
+        if self.cleaned_data['username'].lower() in User.reserved_usernames:
+            raise ValidationError('This username is invalid because it '
```
jwalgran: This is a little specific and jargony for for a user facing error. "This username is already in use" is a white lie, but, perhaps, a bit friendlier.

kdeloach: That thought crossed my mind too, will update.

kdeloach: Settled on "This username is not available."

src/nyc_trees/apps/core/models.py
```diff
@@ -100,6 +100,12 @@ class User(NycModel, AbstractUser):
                       'contributions_are_public': 'contributions',
                       'achievements_are_public': 'achievements'}
 
+    # Please keep in sync with special `/user/*` endpoints.
+    # Ref: src/nyc_trees/apps/users/urls/user.py
+    reserved_usernames = ('profile',
+                          'settings',
+                          'achievements')
```
kdeloach: Should this be a module-level constant?

maurizi:
+1

Issue #952, opened 2015-03-26
Mixed content warning trying to edit event location via admin


hectcastro:
This prevents me from seeing whatever control is associated with the `Location` field on the event admin page:

```
Mixed Content: The page at 'https://treescount.azavea.com/admin/event/event/29/' was loaded over HTTPS, but requested an insecure script 'http://openlayers.org/api/2.13/OpenLayers.js'. This request has been blocked; the content must be served over HTTPS.
```

PR #1206, opened 2015-04-16
Load admin page OpenLayers from https CDN

jwalgran:
The Django admin has a built in slippy map for editing geometries, built
on top of OpenLayers. By default, the admin template loads OpenLayers
from http://openlayers.org/api/2.13/OpenLayers.js, which fails with an
"insecure" error when the site is running under https.

This commit overrides that insecure URL with a secure one from cdnjs.

Connects to #952


maurizi:
+1

jwalgran:
Moving to ``develop``

PR #1207, opened 2015-04-16
Replace variable

jfrankl:
I accidentally removed the correct variable ```counts.tree``` with the integer ```3``` in #1155. Did not mean to do that.


jwalgran:
+1 fixed

PR #1211, opened 2015-04-16
Fix sizing on download button

jfrankl:
Was falling off the page like this:

image

Now it's looking sharp like this:

image


maurizi:
+1

PR #1214, opened 2015-04-16
Fix ios8 scrolling glitch on map page

jfrankl:
I noticed this glitch when testing the map pages on iOS 8. The page height was expanding beyond the height of the screen, making the page scrollable, which we don't want.

Before video:

http://sandbox.azavea.com/temporary/nyc-ios8-scroll-before.mov

After video:

http://sandbox.azavea.com/temporary/nyc-ios8-scroll-after.mov


designmatty:
+1 good find

PR #1215, opened 2015-04-16
Configure git to treat PDF files as binary

jwalgran:
If we don't do this, git may try to convert line endings in the file and corrupt it.


steventlamb:
as per @maurizi's instructions, I have opened https://github.com/azavea/nyc-trees/pull/1216 to do this but also fix the PDF with the original source PDF. I think we should close this and merge that

jwalgran:
Closed in favor of #1216

PR #1218, opened 2015-04-16
Redirect non-mappers to Reservations Instructions page

### LINKED ISSUE #1198: 403 forbidden if try to access reservations page without individual mapper privileges
Users should be able to view the page but not make reservations.

kdeloach:
When non individual mappers try to access the Reservations page, they
will be redirected to the reservations instructions page.

Fixes #1198

### DIFF
--- a/src/nyc_trees/apps/home/routes.py
+++ b/src/nyc_trees/apps/home/routes.py
@@ -21,3 +21,7 @@ retrieve_job_status = do(v.retrieve_job_status)
 individual_mapper_instructions = individual_mapper_do(
     render_template('home/individual_mapper_instructions.html'),
     v.individual_mapper_instructions)
+
+trusted_mapper_request_sent = individual_mapper_do(
+    render_template('home/trusted_mapper_request_sent.html'),
+    v.trusted_mapper_request_sent)
--- a/src/nyc_trees/apps/home/templates/home/individual_mapper_instructions.html
+++ b/src/nyc_trees/apps/home/templates/home/individual_mapper_instructions.html
@@ -1,9 +1,5 @@
 {% extends "base_two_frames.html" %}
 
-{% block nav %}
-{% include "core/partials/nav.html" with active_page="about" %}
-{% endblock %}
-
 {% block aside %}
 <h2>Mapper Status</h2>
 {% endblock aside %}
--- a/src/nyc_trees/apps/home/templates/home/trusted_mapper_request_sent.html
+++ b/src/nyc_trees/apps/home/templates/home/trusted_mapper_request_sent.html
@@ -1,0 +1,17 @@
+{% extends "base_two_frames.html" %}
+
+{% block aside %}
+<h2>Mapper Status</h2>
+{% endblock aside %}
+
+{% block main %}
+
+<p>
+    Your request for individual mapper status has been sent!
+</p>
+<p>
+    Once approved by a group admin, you will receive an email confirmation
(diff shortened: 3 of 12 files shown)


src/tiler/sql/user_reservable.sql
```diff
@@ -46,4 +62,7 @@
     ON (turf.group_id = trustedmapper.group_id
         AND trustedmapper.user_id = <%= user_id %>
         AND trustedmapper.is_approved)
+  LEFT JOIN core_user AS individual_mapper
+    ON (individual_mapper.id = <%= user_id %>
+        AND COALESCE(individual_mapper.individual_mapper, FALSE))
```
maurizi: Does this Model need to be added to the cache buster?

kdeloach: In my testing, the available blockfaces on the reservables page updated immediately as I toggled the Individual Mapper flag on & off. I'll add the User table to the reservable cache buster for good measure.

kdeloach:
Added user model to cache buster.

kdeloach:
@jwalgran This could use some more eyes.

jwalgran:
The most I can give you is 2

dboyer:
Doesn't a user have to have individual mapper status before they can request to map in a group's area? I don't see the request individual mapper status button on a group's detail page unless I'm already an individual mapper.

kdeloach:
Yes, you must be an individual mapper before requesting group mapper status. The only way to become an individual mapper is by clicking a button on your dashboard/profile once you have fulfilled certain requirements.

Maybe I overcomplicated this though. It seems like instead of showing the Reservations map for non-mappers, we should just immediately redirect to the Reservations Instructions page? We could also add a "Become a trusted mapper" button on the group detail page which links to the instructions page. What do you think?

maurizi:
> It seems like instead of showing the Reservations map for non-mappers, we should just immediately redirect to the Reservations Instructions page? We could also add a "Become a trusted mapper" button on the group detail page which links to the instructions page. What do you think?

+1 to this

dboyer:
@kdeloach Parks approved redirecting to the reservations instructions page when a non-individual mapper clicks the reservations link in the navigation. Non-individual mappers don't need to see the reservations map. 

The group detail includes a request individual mapper status within the group area button if that group allows individual mappers.

kdeloach:
Thanks, that makes sense to me. I'll make these changes tomorrow.

kdeloach:
Updated. Will reassign once tests have passed.

maurizi:
@kdeloach You may want to update the PR title now.

maurizi:
I'll look at this shortly

maurizi:
@kdeloach It looks like there is some lint in this PR.

maurizi:
I tried to go to the reservations map as a user w/o individual mapper status and got redirected to the individual mapper instructions page.
When I set the individual mapper flag to True I was able to get to the reservations page.

The process for requesting trusted mapper status mostly worked great.
The other thing weird was that after I requested trusted mapper status, when I went back to the group detail page there was no change indicating that a request had been sent, and I could continue to request trusted mapper status.

However, I think that is a small issue that doesn't need to be fixed right now.

+1

kdeloach:
Fixed lint.

kdeloach:
Once you are approved as a trusted mapper, that button will actually forward to the Reservations page instead of submitting another request (not that that is any less confusing).

PR #1221, opened 2015-04-16
Update training material for four users states

### LINKED ISSUE #589: Training page for anonymous users should show some text and link to PDF training materials
We are restricting the interactive training materials to registered users, but Parks will is developing a PDF version of the training that we will need to make available to the unregistered public.

steventlamb:
fixes #589 on github

finished, in progress, not started, not logged in

here's how this will look.

First you come as an anonymous user:
screenshot from 2015-04-16 17 12 58
then you log in:
screenshot from 2015-04-16 17 12 11
then you start online training:
screenshot from 2015-04-16 17 12 41
then you finish:
screenshot from 2015-04-16 17 09 21


steventlamb:
please disregard last commit, I'm going to update the test harness instead of loosening it.

jwalgran:
I registered a new user walked though the entire training process and saw all 4 messages.

+1

dboyer:
That looks like the right content for the right log ins.

PR #1222, opened 2015-04-17
Fix glitch on Android 4.2 and earlier

jfrankl:
This fixes a glitch with the default browser on Android 4.2 and earlier that caused elements with ```position: fixed``` and ```transform: translate``` to not register the transform.

Here is a stackoverflow discussion about the glitch. http://stackoverflow.com/questions/18572488/transform-doesnt-work-on-position-fixed-divs-on-android

Using ```transform: translate``` to reposition elements—in our case, the side navigation and header when the navigation opens—is generally more performant than alternatives like ```left``` or ```margin-left```. However, in this case, I've switched to ```left```, which works on Android 4.2 and earlier.

I've tested the basic page and map page layouts on latest Chrome, latest Firefox, IE9, Android 4.0, iOS7, and iOS8. All check out. More testing to come.

### DIFF
--- a/src/nyc_trees/sass/partials/_footer.scss
+++ b/src/nyc_trees/sass/partials/_footer.scss
@@ -1,6 +1,9 @@
 footer {
   background-color: $gray-lightest;
   padding: 2rem 0;
+  transition: all 400ms $transition;
+  position: relative;
+  left: 0;
   @media (min-width: $screen-md) {
     padding: 5rem 0 4rem;
   }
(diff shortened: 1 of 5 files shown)


jfrankl:
*Before video*

Note that the side menu is open when a page loads instead of properly hidden. In addition, clicking the hamburger menu has no effect:

http://cl.ly/0b0B13092W1P

*After video*

Note that the side menu now opens properly. Hooray!

http://cl.ly/022g1g313r2U

designmatty:
+1

PR #1223, opened 2015-04-17
Fix circle size on mobile

### LINKED ISSUE #1213: Adjust circle sizes or text
Text can overflow on mobile:

image

jfrankl:
Reduce padding between circles on mobile and reduce margin on left and right of container. Fixes #1213.

image


maurizi:
+1

PR #1235, opened 2015-04-17
Make actionbar invisible when empty

### LINKED ISSUE #721: Add class on body when actionbar is open
I would like for a class to be added to the body element when data is sent to the ```.actionbar``` element for the first time on a page. For example, when someone loads the Progress Map, this class should not be added to ```body```. When they click on a blockface and data is sent to ```.actionbar```, we should add ```.actionbar-triggered``` to the ```body``` element. We don't need to remove this class.

This will allow me to hide the actionbar until it is useful, at which point the map will slide up to show the actionbar. This will serve a helpful double function of removing unneeded white space and drawing attention to that area when it becomes relevant.

jfrankl:
Address but does not fix the issue #721.

Removes the border on the action-bar when it's empty. Slightly better on desktop, but doesn't impact mobile.

Before:

image

After:

image

### DIFF
--- a/src/nyc_trees/sass/partials/_map.scss
+++ b/src/nyc_trees/sass/partials/_map.scss
@@ -2,7 +2,7 @@
   padding-bottom: 0 !important;
   .tab-status-map-tab &, .page-new-reservation & {
     position: static;
-    height: 7rem;
+    height: 7.1rem;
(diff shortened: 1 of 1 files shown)


maurizi:
+1

PR #1240, opened 2015-04-17
Census admin page for uploading group polygons GeoJSON file

### LINKED ISSUE #1134: Allow census admins to create/modify group polygons by uploading a GeoJSON file
When Parks uploads a group polygon GeoJSON file:
* Each polygon must have a group_name attribute, giving the name of an active group
* If we find an error with any polygon, the upload will be rejected
* All uploaded polygons will be immediately set on their groups (no undo)
* After a successful upload we'll show polygon status for each group (polygon updated, polygon unchanged, has no polygon)

RickMohr:
* If any group names are unknown or there is a syntax error, update nothing and report error info
* Otherwise replace the `group.border` multi-polygon for each group present in the upload, and show a summary of changes

File format based on example from Parks (simplified). 
Note that this is not valid GeoJSON (lacking `type` attributes), but we can easily parse the parts we need.
```json
{
    "features": [
        {
            "geometry": {
                "rings": [
                    [
                        [-74.00433784505003, 40.66953878784512],
                        [-74.00321110794214, 40.67192229492897],
                        [-74.00289344462487, 40.67259424901714],
                        [-74.00433784505003, 40.66953878784512]
                    ]
                ]
            },
            "attributes": {
                "GroupName": "Midnight Mappers",
            }
        },
        {
            "geometry": {
                "rings": [
                    [
                        [-73.90236842648156, 40.88726229418421],
                        [-73.90050742517315, 40.88673329379011],
                        [-73.89791910153626, 40.886186228573074],
                        [-73.90236842648156, 40.88726229418421],
                    ]
                ]
            },
            "attributes": {
                "GroupName": "Fledglings",
            }
        }
    ]
}

```
Fixes #1134


RickMohr:
I used this file for a successful upload:
`//fileshare/users/rmohr/notes/nyc/polygonUpload/real.json`

Here is the actual file from parks (pretty-printed), which causes many unknown group errors:
`//fileshare/users/rmohr/notes/nyc/polygonUpload/full.json`

jwalgran:
Since we cannot process GeoJSON files, it would be a good idea to change "Upload a GeoJSON file" instruction to something else.

src/nyc_trees/apps/census_admin/views.py
```diff
@@ -3,33 +3,71 @@
...
+def _update_group_polygons(data, group_data):
+    unknown_group_names = set()
 
-def admin_training_material_list(request):
-    # TODO: implement
-    pass
+    try:
+        for feature in data['features']:
+            group_name = feature['attributes']['GroupName']
+            polygon_points = feature['geometry']['rings'][0]
```
jwalgran: Did we discuss, with Parks, the fact that we are not supporting polygons with holes? I don't want them to assume that it will work.

RickMohr:
@jwalgran, Parks created their sample file by a GeoJSON export add-on for ArcMap. As far as they are concerned it is GeoJSON, and I'm afraid we would confuse them by calling it something else, or complaining that it is not valid GeoJSON.

src/nyc_trees/apps/census_admin/views.py
RickMohr: We didn't, and given what they're doing I would expect them to make any holes. @dboyer, want to check with them?

src/nyc_trees/apps/census_admin/views.py
```diff
@@ -3,33 +3,71 @@
...
 
-def start_admin_surveys_job(request):
-    # TODO: implement
-    pass
 
+def upload_group_polygons(request):
+    file_obj = request.FILES['file']
+    file_obj.seek(0)
+    text = file_obj.read()
+    data = json.loads(text)
```
jwalgran: To avoid raising a 500 if the JSON is malformed, it would be good to wrap this in a try and raise a 400. I don't think we need to go through the trouble of handing the error properly on the client, but we will be treating 500s that show up in our production logs as "red alerts"

jwalgran: I just saw that we already report other types of errors, so let's do that instead of raising 400.

jwalgran:
I was able to upload single and multiploygons and verify them on the group admin page OpenLayers map

+1

src/nyc_trees/apps/census_admin/views.py
dboyer: I'll let them know that we can't support polygons with holes. I doubt groups would be encouraged to have that type of census zone anyway.

PR #1245, opened 2015-04-17
Use location data from form, not event object.

### LINKED ISSUE #638: Event creation - event location is lost after failed validation
How to reproduce:
 - Go to the event creation page
 - Set the event location on the map, but leave some other fields empty or incorrect
 - Attempt to save
 - Fix your errors on non-location fields
 - Attempt to save
 - There will now be an error for the `geom` field.

maurizi:
Fixes #638

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/partials/event_form.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/event_form.html
@@ -62,8 +62,8 @@
             </div>
         </div>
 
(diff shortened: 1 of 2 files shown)


jwalgran:
Looks good. Testing now.

jwalgran:
+1 I created an event with a date in the past, in order to trigger a post back with server side validation. My location selection was not lost during the round trip.

PR #1246, opened 2015-04-17
Don't show the privacy button in the nav if the user is a minor

### LINKED ISSUE #646: Hide privacy button if user is a minor
We hide the privacy panel in the user profile page if the user is a minor, but we still show it in the side nav.

image

maurizi:
Fixes #646


kdeloach:
+1

PR #1248, opened 2015-04-19
Enable Windshaft profiler in development; add JMeter test plans

hectcastro:
These changes add Ansible variables for enabling the Windshaft profiler. By default, the profiler will be enabled in development and off in all other environments.

What does that mean? That means that Windshaft will emit more detailed metrics to the configured StatsD endpoint. In our case, we run Statsite (a C port of StatsD). Statsite accepts these metrics and flushes them to Graphite, which has a UI available on the `services` virtual machine on port `8080`.

In addition to the Windshaft profiler hooks, a new top level directory has been added with Apache JMeter test plans. Each test plan includes a series of requests that can be replayed with JMeter to simulate load against the development environment. See the included `README` for more details on how to maneuver JMeter.

PR #1249, opened 2015-04-19
Upgrade Windshaft to version 0.36.0

hectcastro:
**This pull request is not to be merged.**

These changes upgrade Windshaft from 0.19.4 to 0.36.0. Windshaft was not upgraded to the most recent release because it appears the configuration changed it some way (relative to the configuration we're currently using).

There was also a notable battle with NPM in the upgrade process. See https://github.com/npm/npm/issues/6556 for more details.

---

Performance didn't get drastically better with 0.36.0, so I don't think upgrading right now is worthwhile. Upgrading could get more attractive if there are configuration parameters in 0.36.0 that enable tuning not exposed in 0.19.4. I have not found any as of yet.

Below is a comparison of the JMeter test plan (https://github.com/azavea/nyc-trees/pull/1248) run against 0.19.4 and 0.36.0:

pasted_image_4_19_15__15_13

### DIFF
--- a/deployment/ansible/roles/nyc-trees.tiler/tasks/dependencies.yml
+++ b/deployment/ansible/roles/nyc-trees.tiler/tasks/dependencies.yml
@@ -1,5 +1,15 @@
 ---
+- name: Install canvas rendering dependencies
+  apt: pkg={{ item }} state=present
+  with_items:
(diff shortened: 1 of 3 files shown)


maurizi:
@hectcastro Can this be closed?

PR #1263, opened 2015-04-20
Fix privacy button visibility unit test

jwalgran:
Layout changes have now reduced the number of privacy links that appear on the page.


maurizi:
+1

PR #1264, opened 2015-04-20
Change group tree count from circle to ticker.

### LINKED ISSUE #1208: Replace trees progress circle with ticker on group page
Currently, the group page shows the number of trees counted as a circle. This should be switched to the counter to be consistent with the rest of the site. I get an error when I try to drop in the ticker and switch in the appropriate variable, so I'll need a developer to add it.

Current page:

image

The top left circle should be removed. The ticker should be displayed below the circles on its own row, taking up the full width of the element:

image

maurizi:
Fixes #1208


jfrankl:
+1 well done

image

maurizi:
@azavea-bot rebuild

PR #1283, opened 2015-04-20
Comment-out "My Progress" option on progress page drop-down

### LINKED ISSUE #1276: Disable the "My Progress" option on the progress map until the first mapping events
Drawing the citywide progress map for a specific user puts some load on the tiler but does not draw unique tiles. There will be no personalized progress until events start in May.

RickMohr:
Fixes #1276


maurizi:
+1

PR #1284, opened 2015-04-20
Topic/dont version training pdf

steventlamb:


maurizi:
I feel very wary about including commit caa142a right now.
It's a semantically small change, but affects every page on the site, and I'm worried a file will have been missed somewhere.

steventlamb:
I don't have a problem dropping caa142a, and I will fix the urljoin thing.

jwalgran:
Looks like this may still be in the oven. Toss it back when it is ready.

steventlamb:
@jwalgran closing this and assigning to the lighter version to you.

PR #1286, opened 2015-04-20
Cleanup event title and slug validation

### LINKED ISSUE #409: Update event slug when switching from public -> private or private -> public

### LINKED ISSUE #628: Add event name validation
You shouldn't be able to create an event with a name composed of all blank spaces. 

Also, if you create an event name with symbols, it throws an error and the group detail page stops working.

We should verify that the generated slug is URL friendly. The regex validation should match the slug definition used throughout our URLs `[\w-]+`.

image

maurizi:
Fixes #628
Fixes #409

### DIFF
--- a/src/nyc_trees/apps/event/forms.py
+++ b/src/nyc_trees/apps/event/forms.py
@@ -57,6 +57,13 @@ class EventForm(ModelForm):
         right_now = now().astimezone(est_tz)
         today = right_now.date()
 
+        # If the Event title or is_private fields change, we need to clear
+        # the slug field to let the model's .clean() recreate it.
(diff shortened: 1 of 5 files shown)


kdeloach:
I noticed one regression related to #638. When I submit the form with an invalid Title, the Address Map lat/lng is empty.
image

maurizi:
@kdeloach I just rebased this off of develop, I think the fix for #638 should be on this branch now

kdeloach:
I think we also need to ensure that the event title is URL safe. I get exceptions after adding an event named `?foo=bar&baz=thing`.

kdeloach:
+1 Title validation on adding/editing events works.

PR #1288, opened 2015-04-20
On events page, swap order of training and mapping tabs

### LINKED ISSUE #1277: On events page, move training tab before mapping tab
The order should be All, Training, Mapping

RickMohr:
Fixes #1277


steventlamb:
+1

PR #1292, opened 2015-04-20
Bypass CloudFront origin swap; use public tiler DNS record

hectcastro:
This changeset aims to avoid CloudFront distribution origin changes by using a public DNS record for the tiler (tiles.treescount.foo.com). The private DNS record used by the application server to discover the tile server remains untouched. In addition, the CloudFront distribution endpoint (with HTTPS) will continue to be used to serve tiles because we only have an SSL certificate for the root domain.


maurizi:
Looks good to me

PR #1294, opened 2015-04-21
Add activation email resend workflow

### LINKED ISSUE #647: When attempting to log in before activation, show a more instructive message
Preferably with a link to resend the confirmation email.

screen shot 2015-02-19 at 10 03 24 am

jwalgran:
If a person has registered, but not activated their account, and they attempt to log in, we want to tell them why we can't yet log them in, and provide them a way to request another activation email.

This commit adds a workflow, similar to "forgot username," for resending activation emails. There are a few key differences.

  * If the account is already active, there is no token, so we cannot
    send an email and, instead, prompt the person to log in.

  * Adding an anchor tag to the inactive account validation message
    requires overriding the authentication form, so that the HTML
    content can be marked as "safe."

Test by registering a new account, and then attempting to log in before activation.

Fixes #647

### DIFF
--- a/src/nyc_trees/apps/login/templates/login/send_activation_email.html
+++ b/src/nyc_trees/apps/login/templates/login/send_activation_email.html
@@ -1,0 +1,34 @@
+{% extends "../templates/registration/registration_base.html" %}
+{% load i18n %}
+{% load utils %}
+
+{% block registration_title %}Send Activation Email{% endblock %}
+
+{% block registration_content %}
+<div class="content login-signup">
+  <div class="well login-signup-panel">
+    <form method="POST" id="send-activation-email-form">
+      <fieldset>
+        {% csrf_token %}
+        <p>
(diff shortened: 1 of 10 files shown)


kdeloach:
Nice. This actually happened during our user testing, at least once.

src/nyc_trees/apps/login/templates/login/send_activation_email.html
```diff
@@ -0,0 +1,34 @@
...
+{% block registration_title %}Send Activation Email{% endblock %}
+
+{% block registration_content %}
+<div class="content login-signup">
+  <div class="well login-signup-panel">
+    <form method="POST" id="send-activation-email-form">
+      <fieldset>
+        {% csrf_token %}
+        <p>
+            Need us to send you and account activation email? Enter the Email address you used to register
```
maurizi: Typo here: `and` -> `an`

steventlamb:
+1, tested, works

PR #1303, opened 2015-04-21
Add JMeter test plan for My Blockedge Progress

hectcastro:
This changeset adds a JMeter test plan for the My Blockedge Progress drop down on the progress map. This test plan is very similar to the All Users Blockedge Progress test plan, except that it includes a Random Variable for user ID. This user ID is appended to request paths as a query string: `?user=${RANDOM_USER_ID}`.

The random user ID is per thread, so if you try to simulate 10 users (the default), each will carry its random user ID until all requests are processed.


RickMohr:
+1 Runs, appending random user parameter.

PR #1308, opened 2015-04-21
Make training text the same for anon and new users

### LINKED ISSUE #589: Training page for anonymous users should show some text and link to PDF training materials

steventlamb:
fixes #589 on github


RickMohr:
+1 Somebody at Parks really really really likes exclamation points.

PR #1311, opened 2015-04-21
Hide circumference label until status is chosen

### LINKED ISSUE #943: Change circumference label
We're asking people to input circumference for trees (alive or dead) and diameter for stumps. By default, the area under Tree Trunk shows as Circumference. It doesn't change to Diameter unless Stump is selected.

There's some concern that people will see the circumference prompt and start measuring a stump's circumference before they select "stump" under Tree Structure and realize it should be diameter.

Change that section to just show a header labeled "Tree Trunk/Stump Size" and don't display the circumference or diameter prompt and entry form until after someone has selected the tree structure.

@designmatty Unassign yourself please if this is not a designer issue.

steventlamb:
fixes #943 on github


steventlamb:
As per discussion with @maurizi, this *mostly* solves the problem of #943 but isn't quite done because we need to discuss #943 and how it interacts with other issues that came out of user testing. We should leave this PR open for the time being and discuss when able, probably after our upcoming launch.

kdeloach:
Still on hold?

PR #1313, opened 2015-04-21
Tiler performance improvements

maurizi:
- Removes unnecessary JOIN from progress layers, by querying for that information in the progress blockface popup view
 - Denormalizes latest survey information to speed up tiler performance

EXPLAIN ANALYZE showed that our current query for the progress layer
spent almost half of the time sorting the blockface table, which was
required to use DISTINCT ON (blockface.id) to use only the most recent
survey row.

I compared performance with JMeter between our current progress layer, a "baseline"
version which just selected all items from the blockface table as unmapped,
and a denormalized version with the latest survey id as part of the
blockface table.

The timing information is as follows:

From my workstation at home:

Version  | Average ms. per tile 
------------- | -------------
Current      | 767.313
Baseline     | 18.625
Denormalized | 193.667


maurizi:
Explain Analyze for each query:
Current: http://explain.depesz.com/s/Clt
Baseline: http://explain.depesz.com/s/CIn
Denormalized: http://explain.depesz.com/s/8Ew

src/tiler/sql/user_progress.sql
```diff
@@ -1,16 +1,11 @@
--- The 'DISTINCT ON' and 'ORDER BY' clause are needed so that we only use
--- the latest survey for a blockface
 (SELECT
-  DISTINCT ON (block.id)
-  block.geom, block.id, turf.group_id,
+  block.geom, block.id
```
hectcastro: Missing a trailing comma here.

maurizi: Thanks, fixed

maurizi:
Note to self: This needs updated cache busters

hectcastro:
Preliminary results from the Windshaft profiler and JMeter:

image

`all` = All Progress
`my` = My Progress

PostgreSQL slow query log was still tripped a bunch of times by the tweaked queries in this PR. Still digging into those.

jwalgran:
I loaded the progress page then clicked zoom in three times, then repeated the load and zoom

screen shot 2015-04-21 at 11 12 13 am

maurizi:
@jwalgran @hectcastro I ran the same JMeter tests again on my workstation at Azavea (the ones shown previously were from my workstation at home)

From my workstation at work:
Version  | Average ms. per tile 
------------- | -------------
Current      | 1382.979
Baseline     | 410.875
Denormalized | 1086.895

It looks like the performance gains were more pronounced on my home machine :crying_cat_face:

That said there is still a measureable performance improvement with the newer query, and I think this is as optimized of a DB query as we can hope to get.

maurizi:
@jwalgran I fixed the issue you noted and information about the change to the cache buster logic to the commit message.

maurizi:
@jwalgran @hectcastro This has been rebased against develop, and an index was added to `latest_survey`, which *I think* should fix the massive performance degradation seen when using the "My Progress" layer with a metatile of 1

maurizi:
@jwalgran Any objection to merging this?

jwalgran:
Since we are planning to meet and develop a holistic strategy for tiler updates, I would rather wait until then.

maurizi:
Seems like this is not the way we're going.  Closing this.

Issue #1275, opened 2015-04-20
Text for training content on dashboard


dboyer:
If the user hasn't completed training, training content is visible on their dashboard. That content should be updated to say:

Heading - You're almost there! Get started on your training and help us make trees count!

Body above the different training sections - Welcome to the TreesCount! 2015 online training! Once you complete these modules, sign up for a training event to use what you've learned to map trees!

@jfrankl What are your thoughts on moving the training section above the progress counts since it will be the primary thing non-trained users need to do. Is that possible?

jwalgran:
@jfrankl I was easily able to move the training section to the top and make it the full width of the page. If you are into it, I can leave the change in.

![screen shot 2015-04-21 at 9 34 21 am](https://cloud.githubusercontent.com/assets/17363/7257332/d65e9df0-e809-11e4-9b31-10f21f36d490.png)

jfrankl:
+1

On Tue, Apr 21, 2015 at 12:36 PM, Justin Walgran <notifications@github.com>
wrote:

> @jfrankl <https://github.com/jfrankl> I was easily able to move the
> training section to the top and make it the full width of the page. If you
> are into it, I can leave the change in.
>
> [image: screen shot 2015-04-21 at 9 34 21 am]
> <https://cloud.githubusercontent.com/assets/17363/7257332/d65e9df0-e809-11e4-9b31-10f21f36d490.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/azavea/nyc-trees/issues/1275#issuecomment-94864254>.
>



-- 
Jeff Frankl, UI/UX Designer

Azavea  |  340 N 12th St, Suite 402, Philadelphia, PA
Web azavea.com  |  Blog azavea.com/blogs  |  Twitter @azavea
<http://twitter.com>

PR #1315, opened 2015-04-21
Update training copy and move to top of the dash

jwalgran:
This commit changes the heading and copy of the training section and moves it to the top of the dashboard, so it is the first thing a new user sees when logging in.

First login:

screen shot 2015-04-21 at 9 55 11 am

After some training:

screen shot 2015-04-21 at 9 56 01 am

Training complete:

screen shot 2015-04-21 at 9 56 24 am

Connects to #1275


jwalgran:
Will reopen to get a connection to the issue

PR #1316, opened 2015-04-21
Add group.territory_updated_at field and use for territory cache buster

### LINKED ISSUE #791: Tile cache busters do not change when rows are deleted
This was noticed when testing the Group Admin territory page.  If you only delete territory, and then save, the blockfaces remain red (group territory) when they should change to grey-ish (available).

Because we rely on the `updated_at` field for cache-busting, we only change the cache buster when items are updated or inserted, not when they are removed.

@RickMohr suggests making the cache buster be a combination of row count and maximum `updated_at`

maurizi:
Add `group.territory_updated_at` field and use for territory cache buster

Ensures that the cache buster is updated when blockfaces are unassigned
from a group's territory without any other changes taking place.

To test, remove a blockface from a group's territory and save without
making any other changes.  The color of the removed blockface should
change to indicate that it is now unavailable.

Fixes #791

### DIFF
--- a/src/nyc_trees/apps/core/migrations/0025_group_territory_updated_at.py
+++ b/src/nyc_trees/apps/core/migrations/0025_group_territory_updated_at.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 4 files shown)


RickMohr:
+1 Looks good, and works.

PR #1318, opened 2015-04-21
Add index for blockface cache buster

### LINKED ISSUE #1242: Survey blockface sort by updated timestamp does not use an index
The following query does not appear to be using an index to sort by the `updated_at` timestamp:

```sql
SELECT "survey_blockface"."id", "survey_blockface"."created_at", "survey_blockface"."updated_at", "survey_blockface"."geom", "survey_blockface"."is_available", "survey_blockface"."expert_required", "survey_blockface"."source" FROM "survey_blockface" ORDER BY "survey_blockface"."updated_at" DESC LIMIT 1
```

Here is the `EXPLAIN ANALYZE` output:

```
Limit  (cost=7518.37..7518.37 rows=1 width=120) (actual time=274.048..274.049 rows=1 loops=1)"
  ->  Sort  (cost=7518.37..7839.76 rows=128558 width=120) (actual time=274.045..274.045 rows=1 loops=1)"
        Sort Key: updated_at"
        Sort Method: top-N heapsort  Memory: 25kB"
        ->  Seq Scan on survey_blockface  (cost=0.00..6875.58 rows=128558 width=120) (actual time=0.008..137.075 rows=128558 loops=1)"
Total runtime: 274.079 ms"
```

maurizi:
Builds on top off #1313 since that also had migrations for the `Blockface` model.  Only commit 98a1344 is part of this PR.

This should fix the queries that were triggering the PostgreSQL slow query logger, which occurred on every map page when getting the cache buster value for the blockface table.

Fixes #1242


maurizi:
I'm working on fixing the unit tests for this now

maurizi:
Superseded by #1365

PR #1319, opened 2015-04-21
Add 'Wrapping Up' section to training

steventlamb:


jwalgran:
I walked through training as a new user and stepped through the new Wrapping Up section

+1

PR #1322, opened 2015-04-21
Move training to top of the dash

### LINKED ISSUE #1275: Text for training content on dashboard

jwalgran:
This commit changes the heading and copy of the training section and moves it to the top of the dashboard, so it is the first thing a new user sees when logging in.

First login:

screen shot 2015-04-21 at 9 55 11 am

After some training:

screen shot 2015-04-21 at 9 56 01 am

Training complete:

screen shot 2015-04-21 at 9 56 24 am

Fixes #1275

### DIFF
--- a/src/nyc_trees/apps/home/templates/home/partials/dashboard.html
+++ b/src/nyc_trees/apps/home/templates/home/partials/dashboard.html
@@ -27,6 +27,17 @@
         <div class="row">
             <main class="col-sm-12 main-dashboard">
                 {% flag full_access %}
+
+                    {% if not user.online_training_complete %}
+                        <div class="home-container">
+                            <section class="trainings">
+                                <div class="home-container">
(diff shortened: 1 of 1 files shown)


jwalgran:
cc @dboyer to verify text changes

steventlamb:
This works, I tested it with and without killing the changes to the training partial.

jwalgran:
@steventlamb rebased out the changes to the training partial

steventlamb:
+1 tested and works

PR #1324, opened 2015-04-21
Updates to census admin pages

### LINKED ISSUE #1297: Change text on page for assigning zones to groups
Text at /census_admin/territory/ should be changed to:

Heading - Partner Group Census Zones

Body - Select a partner group to modify its census zone. Add and remove blocks by clicking them or by drawing an area. 

Is there anyway the URL can change to /censuszone/ ?

### LINKED ISSUE #1298: Yellow block edges on partner zone assignment page on production
The /census_admin/territory/ page on production includes yellow line segments. What do these correspond to? If they're unavailable segments, shouldn't they be a different color?

### LINKED ISSUE #1307: Census admin upload group polygons -- don't require all group names to match
Parks will often have groups in their polygon file that have not yet been created in the database. We will upload polygons for groups that match, and list groups that don't match.

RickMohr:
Fixes #1307 
Fixes #1297 
Fixes #1298

### DIFF
--- a/src/nyc_trees/apps/census_admin/templates/census_admin/upload_group_polygons.html
+++ b/src/nyc_trees/apps/census_admin/templates/census_admin/upload_group_polygons.html
@@ -24,7 +24,7 @@
         </p>
     </form>
 
-    {% if request.method == 'POST' and not updates %}
+    {% if not success %}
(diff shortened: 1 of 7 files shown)


RickMohr:
waffle

maurizi:
+1

PR #1326, opened 2015-04-21
Fix typo in layer_context.py causing exception on progress page

maurizi:


jwalgran:
+1 confirmed this prevents the 500 on the progress page

PR #1332, opened 2015-04-21
Show blank event map when there are no future events

### LINKED ISSUE #1329: Events map does not work when there are no future events
It's clear what should happen on the map tab when there are no future events.  Perhaps we should just hide the "map" tab?

This is what it looks like now:
events-map

maurizi:
Fixes #1329


steventlamb:
+1, tested

PR #1333, opened 2015-04-21
Prevent scrolls being swallowed up by leaflet on Chrome for Android

### LINKED ISSUE #631: Leaflet swallows touch events on group detail territory map
On a touch device, if you try to do a drag scroll with your finger starting on the group detail map, the page does not scroll.

maurizi:
Fixes #631


kdeloach:
+1 Works. FYI there's a typo in the commit description which might interfere with GitHub/Waffle.

PR #1369, opened 2015-04-22
Update training content in version control

steventlamb:


steventlamb:
@dboyer @designmatty I had to do a bit of markup munging to merge this content together. I think I got it right, but since you're both more familiar with the content, would you mind taking a look at the diff?

designmatty:
Looks good to me. @dboyer may still want to look this over too.

dboyer:
I'm late to the party but these changes look great.

PR #1370, opened 2015-04-22
Add JMeter test plan for All Blockedge Progress with 3x zoom

hectcastro:
This changeset adds a JMeter test plan for the All Blockedge Progress map with steps for zooming in 3x. So, first the initial map loads, then zoom in 3x total, waiting each time for all the blocks to load.

---

On a development virtual machine:

3x


jwalgran:
+1 This has proved useful in profiling

PR #1376, opened 2015-04-22
Add CloudWatch alarms to CloudFront distribution

hectcastro:
This changeset adds CloudWatch alarms to CloudFront tile distribution. The two metrics right now are backend 4XX and 5XX failures.


jwalgran:
+1 I saw the alarm email when this was applied to staging

PR #1381, opened 2015-04-22
Hide private events on the map page

### LINKED ISSUE #1379: Event map view shows private events
screen shot 2015-04-22 at 1 03 52 pm

screen shot 2015-04-22 at 1 01 51 pm

jwalgran:
They are hidden on the list view and should also be hidden on the map.

Fixes #1379


RickMohr:
+1 works.

PR #1386, opened 2015-04-23
Sort species modal by tree count for on public home page

### LINKED ISSUE #1102: Species list not sorted in order of prevalence and needs scroll bar
When clicking the species progress circle, the list should be sorted from species with most trees to fewest trees.

maurizi:
This had already been done for the dashboard, but the public homepage was missed.

Not sure if this should be against develop or the release branch.

Fixes #1102


jwalgran:
There will not be any species in the list until mapping begins, so merging into develop is good.

PR #1406, opened 2015-04-23
Don't escape punctuation in email subject lines

### LINKED ISSUE #1401: Odd characters in email subject line
You are registered for NYC Parks&#39; 1st Training Event on Tuesday, May 19th

Received an email with that as the subject line when I registered for /group/nyc-parks/event/trees-count-2015-mapping-event/

jwalgran:
By default Django's template engine escapes template content to protects against injection. When we are using a template to render a plain text email subject, we do not want this escaping.

Fixes #1401


hectcastro:
:+1:

hectcastro:
Merging to start the AMI building process.

PR #1408, opened 2015-04-23
Show group border even if territory is not defined

### LINKED ISSUE #1407: Group borders are reported as uploaded, but do not appear on the map
Parks did an geojson upload to staging and got statuses like this

screen shot 2015-04-22 at 8 27 43 pm

Indicating that some boundaries were uploaded, but when they browsed the group progress page the polygons were not there (only 2 manually created test ones were)

screen shot 2015-04-22 at 8 29 33 pm

jwalgran:
If a group has a border polygon defined, we can show it on the group map even if the territory for the group has not been defined.

Fixes #1407

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -228,10 +228,11 @@ def group_borders_geojson(request):
                 'bounds': GeometryCollection(
                     [territory.blockface.geom
                      for territory in group.turf]).extent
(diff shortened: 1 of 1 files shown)


dboyer:
If this is showing the group polygon on the group's detail page, @jfrankl had actually decided to go with showing the block edges as it is more specific than the general polygon. Polygons are limited to the groups view of the progress map.

RickMohr:
+1

PR #1410, opened 2015-04-23
Show group polygon on progress page even if group has no territory

### LINKED ISSUE #1407: Group borders are reported as uploaded, but do not appear on the map

RickMohr:
Fixes #1407


RickMohr:
Closing in favor of #1408, essentially identical.

PR #1428, opened 2015-04-23
Fix long group names in actionbar

### LINKED ISSUE #1387: Long group names run into chevron in action bar on progress map
This is only on desktop. The arrangement of the bar on on mobile doesn't cause the same issue.

group-name

jfrankl:
Fixes #1387.

image


maurizi:
+1

PR #1437, opened 2015-04-24
Update achievement detail text

### LINKED ISSUE #1426: Update achievement detail text

jfrankl:
Fixes #1426.

### DIFF
--- a/src/nyc_trees/apps/users/models.py
+++ b/src/nyc_trees/apps/users/models.py
@@ -66,13 +66,13 @@ achievements = OrderedDict([
         achieved=lambda user: user.online_training_complete
     )),
     (AchievementDefinition.TRAINING_EVENT, AchievementDefinition(
-        name='Treerifically Trained',
+        name='<b>Treer</b>ifically Trained',
         description='Attend a Training Event',
         description_achieved='Attended a Training Event',
         badge='img/badges/ic_badge_training_event.png',
         achieved=lambda user: user.field_training_complete,
         reward='Receive a Census Gear Pack with a measuring tape, '
-        'Tree Species ID Guide, voluntreer lanyard and badge, and a '
+        'Tree Species ID Guide, volun<b>treer</b> lanyard and badge, and a '
         'TreesCount safety vest!'
     )),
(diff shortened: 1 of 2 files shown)


src/nyc_trees/apps/users/models.py
```diff
@@ -89,13 +89,10 @@ def __init__(self, name, description, description_achieved, badge,
...
         sponsor='Various sponsors',
-        reward='The first 250 volun<b>treers</b> to reach '
-                'this level will receive tickets to a NY '
-                'Liberty Game! All volun<b>treers</b> will '
-                'be entered into lotteries to attend a '
-                'Picnic in the Park, win a personal '
-                'Tree Care Workshop, a Knicks gift '
-                'bag and other great prizes while supplies last.'
+        reward='The first 250 voluntreers to reach this level will '
+        'receive tickets to a NY Liberty Game! All voluntreers will '
```
maurizi: I assume we're purposefully removing the `<b>` around "treer"?

If so, can you remove the `|safe` template filter here: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/users/templates/users/partials/achievement_modal.html#L6?

src/nyc_trees/apps/users/models.py
```diff
@@ -114,10 +111,9 @@ def __init__(self, name, description, description_achieved, badge,
         badge='img/badges/ic_badge_map_200.png',
         achieved=lambda user: user.blocks_mapped_count >= 200,
         sponsor='Sponsored by Whole Foods Market',
-        reward='TreesCount! Tote &amp; Tree Care Kit '
-               'that includes gloves, weeder, '
-               'cultivator, trowel, a collapsible bucket '
-               'and other helpful tools'
+        reward='TreesCount! Tote & Tree Care Kit that includes a limited '
```
maurizi: If we keep the `|safe` template filter in place, this `&` will need to stay as `&amp;`.  If we do remove the `|safe`, I _think_ it will be ok to leave it as a literal `&`, but we should double check.

jfrankl:
We're going to leave the bold treers in place. Please advise about ```&```.

maurizi:
+1

PR #1439, opened 2015-04-24
Set an appropriately sized iOS shortcut title

jwalgran:
The ! in the page title results in "TreesCoun..." as the default iOS home screen short cut label. Adding this meta tag provides a default that fits in the allowable home screen icon label space.

Before

image1

After

image1-1


kdeloach:
+1 Tested and working

PR #1444, opened 2015-04-24
Remove unecessary join to Territory table from progress layer queries

maurizi:
This was part of #1313, I am pulling it out because it is a smaller simpler change with less risk involved

### DIFF
--- a/src/tiler/sql/user_progress.sql
+++ b/src/tiler/sql/user_progress.sql
@@ -2,14 +2,12 @@
 -- the latest survey for a blockface
 (SELECT
   DISTINCT ON (block.id)
-  block.geom, block.id, turf.group_id,
+  block.geom, block.id,
   CASE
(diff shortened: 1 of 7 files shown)


src/tiler/sql/user_progress.sql
```diff
@@ -2,14 +2,12 @@
 -- the latest survey for a blockface
 (SELECT
   DISTINCT ON (block.id)
-  block.geom, block.id, turf.group_id,
+  block.geom, block.id
```
jwalgran: This crashes the tiler because a comma is missing before the ``CASE``

jwalgran:
The progress pages function normally. The more complex popup request is still only 10s of milliseconds. Running JMeter against my dev VM showed a couple percent decrease in average request time.

src/tiler/sql/user_progress.sql
maurizi: Fixed

jwalgran:
+1

PR #1449, opened 2015-04-27
Add support for C3 and C4 class instance types

hectcastro:
This changeset allows application and tiler instances to be launched as C3 and C4 class instance types.


jwalgran:
This has already proven useful for profiling the effects of machine size on tiler performance.

:+1:

Issue #1238, opened 2015-04-17
map controls show on tree logging page


lliss:
Firefox on Android KitKat

![map_controls](https://cloud.githubusercontent.com/assets/903219/7209068/10523264-e513-11e4-91c2-508c979f501f.png)

maurizi:
This is also an issue on the progress map with the side navigation (on Firefox for android)

maurizi:
![tmp_22643-screenshot_2015-04-26-19-55-03971671261](https://cloud.githubusercontent.com/assets/4432106/7339994/5b0a3b5a-ec50-11e4-8027-1f39dd55560a.png)

PR #1450, opened 2015-04-27
fixed map control z-indexes

designmatty:
fixes - #1238


maurizi:
+1, no more controls where they shouldn't be

PR #1452, opened 2015-04-27
Change language for training/mapping events

### LINKED ISSUE #1442: Change reservations instructions
The third step should be "attend a mapping event" rather than "attend a second training event."

If this series of steps is listed anywhere else, we should make the same change.

RickMohr:
* "Mapping event" instead of "second training event"
* "Training event" instead of "first training event"

Fixes #1442


RickMohr:
waffle

RickMohr:
waffle 2

RickMohr:
waffle three

maurizi:
+1

PR #1455, opened 2015-04-27
Set maximum zoom on all maps

### LINKED ISSUE #1120: Map pages zoom in too far and don't work right
1) In a full-screen desktop browser, reserve a single block
2) Return to the Reservations page -- you see no base map

In zooming to the extent of the user's reservations we have zoomed in too far.

If you click "-" to zoom out the map appears, but then you can't select a block.

RickMohr:
Fixes #1120

### DIFF
--- a/src/nyc_trees/js/src/map.js
+++ b/src/nyc_trees/js/src/map.js
@@ -23,6 +23,7 @@ function create(options) {
     }, options);
 
     var mapOptions = {
+        maxZoom: zoom.MAX,
         attributionControl: false,
         zoomControl: false,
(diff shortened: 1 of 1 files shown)


src/nyc_trees/js/src/map.js
```diff
@@ -23,6 +23,7 @@ function create(options) {
     }, options);
 
     var mapOptions = {
+        maxZoom: zoom.MAX,
```
kdeloach: Consider adding `minZoom` too.

kdeloach:
+1

src/nyc_trees/js/src/map.js
RickMohr: Good idea, done.

PR #1459, opened 2015-04-27
Cleanup HTML

### LINKED ISSUE #1415: Validate markup
There's some invalid markup on the homepage and probably other pages. I happened to run the homepage through the w3c validator. This might account for some of the issues we were seeing with older browsers during the testing session. One obvious thing is that we have script tags which occur outside of the closing body tag. I'm pretty sure that shouldn't be done.

kdeloach:
Various changes to produce W3C valid markup.

This only fixes errors, not warnings, and does not alter any flatpages
markup.

Summary of changes:

* Add alt attribute to images
* Add form actions
* Fix invalid element nesting (block level elems inside inline elems)
* Fix stray script tags (declare scripts inside body)
* Fix stray closing tags

Fixes #1415

### DIFF
--- a/src/nyc_trees/apps/core/templates/base.html
+++ b/src/nyc_trees/apps/core/templates/base.html
@@ -17,7 +17,7 @@
                     <span class="icon-bar"></span>
                 </button>
                 <div class="brand nav-left">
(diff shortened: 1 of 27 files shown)


jfrankl:
+1

PR #1460, opened 2015-04-27
Update gulp-ruby-sass and add autoprefixer

### LINKED ISSUE #142: Sourcemaps not loading for CSS files
It looks like the URLs have `/opt/app` in front of them.

### LINKED ISSUE #1233: Implement css Autoprefixer

maurizi:
- Update gulp-ruby-sass to 1.0.5 to fix sourcemap and minification support.
- Use autoprefixer to add vendor prefixes to CSS selectors and properties.

Sourcemaps were broken in the version of gulp-ruby-sass we were using
because it wanted us to serve the original files and would not allow us
to choose the path to serve them from.

Minification was not implemented in favor of using a "compressed"
outputStyle because we did not have sourcemap support and I wanted to
enable debugging production issues easily.

Minification doesn't get us much (33Kb uncompressed vs. 30Kb compressed),
but it was trivial to add.

I chose to use '>1% in US' as the browsers to support with autoprefixer, though it is trivial
to tweak this if we need to.

Bringing in the latest autoprefixer also required updating gulp-postcss
and csswring, since they all need to be using the same version of postcss,
and the latest version of autoprefixer requires postcss of 4.0+

Fixes #142
Fixes #1233

### DIFF
--- a/src/nyc_trees/gulpfile.js
+++ b/src/nyc_trees/gulpfile.js
@@ -199,10 +199,14 @@ function browserifyTask(bundler) {
 }
 
 gulp.task('sass', ['clean'], function() {
-    return gulp.src('sass/main.scss')
-        .pipe(sourcemaps.init())
-        .pipe(sass({style: 'compressed'}))
-        .pipe(sourcemaps.write('./'))
+    return sass('sass/main.scss', { sourcemap: true })
+        .on('error', function (err) {
+            console.error('Sass error', err.message);
+        })
+        .pipe(gulpif(! args.debug, postcss([csswring])))
+        .pipe(sourcemaps.write('./', {
+            includeContent: true,
+        }))
         .pipe(gulp.dest(cssDir));
 });
 
--- a/src/nyc_trees/gulpfile.js
+++ b/src/nyc_trees/gulpfile.js
@@ -24,7 +24,8 @@ var gulp = require('gulp'),
     del = require('del'),
     shell = require('gulp-shell'),
     runSequence = require('run-sequence'),
-    jshint = require('gulp-jshint');
+    jshint = require('gulp-jshint'),
+    autoprefixer = require('autoprefixer-core');
 
 var args = minimist(process.argv.slice(2),
                     {default: {debug: false}}),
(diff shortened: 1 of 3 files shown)


maurizi:
@jwalgran Can I get your eyes on this too?

src/nyc_trees/gulpfile.js
```diff
@@ -199,10 +200,19 @@ function browserifyTask(bundler) {
...
 gulp.task('sass', ['clean'], function() {
-    return gulp.src('sass/main.scss')
-        .pipe(sourcemaps.init())
-        .pipe(sass({style: 'compressed'}))
-        .pipe(sourcemaps.write('./'))
+    var autoprefixProcessor = autoprefixer({ browsers: ['> 1% in US'] });
+
+    return sass('sass/main.scss', { sourcemap: true })
+        .on('error', function (err) {
+            console.error('Sass error', err.message);
```
jwalgran: Do you know if we exit non-zero if there is a sass error? A failure in sass compilation should fail the build.

jwalgran:
How do I correctly test that ``autoprefixer`` is working correctly. I tried removing some prefixes from a sass partial (https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/sass/partials/_map.scss#L230) and they were not added back in by ``autoprefixer``

screen shot 2015-04-27 at 4 23 56 pm

src/nyc_trees/gulpfile.js
maurizi: I don't think this PR will have changed anything in regards to how we handle sass errors.

That said, I think it is definitely a good idea to exit with a non-zero error code if the build fails (for production builds, for debug builds I think we should just warn).  I can see if it is easy to do, if not I'll open a separate issue.

maurizi:
@jwalgran Funny, that is exactly what I tried to do to test this out.  Since autoprefixer only adds prefixes for things that need them, and `transition`  is [actually really well supported](http://caniuse.com/#search=css%20transition) it won't get prefixed.

I verified this by removing some prefixes from `animation`, which gets [prefixed on WebKit browsers](http://caniuse.com/#search=css3%20animation)

designmatty:
@jwalgran pretty often we will prefix things just because we're not used to browsers actually working. So if you remove a prefix and it doesn't become autoprefixed then that's the reasoning. @maurizi I'm still looking over this but so far it looks good to me.

src/nyc_trees/gulpfile.js
maurizi: I tried a few things but I can't seem to get `gulp-ruby-sass` to actually pass errors through to us.  I'm going to open an issue for this for now.

jwalgran:
I removed the vendor prefixes from ``.animate-spin`` (https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/sass/partials/_icon_animation.scss#L4) and gulp put ``-webkit-animation`` back into ``main.css``. Looks good.

screen shot 2015-04-28 at 12 43 50 pm

designmatty:
Everything seems to be working on my end as well. +1 from me

PR #1464, opened 2015-04-27
Add a class to the body when content is added to the actionbar.

### LINKED ISSUE #721: Add class on body when actionbar is open

maurizi:
The survey, event list, and reservations maps all show content in the
action bar on page load, so I did not do this on those pages.

Fixes #721

### DIFF
--- a/src/nyc_trees/js/src/progressPage.js
+++ b/src/nyc_trees/js/src/progressPage.js
@@ -61,6 +61,7 @@ function loadLayers($mode) {
 
     // Clear action bar
     $actionBar.empty();
+    $('body').removeClass('actionbar-triggered');
 
     // Replace layers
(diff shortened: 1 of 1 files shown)


jfrankl:
+1 after the switch from removeClass to addClass we discussed

PR #1465, opened 2015-04-27
Topic/1454/quiz edits

steventlamb:


steventlamb:
does this look ok?
screenshot from 2015-04-27 17 51 50


screenshot from 2015-04-27 17 52 15

steventlamb:
@dboyer @jwalgran I think these are all the changes mentioned in the comments as well. Please let me know if we need to cut any of these commits out.

jwalgran:
+1

I think the help text over Question 1 looks good.

PR #1467, opened 2015-04-27
Allow any active user to be selected as a teammate

### LINKED ISSUE #335: Mappers may select a teammate who assisted them with mapping
Refers to Task 28 in the scope - Support Team Mapping. A team can be no more than two people so the user entering the data can only note that one other person helped them with mapping.

kdeloach:
Fixes #335

### DIFF
--- a/src/nyc_trees/apps/survey/helpers.py
+++ b/src/nyc_trees/apps/survey/helpers.py
@@ -9,18 +9,11 @@ from apps.users import can_show_full_name
 from apps.survey.models import Survey, Territory
 
 
-def teammates_for_event(group, event, current_user):
-    attendees = User.objects.exclude(id=current_user.id)\
-                            .filter(eventregistration__event=event,
-                                    eventregistration__did_attend=True)\
+def teammates_for_mapping(current_user):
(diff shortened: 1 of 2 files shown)


src/nyc_trees/apps/survey/helpers.py
```diff
@@ -10,16 +10,16 @@
...
-    attendees = User.objects.exclude(id=current_user.id)\
-                            .filter(eventregistration__event=event,
-                                    eventregistration__did_attend=True)\
+    attendees = User.objects.exclude(id=current_user.id) \
+                            .filter(is_active=True) \
                             .order_by('username')
     return _teammates_context(attendees)
 
 
 def teammates_for_individual_mapping(current_user):
```
kdeloach: Could probably combine these functions...

maurizi: Yeah, I think that would be a good idea.

kdeloach: Updated.

maurizi:
+1, works great!

PR #1478, opened 2015-04-28
Topic/499/add empty text to chrono filters

steventlamb:


steventlamb:
@dboyer the second commit could use some content review. What do we want to say here?

RickMohr:
+1, tested. Text is great.

dboyer:
Text looks good to me too.

PR #1479, opened 2015-04-28
Only allow user to select reserved blocks on survey page

### LINKED ISSUE #834: Treecorder - You can survey any block by putting it's ID in the hash
Steps to reproduce:

 1) Find a blockface that is not reserved by the current user, and note it's ID (say 16489)
 2) Go to the survey URL with that ID in the hash: `http://localhost:8000/survey/#16489`

Expected results: The blockface is not selected, and you must pick one of your reserved blockfaces.
Actual results: The blockface is selected.

kdeloach:
This removes the ability to select any blockface on the survey page by
changing the blockface ID in the URL hash. The blockface present in the
URL hash will only be selected if the user has reserved that block.

Fixes #834

### DIFF
--- a/src/nyc_trees/apps/survey/urls/blockface.py
+++ b/src/nyc_trees/apps/survey/urls/blockface.py
@@ -7,7 +7,7 @@ from django.conf.urls import patterns, url
 
 from apps.survey.routes import (
     reserve_blockface_page, reserve_blockfaces, reservations_instructions,
-    blockface, progress_page, blockface_reservations_confirmation_page,
+    progress_page, blockface_reservations_confirmation_page,
     progress_page_blockface_popup, printable_reservations_map,
     reservations_map_pdf_poll, user_reserved_blockfaces_geojson,
     group_borders_geojson, group_popup
@@ -40,11 +40,6 @@ urlpatterns = patterns(
 
     url(r'^checkout-confirmation/$', blockface_reservations_confirmation_page,
         name='blockface_reservations_confirmation_page'),
-
-    # Note: this must be kept in sync with the hardcoded url in
-    # js/src/mapUtil.js
-    url(r'^(?P<blockface_id>\d+)/$', blockface,
-        name='blockface'),
 
     url(r'^map-poll/$', reservations_map_pdf_poll,
         name='reservations_map_pdf_poll'),
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -422,12 +422,49 @@ def _get_reservations_to_cancel(ids, user):
         .current()
 
 
-def blockface(request, blockface_id):
+def fetch_user_blockface(request, blockface_id):
+    """Return blockface if user has reserved it"""
     blockface = get_object_or_404(Blockface, id=blockface_id)
+    reservation = BlockfaceReservation.objects.filter(
+                      blockface=blockface,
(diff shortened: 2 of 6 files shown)


maurizi:
-1.  What about surveying at events?

maurizi:
The code for the reservations workflow looks good.
Please assign this back to me when surveying from events is also handled

src/nyc_trees/js/src/surveyPage.js
```diff
@@ -110,7 +110,9 @@ var dom = {
 
     formTemplate = Handlebars.compile($(dom.treeFormTemplate).html()),
 
-    blockfaceId = mapUtil.getBlockfaceIdFromUrl(),
+    // jshint complains when initializing this to undefined
+    blockfaceId = -1,
```
maurizi: Consider just declaring this but leaving it undefined, to prevent the lint, instead of assigning `-1`:
```javascript
var ... = ...,
     blockface,
     ... = ...,
```

kdeloach:
This is basically a noop when surveying at an event. I guess you could abuse this loophole to survey a block *that you have reserved*, even if it's not part of the current event? Not sure if that's a real problem though. 

Since we don't (as far as I know) deeplink to blocks for event surveys, I'm inclined to leave as-is. Open to suggestions though.

src/nyc_trees/js/src/surveyPage.js
kdeloach: Updated.

kdeloach:
It was pointed out to me that we actually *do* deeplink to blocks on the events survey page. When you click "No, Restart" after submitting a survey, it should select the block you just tried to map even if you don't have it reserved.

kdeloach:
OK, take a look at the latest commit. It's 5x more code than the original commit and does essentially the same thing.

src/nyc_trees/apps/survey/views.py
```diff
@@ -422,12 +422,49 @@ def _get_reservations_to_cancel(ids, user):
...
+        'success': False,
+        'message': 'User has not reserved this blockface'
+    }
+
+
+def fetch_event_blockface(request, event_slug, blockface_id):
+    """Return blockface if it is part of this event"""
+    blockface = get_object_or_404(Blockface, id=blockface_id)
+    event = get_object_or_404(Event, slug=event_slug)
+
```
maurizi: I think we need a third check here, to see if `blockface.is_available` is `True`

src/nyc_trees/apps/survey/urls/blockface.py
```diff
@@ -41,11 +41,6 @@
     url(r'^checkout-confirmation/$', blockface_reservations_confirmation_page,
         name='blockface_reservations_confirmation_page'),
 
-    # Note: this must be kept in sync with the hardcoded url in
-    # js/src/mapUtil.js
-    url(r'^(?P<blockface_id>\d+)/$', blockface,
-        name='blockface'),
```
maurizi: Removing this view has broken the reservations page.  I think we need to add it back in.

maurizi:
I tested this and the behaviour seems correct, but I think the 2 comments I raised need to be fixed before this can be merged.

kdeloach:
I don't think it would be wise to spend any more time on this bug considering the outstanding list of issues remaining. I'll revisit this issue if I discover a simpler fix, but until then I think it's best to close this.

PR #1485, opened 2015-04-28
Don't load all blockfaces to an admin site dropdown

### LINKED ISSUE #1141: Prevent loading all the blockfaces into a dropdown on the admin site
Quick and dirty, raw_id_fields

https://docs.djangoproject.com/en/1.7/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields

http://stackoverflow.com/questions/5108080/improving-performance-of-django-foreignkey-fields-in-admin

RickMohr:
Fixes #1141


maurizi:
Probably a good idea to do this for the `id` of `Blockface` itself as well.

steventlamb:
+1, tested

RickMohr:
@maurizi, it only works with foreign keys.

PR #1486, opened 2015-04-28
Prevent editing index fields via the Django admin site.

### LINKED ISSUE #1413: Don't show "Territory updated at" on Django admin for groups
We don't want anybody inadvertently changing this.

image

maurizi:
It's unclear to me why Django won't pick up the changes to fields other than `Group.territory_updated_at` when I run `makemigrations`.  Since they do not reflect DB changes,
I am not overly worried about it.

Fixes #1413

### DIFF
--- a/src/nyc_trees/apps/core/migrations/0027_auto_20150428_1529.py
+++ b/src/nyc_trees/apps/core/migrations/0027_auto_20150428_1529.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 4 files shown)


steventlamb:
+1, tested.

For completeness, I tried making the migrations on my machine and had the same results you did. I ran the migrations, backed them out, ran them again, checked the admin interface for the changed models and a few of the `NycModel`s, everything looks good.

We won't know if this migration weirdness is a problem until we merge it and see if a problem turns up. probably it won't be a problem.

PR #1487, opened 2015-04-28
PREVIEW: send paper worksheet

steventlamb:
@jwalgran this looks like it'll work, how do you want to handle storing and retrieving the PDF?


jwalgran:
We have the training material PDF in our static content folder, so that is where I would expect this other, static PDF to be saved. We could pull it from the filesystem using ``settings.STATIC_ROOT + 'what/ever/subpath/TreesCount2015_worksheet.pdf``

jwalgran:
I like where this is headed.

steventlamb:
@jwalgran soft-coded the path, ran tests, poked the modified the other changed email tasks. Ready for official review.

jwalgran:
Testing now.

jwalgran:
+1 confirmed that 2 PDFs were attached when I selected paper mapping. Existing RSVP PDF emailing still works as before.

PR #1489, opened 2015-04-29
expand select2 dropdown full height for tablet size and down

designmatty:
This will make select2 a bit easier to use when on tablets


maurizi:
+1

PR #1495, opened 2015-04-29
Focus on first input field when adding a new tree in treecorder.

### LINKED ISSUE #878: Scroll page to first data field after user clicks + Another Tree
In the current version of the Treecorder, clicking + Another Tree adds the fields for another tree but the page stays scrolled to the bottom. The page should scroll up to the first data field so users understand where they are supposed to add data.

maurizi:
Fixes #878

### DIFF
--- a/src/nyc_trees/js/src/surveyPage.js
+++ b/src/nyc_trees/js/src/surveyPage.js
@@ -359,6 +359,12 @@ $(dom.addTree).click(function (){
         $treeForms.collapse('hide');
         $newForm.collapse('show');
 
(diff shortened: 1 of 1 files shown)


maurizi:
I briefly experimented with doing an animated scroll to the first field, but it is really tricky to do so in combination with the animation Bootstrap does when collapsing sections.  If that is something that is desirable, it will probably need some design/engineering collaboration.

RickMohr:
Can we also focus on the first input field for the very first tree?

RickMohr:
Otherwise +1, tested.

PR #1496, opened 2015-04-29
Hide other problems sections when "No Problems" is selected.

### LINKED ISSUE #878: Scroll page to first data field after user clicks + Another Tree

### LINKED ISSUE #879: Clicking "no problems" should hide sub-sections with root, trunk, and branch problems
This is meant to prevent users from accidentally changing their answer.

maurizi:
Fixes #879


steventlamb:
+1, tested

PR #1497, opened 2015-04-29
Avoid unnecessary tiler requests

### LINKED ISSUE #1491: User progress link should not show full map if the user has no mapped blocks
The layer is personalized, and expensive to draw. Don't show it if it is no unique.

### LINKED ISSUE #1494: Zoom before issuing tile requests
On the many pages where we zoom to show particular blockfaces, make sure we are zooming before issuing tile requests. Rendering a small number of blockfaces should be quick for the tiler, but not if we mistakenly render the entire city first.

RickMohr:
Fixes #1491 
Fixes #1494

These map pages were generating spurious tiler requests (fixed by this PR):
* Progress - My Progress
* Progress - Groups

These were fine:
* Group Detail
* Event PDF
* Reservations
* Reservations PDF
* Treecorder
* Admin Territory

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/progress.html
+++ b/src/nyc_trees/apps/survey/templates/survey/progress.html
@@ -21,18 +21,21 @@
                            data-grid-url="{{ layer_all.grid_url }}"
                         >All Users</a>
                     </li>
-{% comment %}
-TODO: Uncomment before May 19
                     {% if request.user.is_authenticated %}
                     <li class="text-right" role="presentation">
-                        <a role="menuitem" href="#my"
-                           data-tile-url="{{ layer_my.tile_url }}"
-                           data-grid-url="{{ layer_my.grid_url }}"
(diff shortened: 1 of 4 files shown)


jwalgran:
The map still works, but I am having a hard time finding a way to test this. I put a breakpoint here https://github.com/azavea/nyc-trees/pull/1497/files#diff-3178765eaf2e28def49d127a49277a1bR182  but it was never hit.

RickMohr:
I tested it by looking at the network tab in chrome developer tools, filtering "png". Before this change, choosing "My Progress" or "Group" I would see cancelled tile requests at the current zoom level (11 for me) before the zoomed-in requests. With this change I saw only the zoomed-in requests.

If I set that breakpoint and click "My Progress" or "Group" the debugger stops there.

jwalgran:
My mistake. I do now see it working as designed. 

+1

Sorry that this will be a bit of a messy merge due to 0db5901a

PR #1501, opened 2015-04-29
Disable map during survey

### LINKED ISSUE #766: Disable survey map interactions while filling out tree details

steventlamb:
fixes #766 on github


src/nyc_trees/js/src/map.js
```diff
@@ -75,7 +80,7 @@ function create(options) {
         zoomControl.removeFrom(map);
     }
 
-    return map;
+    return {map: map, zoomControl: zoomControl};
```
RickMohr: Rather than returning and removing `zoomControl` I suggest you return and remove `$controlsContainer`, so we remove whatever map controls are there. There's an issue open to add search to this page, and the zoom control might be removed for mobile.

RickMohr:
Otherwise +1, works. Why does removing `zoomControl` also remove the geolocation control?

steventlamb:
closed in favor of 1517

Issue #1492, opened 2015-04-29
Remove grid requests from progress page


jwalgran:
~~Measure first - remove grid requests from JMeter script and run on existing code~~

Trade clickability for cutting tile requests in half. 

Measurements to compare before/after change
  - Data size passed between VMs
  - Slow query log output

hectcastro:
Graphed output that compares the `all-users-3x-zoom-blockedge-progress.jmx` JMeter test plan with and without the UTF grid requests:

![image](https://cloud.githubusercontent.com/assets/43639/7397915/79c93b38-ee79-11e4-9828-2757ac9d0b32.png)

![image](https://cloud.githubusercontent.com/assets/43639/7397925/82054b84-ee79-11e4-958a-e79d28598325.png)

![image](https://cloud.githubusercontent.com/assets/43639/7397929/884ab9fc-ee79-11e4-83c8-c5e7c6af235d.png)

![image](https://cloud.githubusercontent.com/assets/43639/7397941/8d1aa992-ee79-11e4-9b6b-576421d60ece.png)

![image](https://cloud.githubusercontent.com/assets/43639/7397942/912b94ce-ee79-11e4-811e-6021a5dc6c83.png)

![image](https://cloud.githubusercontent.com/assets/43639/7397944/9527e8fc-ee79-11e4-9fcb-fa8d3dfd237b.png)

hectcastro:
How I'm measuring traffic from PostgreSQL to the tiler virtual machine:

```bash
vagrant@tiler:~$ sudo iftop -i eth1 -B -f "src port 5432 and host 33.33.33.30”
```

jwalgran:
All graphs show a decrease in resource usage. Since we round trip to the server to get block details, it is pretty clear that we should follow through with removing the grid.

jwalgran:
Fetching a geom by ID is  0.035 ms. Fetching via a spherical distance query is 0.247 ms.

```
SELECT  "survey_blockface"."id", "survey_blockface"."geom", "survey_blockface"."is_available", "survey_blockface"."expert_required", "survey_blockface"."source", "survey_blockface"."created_at",
"survey_blockface"."updated_at" FROM "survey_blockface" WHERE "survey_blockface"."id" = 220441; 

 Index Scan using survey_blockface_pkey on survey_blockface  (cost=0.42..8.44 rows=1 width=119) (actual time=0.015..0.016 rows=1 loops=1)
   Index Cond: (id = 220441)
[...]

PR #1502, opened 2015-04-30
Remove progress UTF grid, add fetch by lat/lng

jwalgran:
It is expensive to render UTF grids for the entire city. This commit removes the grid layer from the progress page, and keeps the click-to-select functionality by adding an endpoint that fetches the feature closest to a point.

Connects to #1492

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -11,13 +11,13 @@ from pytz import timezone
 from celery import chain
 
 from django.conf import settings
-from django.contrib.gis.geos import GeometryCollection
+from django.contrib.gis.geos import Point, GeometryCollection
 from django.core.exceptions import ValidationError, PermissionDenied
 from django.core.urlresolvers import reverse
 from django.db import transaction, connection
 from django.db.models import Prefetch
 from django.http import (HttpResponse, HttpResponseForbidden,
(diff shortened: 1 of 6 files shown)


RickMohr:
Note potential optimization described in http://postgis.refractions.net/documentation/manual-2.0/geometry_distance_centroid.html -- get nearest neighbor candidates using fast imprecise <-> operator (which uses spatial index) and refine using slow precise ST_distance_sphere.

src/nyc_trees/apps/survey/views.py
```diff
@@ -429,6 +428,28 @@ def blockface(request, blockface_id):
...
+    # The size of the distance filter was chosen through trial and
+    # error by testing tap precision on a mobile device
+    qs = Blockface.objects.filter(geom__dwithin=(p, 0.0002))\
+                          .distance(p)\
+                          .order_by('distance')
+    blockfaces = qs[:1]  # We only want the closest blockface
+    if blockfaces:
+        return _blockface_context(blockfaces[0])
+    else:
+        raise Http404
```
maurizi: Raising `Http404` leads to a lot of spurious "errors" in the browser console log.  Consider returning `{}` with a status code of 200 instead.

maurizi:
+1.  This works great!  I played with the network throttling and even on "Regular 2G" this seemed fast enough to me.

PR #1504, opened 2015-04-30
Add ability to delete the last tree in the treecorder.

### LINKED ISSUE #954: Allow users to delete the last tree in the Treecorder survey form
Currently, users may click the add another tree button and then realize there are no more trees on the block. There isn't anyway to get rid of that last set of tree fields.

Users should only be able to delete the last tree on a survey. They cannot delete other trees since that will impact the distance measurements.

maurizi:
The button will only be present on the last tree, and only if there is
more than one tree form.

If any data has been entered for the tree, a modal  window will be
presented asking for confirmation.

Fixes #954


RickMohr:
+1, tested.

PR #1509, opened 2015-04-30
Disallow actions on past events

### LINKED ISSUE #939: We don't prevent "Edit" and "RSVP" for events in the past
1) On a Group page, click Events > PAST
2) Click on an event
3) Click Edit or RSVP

Expected: Edit and RSVP buttons are disabled or absent
Actual: You can edit or RSVP even though the event is history.

Allowing editing seems bad, tempting admins to re-use an old event, obscuring history. And what if the event has RSVP's?

steventlamb:
fixes #939 on github

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/partials/rsvp.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/rsvp.html
@@ -16,37 +16,41 @@ The following variables must be in scope:
     </div>
 </div>
 <div class="col-xs-8 text-right">
-    {% if is_admin %}
-        <a href="{{ event_edit_url }}"
-           class="btn btn-secondary"><i class="icon-cog"></i>Edit</a>
-    {% endif %}
-    {% if is_rsvped %}
-        <a id="rsvp"
-           href="#"
-           data-verb="DELETE"
-           data-url="{{ rsvp_url }}"
-           class="btn btn-switch btn-rsvp active">Attending</a>
-    {% elif can_rsvp %}
-        {% if not user.is_authenticated %}
-            <a id="rsvp"
-               href="{% url "login" %}?next={{ event.get_absolute_url }}"
-               class="btn btn-switch">RSVP</a>
-        {% elif not user.online_training_complete %}
-            <a id="rsvp"
-               href="{% url "training_instructions" %}"
-               class="btn btn-switch btn-rsvp">RSVP</a>
-        {% elif not event.includes_training and not user.training_complete %}
-            <a id="rsvp"
-               href="{% url "training_instructions" %}"
-               class="btn btn-switch btn-rsvp">RSVP</a>
-        {% else %}
+    {% if event.is_past %}
+        <a id="rsvp" href="#" class="btn btn-switch disabled">This event has ended.</a>
+    {% else %}
+        {% if is_admin %}
+            <a href="{{ event_edit_url }}"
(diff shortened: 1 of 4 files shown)


src/nyc_trees/apps/event/templates/event/partials/rsvp.html
```diff
@@ -16,37 +16,41 @@ <h1 class="rsvp-count-total pull-left color--primary nomargin">{{ rsvp_count }}<
...
-            <a id="rsvp"
-               href="{% url "training_instructions" %}"
-               class="btn btn-switch btn-rsvp">RSVP</a>
-        {% elif not event.includes_training and not user.training_complete %}
-            <a id="rsvp"
-               href="{% url "training_instructions" %}"
-               class="btn btn-switch btn-rsvp">RSVP</a>
-        {% else %}
+    {% if event.is_past %}
+        <a id="rsvp" href="#" class="btn btn-switch disabled">This event has Ended</a>
```
maurizi: Consider using ` href="javscript:" ` instead to avoid messing up the browser history.

steventlamb: Good idea, but none of the other disabled buttons in this section are using that. Does it really matter for a disabled button?

maurizi: If it was a button with a disabled attribute (`<button disabled="disabled">`) then it would not matter at all, but adding a `disabled` class to an anchor tag doesn't actually prevent clicks or changes to the URL.

Not a huge deal though, feel free to ignore the suggestion.

steventlamb: Based on testing and our discussion, this actually works fine, so I will not change it.

kdeloach:
On the group detail page, if you go to the "Past" events tab, and click "Edit Details", it results in an empty page:
image

src/nyc_trees/apps/event/templates/event/partials/rsvp.html
kdeloach: Consider not showing any buttons if the event has ended.

kdeloach:
+1

steventlamb:
Ah, nice catch. That empty page is a 403, but I need to change that on the client.

RickMohr:
You need to be able to email attendees after the event is ended, but you shouldn't be able to edit details or check people in.

src/nyc_trees/apps/event/templates/event/partials/rsvp.html
steventlamb: I'm falling in line with what already exists. We already show a stand-in disabled button for the other disable case (at capacity).

Granted, in that case, it's just toggle RSVP with a disabled at capacity whereas now I'm replacing two buttons with one disabled button, but it seems consistent to me.

Happy to defer to @dboyer @jfrankl .

steventlamb:
@RickMohr @kdeloach fixed.

src/nyc_trees/apps/event/templates/event/partials/rsvp.html
kdeloach: That makes sense.

RickMohr: I would say "This event has ended." (Don't capitalize "Ended", and add a period.)

RickMohr:
Otherwise +1, tested.

PR #1510, opened 2015-04-30
Stop modifying training pages during setup

steventlamb:
Now that this app is live, there's not ever going to be a case where we
want to load training pages from version control without a one-off
manual review. This line could now only hurt us.


hectcastro:
Sounds good. Is the current state of training on all of our stacks (staging, training, and production) OK with this not getting run on the next deployment?

kdeloach:
+1

hectcastro:
:+1:

PR #1512, opened 2015-04-30
Fix group attendees count on group detail page.

### LINKED ISSUE #1511: Only one attendee being counted for each event
I set up a couple events for the Staten Island Street Tree Planters group and checked in 3 people for each event. Each time it only counted one attendee, rather than three. The group page updated to 2 events, 2 attendees rather than 2 events, 6 attendees.

maurizi:
Fixes #1511


kdeloach:
+1

PR #1513, opened 2015-04-30
Zoom to deep linked block on reservations

### LINKED ISSUE #1224: Reservations page no longer zooms to blockface in URL hash
You used to be able to click on your reserved blockfaces from the Dashboard, which would load the Reservations page zoomed in to that blockface, and it would also be selected.

maurizi:
Zoom to and preselect blockface on reservations page based on hash

Used on the dashboard to deeplink to reservations.
Connects to #1224


kdeloach:
Want to take the leap and be the first person to use "Connects"?

kdeloach:
+1 tested

RickMohr:
Selecting blocks directly on the Reservations page does not cause a zoom, so I think this should not zoom either.

PR #1515, opened 2015-05-01
Remove grid JSON requests from all JMeter test plans

hectcastro:
This changeset removes all of the UTF grid JSON requests from the JMeter test plans now that we've removed those requests from the map.


steventlamb:
+1, tested in jmeter against develop - the grid requests are removed

Issue #1358, opened 2015-04-22
Implement "event starting soon" revisions


jfrankl:
There are a few changes that address issues people had with "event is starting soon" reminder text. These are in wireframes from March (documents/design/wireframes/2015_03_26_nyc_wireframes_v4.pdf pages 39–43) but I don't think they ever made it into github issues.

## Remove heading reminder

The global heading ( ! ) reminder should be removed. People found it confusing during user testing.

![image](https://cloud.githubusercontent.com/assets/1809908/7278864/075aa00c-e8e6-11e4-94fb-e9d14934fd8d.png)

## Add in context reminders

2 ) We will show notifications for starting soon and in progress events on the dashboard and on the related event page.

### Event Page

*Starting Soon*

![image](https://cloud.githubusercontent.com/assets/1809908/7278961/a85dce20-e8e6-11e4-99db-e5438766fe57.png)

*In Progress*

![image](https://cloud.githubusercontent.com/assets/1809908/7278967/b08db448-e8e6-11e4-8960-c9329093de6e.png)

### Dashboard

*Starting Soon*

![image](https://cloud.githubusercontent.com/assets/1809908/7278983/c3bc22de-e8e6-11e4-872a-14d179d830e8.png)

*In Progress*

![image](https://cloud.githubusercontent.com/assets/1809908/7279002/dd7b91be-e8e6-11e4-8727-c9aaf2260afb.png)

PR #1518, opened 2015-05-01
Replace event starting soon alert based on wireframe revisions.

maurizi:
Connects to #1358


jwalgran:
The banner on the event page says "starting soon" rather than "this event has started" when the event is in progress

This screen shot was taken May 1 at noon.

screen shot 2015-05-01 at 8 59 19 am

jwalgran:
If I click RSVP on an event in progress, I don't see the check in button until I refresh the page. 

screen shot 2015-05-01 at 9 02 21 am

jwalgran:
Edge case. If you remove someone's check in after they have loaded the dashboard/event page, the start mapping button takes them to a generic 403 page. It is not worth a lot of effort to fix this, but if it is possible to easily redirect to the event check in page rather than a forbidden URL it would be a bit more person friendly.

src/nyc_trees/apps/event/templates/event/event_detail.html
```diff
@@ -9,23 +9,22 @@
...
-                        class="btn btn-switch pull-right">Start Mapping</a>
-                    {% endif %}
+    <div class="event-alert">
+        <div class="container">
+            <div class="row">
+                <div class="event-alert-text col-md-8 col-md-offset-1 col-xs-8">
+                    This event is
                     {% if event.in_progress %}
-                      <p>This event is currently in-progress!</p>
+                        starting soon.
```
jwalgran: This conditional is backwards

jwalgran:
The only bit of styling that does not match the wireframes is the lack of vertical centering on the action button.

screen shot 2015-05-01 at 9 33 38 am

src/nyc_trees/apps/home/templates/home/partials/dashboard.html
```diff
@@ -3,6 +3,25 @@
...
 {% url 'events_list_page' as events_page_url %}
 {% url 'progress_page' as progress_page_url %}
+{% if event %}
+    <div class="event-alert">
+        <div class="container">
+            <div class="row">
+                <div class="event-alert-text col-xs-8">
+                    The event {{ event.title }} for {{ event.group.name }} is
+                    {% if event.in_progress %}
+                        starting soon.
```
jwalgran: This conditional is backwards

src/nyc_trees/nyc_trees/context_processors.py
```diff
@@ -32,16 +16,6 @@ def user_settings_privacy_url(request):
 
 
 def config(request):
-    # At the time this function was written, the generated context was
-    # only used in the base.html template, and our AJAX requests all
-    # render partials, not full templates inheriting from base. Given
-    # those constraints, we can skip generating this context, which
-    # saves some database query time.
-    if request.is_ajax():
```
jwalgran: Removing this broke a unit test

maurizi:
@jwalgran I think all of the issues you raised have been addressed, can you take another look?

jwalgran:
The event admin does not see a "Check-in Guests" button unless they RSVP. Not sure if this is by design.

screen shot 2015-05-01 at 10 46 35 am

screen shot 2015-05-01 at 10 47 13 am

jwalgran:
+1 I tried various combinations of user, event time, and RSVP status and saw all the appropriate buttons.

maurizi:
@jwalgran The issue with  "Check-in Guests" is not by design, but is kind of tricky to deal with properly.  I'm inclined to leave it as-is for now.

Issue #709, opened 2015-03-02
Add address search back onto the Survey Map


maurizi:
It's not there in the wireframes, but it could be useful for finding a specific block, especially if you mapped on paper and are re-entering your data on the website.

jfrankl:
Yes, please include the address search.

PR #1528, opened 2015-05-01
Add address search to survey map

kdeloach:
Connects #709


jwalgran:
+1 confirmed that the linter is now happy and the progress page still has full functionality.

jwalgran:
The comment I left was for a different issue, however I did test this and I was able to search for addresses on the survey page.

kdeloach:
Cool, that's what I figured.

Issue #1247, opened 2015-04-17
Unavailable blocks on the reservations page are not styled as unavailable according to the legend


jwalgran:
Legend showing different colors

![screen shot 2015-04-17 at 3 22 34 pm](https://cloud.githubusercontent.com/assets/17363/7211980/d2451cc2-e515-11e4-991a-2b17e77e5731.png)

Single tile

![98519](https://cloud.githubusercontent.com/assets/17363/7211965/ad9f2020-e515-11e4-9eda-a01ca4d6e62f.png)

Grid associated with the tile

![screen shot 2015-04-17 at 3 23 46 pm](https://cloud.githubusercontent.com/assets/17363/7211973/ca6b8edc-e515-11e4-8c88-2ba10213f928.png)

After clicking an unavailable block

![screen shot 2015-04-17 at 3 25 11 pm](https://cloud.githubusercontent.com/assets/17363/7211991/ffc2340a-e515-11e4-9cf9-7fbed5460626.png)

RickMohr:
@jwalgran, I could only reproduce this when logged in as the administrator for the group sponsoring the event.

kdeloach:
I was able to reproduce this issue for trusted mappers and also group admins.

PR #1535, opened 2015-05-01
Fix styling on surveyed blocks

kdeloach:
This fix ensures that blocks do not appear as available on the
reservations page after they have been surveyed.

Connects #1247

### DIFF
--- a/src/tiler/sql/user_reservable.sql
+++ b/src/tiler/sql/user_reservable.sql
@@ -19,7 +19,8 @@
   CASE
     WHEN NOT block.is_available THEN 'unavailable'
     WHEN (core_group.id IS NOT NULL
-          AND NOT core_group.allows_individual_mappers) THEN 'unavailable'
+          AND NOT core_group.allows_individual_mappers
+          AND NOT core_group.admin_id = <%= user_id %>) THEN 'unavailable'
     WHEN reservation.id IS NOT NULL AND reservation.user_id <> <%= user_id %> THEN 'reserved'
     -- Designate as "group_territory" if block belongs to a group and the user
     -- is neither a trusted mapper nor a group admin.
--- a/src/tiler/sql/user_reservable.sql
+++ b/src/tiler/sql/user_reservable.sql
@@ -12,7 +12,8 @@
     -- 'available'.  We use a GeoJSON layer in the UI for the user's
     -- reserved blockfaces, and when a blockface is deselected we want to
     -- be able to reselect it
-    WHEN reservation.user_id IS NOT DISTINCT FROM <%= user_id %> THEN 'available'
+    WHEN (block.is_available
+          AND reservation.user_id IS NOT DISTINCT FROM <%= user_id %>) THEN 'available'
     ELSE 'unavailable'
   END AS survey_type,
   CASE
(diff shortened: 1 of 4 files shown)


src/tiler/sql/user_reservable.sql
```diff
@@ -12,14 +12,15 @@
     -- 'available'.  We use a GeoJSON layer in the UI for the user's
     -- reserved blockfaces, and when a blockface is deselected we want to
     -- be able to reselect it
-    WHEN reservation.user_id IS NOT DISTINCT FROM <%= user_id %> THEN 'available'
+    WHEN (block.is_available
+          AND reservation.user_id IS NOT DISTINCT FROM <%= user_id %>) THEN 'available'
```
maurizi: Are there known cases where a blockface can become unavailable while it is reserved by someone else?

That seems like a bug elsewhere in our code (though sanity checking here doesn't hurt).

kdeloach: Looks like reservations are not being deleted after submitting a survey. Not sure if that is by design or not.

maurizi: Ah yes of course.

kdeloach: Side note: Since I re-ordered these cases it might be easier to read the individual commits instead.

jwalgran:
Looks good. Testing now.

jwalgran:
I think these changes are good, but they do not address #1247, which has to do with the strangeness of the NYC Parks Experts group. This group has ``is_active`` set to ``False`` so that is does not appear on the public groups page. This chunk of the query prevents the ``restriction`` field from being set to `group_territory` 

```sql
    WHEN (turf.id IS NOT NULL
          AND (core_group.is_active = TRUE
               AND trustedmapper.id IS NULL
               AND core_group.admin_id <> <%= user_id %>)) THEN 'group_territory'
```

As a result, you still see the behavior described in #1247, where the line appears "available" but trusted mappers in the NYC Parks Experts group are denied access.

I think it makes sense to also remove the ``core_group.is_active = TRUE`` clause as part of this PR. "is_active" ends up being a bad name for the flag's actual purpose.

kdeloach:
The original issue didn't mention expert mappers so I wasn't aware of that problem. I'll remove `core_group.is_active = TRUE` and then test with an "expert mappers" group next week. Thanks for reviewing.

jwalgran:
I didn't really put it together myself until the other fixes were in.

kdeloach:
I actually experienced a different issue. Trusted mappers in the "expert" group are able to reserve "expert" blocks... but non trusted mappers are also able to reserve these blocks.

kdeloach:
@jwalgran Pushed a new commit to fix the other reservations styling issue.

jwalgran:
Logged in as an individual mapper with no trusted mapper status, I still saw the NYC Parks Experts blocks colored as available. I could not reserve them, but they were blue, not grey, on the reservations map. I added two small commits on top of this PR that color the blocks as unavailable https://github.com/azavea/nyc-trees/commits/fix/jcw/reservations-style-2

I'm turning the review table on you. What do you think of my additions?

kdeloach:
Your changes look good, but since I can't reproduce the problem as you described, I'm not able to test those changes. I'd like to get another pair of eyes on this. @RickMohr?

RickMohr:
+1 for the combination of this PR with the extra commits from @jwalgran. As the administrator of the experts group the expert blocks were blue and I could reserve them. As a non-expert the expert blocks were gray and I could not reserve them. Blocks already mapped and blocks reserved by another user were gray and I could not reserve them.

kdeloach:
After discussing with Rick, it looks like I didn't have any `expert_required` blocks in my system.

kdeloach:
I cherry picked Justin's changes on top of this branch.

kdeloach:
@azavea-bot rebuild

PR #1541, opened 2015-05-04
added max-width to select2 dropdown on tablet on down

### LINKED ISSUE #1539: Shorten width of species autocomplete list
From Parks:

It is also hard to click away from the list, especially on the mobile screen - there is just not enough inactive white space to click on, to make the list go away.  Perhaps this could be addressed by making the field slightly less wide than the screen, so that when the user scrolls down the screen to the next variable they do not accidentally trigger the species list, which is making them frustrated.

Thoughts @designmatty ? This is likely also connected to the other species list issues described in #885

designmatty:
Fixes #1539


jfrankl:
+1

Just the right amount of space.

image

Issue #936, opened 2015-03-24
Create data dictionary and include it in the data dump zip file


jwalgran:

jwalgran:
Related to #866

RickMohr:
One approach would be adding `help_text ` attributes to all model field definitions and using Sphinx to generate documentation.

PR #1543, opened 2015-05-04
Create data dictionary and include it in the data dump zip file

RickMohr:
Connects #936


maurizi:
Migrations need to be generated for this PR

maurizi:
Looking at the data dictionary file, the descriptions for `updated_at` and `created_at` don't appear to be on any of the models except for `Blockface`.

Probably because it is a mixin?  It may just be easiest to remove the mixin and specify those fields on each model, especially since we'll have to do that for any model which needs to create an index on `updated_at` or `created_at`.

maurizi:
Other than the 2 comments I mentioned, this works well.
+1

PR #1545, opened 2015-05-04
Add includes_training field to event feed

### LINKED ISSUE #1457: Add training/mapping field to events feed
Parks would like to have the option to filter the feed and only integrate training events with the Parks calendar.

kdeloach:
This change adds the `includes_training` field to the event feed and
also adds the ablilty to filter the feed by adding querystring arguments.

To show training events:

    /event/feed/?includes_training=True

To show mapping events:

    /event/feed/?includes_training=False

To show all events (omit querystring):

    /event/feed/

You can only filter by the includes_training field for now.

Connects #1457


steventlamb:
+1 tested

kdeloach:
I accidentally put Fixes instead of Connects.

PR #1548, opened 2015-05-04
Show actionbar on progress map when someone interacts with map

jfrankl:
image

image

### DIFF
--- a/src/nyc_trees/sass/partials/_map.scss
+++ b/src/nyc_trees/sass/partials/_map.scss
@@ -15,6 +15,9 @@
 }
 
 .map-container {
+  transition: bottom 400ms;
+  -ms-transition: bottom 400ms;
+  -webkit-transition: bottom 400ms;
   .tab-status-map-tab & {
     position: fixed;
     bottom: 8rem;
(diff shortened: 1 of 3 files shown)


src/nyc_trees/sass/partials/_map.scss
```diff
@@ -15,6 +15,9 @@
 }
 
 .map-container {
+  transition: bottom 400ms;
+  -ms-transition: bottom 400ms;
+  -webkit-transition: bottom 400ms;
```
maurizi: I don't think it's crucial that you remove these for this PR, but as an FYI, there is no need to include these prefixed selectors, since we now run our CSS files through autoprefixer

maurizi:
Since we don't add the `actionbar-triggered` class on any of the maps besides the progress map (because the action bar has elements in it on page load on those maps), this has broken the styling on the event list map and the reservations map.

maurizi:
reservations-map
event-list

jfrankl:
Thanks for catching and helping me fix that error! This should be good to go now.

maurizi:
+1

PR #1549, opened 2015-05-04
Add conditional for no reservations on dashboard

### LINKED ISSUE #1523: Add conditional for no reservations on dashboard
If someone has the ability to make reservations and has no reservations, we should show some text under the reservations headline on the dashboard. Right now it's empty.

@dboyer could you take a shot at writing this copy or should Parks do that?

image

jfrankl:
Fixes #1523. Now help text is displayed under reservations when there are no active reservations.

image


jwalgran:
+1 Confirmed

Issue #1538, opened 2015-05-04
Change species input


dboyer:
1. Add help text in empty species field - "Start typing species name" - Users are not understanding they can type to get the autocomplete list.

2. Only show autocomplete after three letters have been entered. As soon as users see the list, they start scrolling through the options and encounter the species list issues mentioned in #885 or just space issues in general as the species autocomplete list takes up a lot of space on a small screen.  

This can be part of #1431

PR #1552, opened 2015-05-05
Add minimum input and placeholder text to species autocomplete

maurizi:
Connects to #1538

### DIFF
--- a/src/nyc_trees/js/src/surveyPage.js
+++ b/src/nyc_trees/js/src/surveyPage.js
@@ -679,6 +679,8 @@ function setupSpeciesAutocomplete($form) {
 
     $form.find('select[name="species_id"]').select2({
         allowClear: false,
+        minimumInputLength: 3,
+        placeholder: "Select and type species name",
         matcher: function(term, commonName, $option) {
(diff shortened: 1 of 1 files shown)


dboyer:
I'm not sure about the "Select and type species name." Is there a reason to not have just "Start typing species name"? Or "Start typing species and then select from list" ?

RickMohr:
The goal was getting them to tap in the text area rather than hitting the drop-down. I proposed "Tap and type species name". @maurizi preferred "Select and type species name".

maurizi:
@dboyer If you look at the attached screenshots, the placeholder text shows up *before* you have tapped/clicked on the species dropdown, and so in is incorrect to tell people to type to choose a species, since they must open the dropdown first.

There is no option in the library we are using to _easily_ show placeholder text once the dropdown is open (though it does have the "please enter X or more characters" text).

maurizi:
And using "Start typing species and then select from list" both neglects to tell people to open the dropdown first *and* it is too long to fit on small screens

dboyer:
Okay let's stick with "Select and type species name"

steventlamb:
+1 tested

Issue #1349, opened 2015-04-22
Popup on reservations page goes under the legend.


maurizi:
![popup-under-legend](https://cloud.githubusercontent.com/assets/4432106/7277846/3f9db2ca-e8e0-11e4-896a-e3f293a3c745.png)

jfrankl:
This might be impossible to fix without:

1. Changing the layout of the page (using a sidebar instead of an overlay)
2. Moving the leaflet popup outside of the map container in the DOM (not going to happen)

I'm going to look into option 1.

RickMohr:
Another option would be using JavaScript to change the legend's DOM location to be inside the leaflet map.

jfrankl:
@RickMohr that would definitely be my ideal solution. Does that seem feasible? Is it something you all could get to next sprint?

jfrankl:
If not, I'll do the sidebar.

RickMohr:
Moving the legend into the map is straightforward, but we might have to strategize together how to have it continue to show up as a modal for mobile.

![image](https://cloud.githubusercontent.com/assets/2162993/7281207/0dcbbd12-e8f6-11e4-884c-23c56332811b.png)

jfrankl:
Removing design label because Rick is on this.

RickMohr:
We decided to fix this by auto-panning the map to avoid the sidebar. #1559

PR #1559, opened 2015-05-05
When showing blockface popup, auto-pan map to avoid sidebar

RickMohr:
Also fixes a problem where auto-panning was happening in "mobile" view when you selected a blockface even though no popup was shown.

Connects #1349


maurizi:
+1, works great!

RickMohr:
@maurizi, fixed up.

maurizi:
+1, works great! (for real this time)

PR #1561, opened 2015-05-05
Add tiler unit tests

kdeloach:
To test:

    ./scripts/manage.sh test apps.survey.tests.TilerTests


Vagrantfile
```diff
@@ -168,6 +168,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       app.vm.synced_folder "src/nyc_trees/apps", "/opt/app/apps"
     else
       app.vm.synced_folder "src/nyc_trees", "/opt/app/"
+      app.vm.synced_folder "src/tiler/sql", "/opt/tiler/sql/"
```
maurizi: I think this needs a comment to explain _why_ we're syncing the tiler SQL into the app

maurizi:
@kdeloach I think this is headed in the right direction

Vagrantfile
kdeloach: Good idea.

jwalgran:
I like it.

src/nyc_trees/apps/survey/tests.py
```diff
@@ -261,3 +263,68 @@ def _assert_blockface_achievements(self, amount_blockfaces,
...
         self.assertEqual(amount_achievements, len(achievements))
+
+
+class TilerTests(SurveyTestCase):
+    def _exec_tiler_query(self, filename, **kwargs):
+        group_id = kwargs.get('group_id', '0')
+
+        # The path to the tiler SQL queries is defined as a synced
+        # folder in the Vagrant file.
+        path = os.path.join('/opt/tiler/sql/', filename)
```
jwalgran: It would be more conventional to have this path saved in ``settings/test.py`` and ``settings/development.py.``. It behaves sort of like the static assets directory.

src/nyc_trees/apps/survey/tests.py
```diff
@@ -261,3 +263,68 @@ def _assert_blockface_achievements(self, amount_blockfaces,
...
+        # All tiler queries are in the form "(...) AS query" which isn't
+        # a valid query. We need to prepend a select to this subquery
+        # to make it valid.
+        query = 'SELECT * FROM '
+
+        with open(path, 'r') as f:
+            query += f.read()
+
+        # Replace template placeholders.
+        query = query.replace('<%= group_id %>', group_id)
```
jwalgran: A regex match to handle any amount of whitespace after ``<%=`` and before ``%>`` may save us from some head-scratching failures.

hectcastro:
If we plan on having Jenkins run this, it needs to go inside `scripts/test.sh` too.

maurizi:
@hectcastro These are being run by the standard Django test runner, so they're already being picked up without needing to change `test.sh`

src/nyc_trees/apps/survey/tests.py
kdeloach: Moved this path to settings.

src/nyc_trees/apps/survey/tests.py
kdeloach: I'm trying to avoid regex here because in this case I think it would just make things more complicated. Worst case scenario is that the unit tests would fail if there is extra whitespace somewhere.

kdeloach:
I've squashed everything down to a single commit. 

I think that the majority of test cases are covered but I'm still experimenting. I'm open to ideas for test cases and also suggestions on how to improve the clarity of these unit tests.

Some of the results actually surprised me so throughout these unit tests you'll notice some TODO comments. We may want to address those at some point.

kdeloach:
Rebased onto develop.

jwalgran:
+1 tests pass and are well documented.

Issue #1542, opened 2015-05-04
Crosshairs don't allow selecting a block if have chosen to remap it


dboyer:
I mapped a block and on the survey review page, I selected "no, restart." I returned to the first page of the Treecorder and the block edge I wanted to remap was in blue and I had the crosshairs to select it. While I could move the crosshairs, I could not get them to select the block and turn it yellow to begin mapping.

Perhaps related to #1514

Issue #1514, opened 2015-04-30
Treecorder: Reservation under crosshair not selected on page load


kdeloach:
If you only have 1 reservation and you go to the Treecorder, the crosshair appears above the block edge, but it is not selected.

You have to wiggle the crosshairs a little to select it.

![image](https://cloud.githubusercontent.com/assets/43062/7422704/d938884c-ef5c-11e4-8f73-f4d501a0bb57.png)

PR #1562, opened 2015-05-05
Use same logic for selecting blockface from hash as used for crosshairs

maurizi:
Connects to #1542


maurizi:
Waffle

RickMohr:
#1542 appears to be fixed, but not #1514. 

On page load (if there is no deep-linked blockface) we want to see if there is a blockface underneath the crosshairs, and select it if so.

maurizi:
@RickMohr You're right, my bad!  I'll just disconnect that issue from this PR.

RickMohr:
+1

Issue #1443, opened 2015-04-24
Exception in blockedge reservation instructions


hectcastro:
Relevant traceback:

```
Internal Server Error: /blockedge/reservations-instructions/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 87, in wrapper
    RequestContext(request), **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/shortcuts.py", line 25, in render_to_response
    return HttpResponse(loader.render_to_string(*args, **kwargs), **httpresponse_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader.py", line 178, in render_to_string
    return t.render(context_instance)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 148, in render
    return self._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 142, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 844, in render
    bit = self.render_node(node, context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 858, in render_node
    return node.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader_tags.py", line 126, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 142, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 844, in render
    bit = self.render_node(node, context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 858, in render_node
    return node.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader_tags.py", line 126, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 142, in _render
    return self.nodelist.render(context)
[...]

Issue #1440, opened 2015-04-24
Exception in user_detail_redirect


jwalgran:
```
Internal Server Error: /user/profile/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/opt/app/apps/users/views/user.py", line 27, in user_detail_redirect
    reverse('user_detail', kwargs={'username': request.user.username}))
  File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py", line 551, in reverse
    return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
  File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py", line 468, in _reverse_with_prefix
    (lookup_view_s, args, kwargs, len(patterns), patterns))

jwalgran:
I have not been able to reproduce this.

We could try side stepping the issue and linking directly to ``/user/{username}``

hectcastro:
I was able to reproduce this with usernames that have unexpected characters:

- ryan.mcdurmon
- tonie@brazzley530

PR #1563, opened 2015-05-05
Change URLs to accept all valid usernames

maurizi:
I used the same regex that Django uses when validating usernames:
https://github.com/django/django/blob/1.7.8/django/contrib/auth/models.py#L371
so I think this should cover all possible usernames

Connects to #1440
Connects to #1443


jwalgran:
+1

PR #1566, opened 2015-05-06
added no further trees on block action for treecorder

### LINKED ISSUE #1483: Simplify process of adding another tree or submitting data in the Treecorder
From Parks: we also found that users were getting confused with where the "Distance to End" button is located - they automaticaly click "Add another tree" to only then realize there isn't another tree to be added because they are at the end of the block.

Possible change -  Perhaps a way to tweak this would be to give the user two button options at the end of each tree's data entry screen - "Add another Tree" or "End of the Block" so they have to make a deliberate decision whether or not to add another tree.

Thoughts @jfrankl  or @designmatty ?

designmatty:
Fixes #1483


maurizi:
+1

Issue #275, opened 2015-01-09
Add indexes for tiler cache busting


maurizi:
We are using the `updated_at` field of several of our models (currently `Blockface` and `Survey`, but there will almost certainly be more) as a cache buster.  (see https://github.com/maurizi/nyc-trees/blob/3bcc33d8c993fc3211435d1e712030df7e5c457b/src/nyc_trees/apps/core/views.py)

We should add indexes to these fields to ensure they can be queried quickly.

maurizi:
Our cache busters have moved here: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/survey/layer_context.py

maurizi:
One index was added in https://github.com/azavea/nyc-trees/pull/1318, but we still need to add the others.

kdeloach:
We should wait to do this until we optimize tiler performance, since the tables being queries may change

PR #1570, opened 2015-05-06
Add indexes to 'updated_at' for Survey and BlockfaceReservation

maurizi:
These tables are queried to find the most recent update time when
computing tiler cache buster values, when loading that map page.

Survey is queried for the most recent 'updated_at' for every progress page
layer, and BlockfaceReservation.updated_at is queried for every other
map layer.

I am only adding indexes for these 2 table's 'updated_at' columns because
I expect that the number of rows in the other tables we use to calculate
cache busters will remain small, or already have indexes.

I attempted to measure the performance impact of these changes, but the
Postgresql query plan did not use the new indexes.  I suspect this is
because the amount of data in my development DB is rather small.

That said, I expect we will have a large number of rows in these tables
by the end of the census, perhaps numbering in the hundreds of thousands.

Connects to #275

### DIFF
--- a/src/nyc_trees/apps/core/test_utils.py
+++ b/src/nyc_trees/apps/core/test_utils.py
@@ -78,7 +78,8 @@ def make_survey(user, blockface, **kwargs):
     defaults.update(kwargs)
 
     s = Survey(**defaults)
-    s.clean_and_save()
+    s.full_clean()
+    s.save()
 
     return s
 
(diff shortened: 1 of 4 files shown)


maurizi:
Fixing unit tests now

maurizi:
@kdeloach tests fixed

kdeloach:
+1 Everything seemed to work and the migration is backwards compatible.

Issue #1524, opened 2015-05-01
Attempting to add an event with a title over 50 characters long raises an exception


jwalgran:
```
Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)
File "/opt/app/apps/core/decorators.py" in wrapper
  32.         return view_fn(request, *args, **kwargs)
File "/opt/app/apps/core/decorators.py" in wrapper
  58.             return view_fn(request, *args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py" in wrapper
  79.                 params = callable_or_dict(request, *args, **wrapper_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py" in routed
  33.             return req_method(request, *args2, **kwargs2)
File "/opt/app/apps/event/views.py" in add_event
  39.     is_valid = _process_event_form(form, request)
File "/opt/app/apps/event/views.py" in _process_event_form
  72.         event = form.save()
File "/usr/local/lib/python2.7/dist-packages/django/forms/models.py" in save
  455.                              construct=False)
File "/usr/local/lib/python2.7/dist-packages/django/forms/models.py" in save_instance
  103.         instance.save()
File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py" in save
  589.                        force_update=force_update, update_fields=update_fields)
File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py" in save_base
  617.             updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py" in _save_table
  698.             result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py" in _do_insert
  731.                                using=using, raw=raw)
File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py" in manager_method
  92.                 return getattr(self.get_queryset(), name)(*args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py" in _insert
[...]

PR #1574, opened 2015-05-06
Increase max slug length to match connected fields.

maurizi:
Creating an event or group with a name greater than 50 characters would
fail, because the automatically generated slugs were longer than the
default max_length.

Connects to #1524


RickMohr:
+1, tested.

Without this PR I was able to make a group with a 55-character name, but not an event with a 55-character name. With this PR I could make both.

Issue #1104, opened 2015-04-09
Switch tile endpoint to https


dboyer:
satellite on survey confirmation page

PR #1575, opened 2015-05-06
Use https for survey detail map.

maurizi:
Connects to #1104


RickMohr:
+1, tested.

I think your abstractions here are a little complex, but since it's a Pterodactyl project I won't push it. :)

PR #1576, opened 2015-05-06
WIP - remove restriction column

maurizi:
this hasn't been tested yet, but I think this covers everything


maurizi:
This was more complex than I had hoped for.  Closing this in favor of #1580

PR #1578, opened 2015-05-07
PREVIEW Add block edge regions

jwalgran:
This is a proof of concept preview of aggregating and displaying progress as polygons at low zoom levels.

TODO:

- [ ] Pull properly cache busted URLs from the DOM
- [ ] Remove ORDER BY clauses
- [ ] Change legend to match current view
- [ ] Do not show polygon layers on the group progress view

Initial progress view shows boroughs

screen shot 2015-05-06 at 9 35 01 pm

Switches to neighborhoods at zoom 13

screen shot 2015-05-06 at 9 27 51 pm

Switches to blocks at 16

screen shot 2015-05-06 at 8 22 39 pm

I ran a search and replace on the ``all-users-3x-zoom-blockedge-progress.jmx`` to insert polygon layer names in place of the block edge layer name. The results of running the new ``all-users-3x-zoom-polygon-progress.jmx`` against my dev VM show a dramatic increase in performance.

screen shot 2015-05-06 at 8 20 36 pm

```
          Samples Avg     Min Max     Std. Dev.           Err Throughput          KB/Sec              Avg. Bytes
blocks    3270    3312    0   285844  17858.862920433352  0.0 5.702934656925752   309.1862592411086   55516.45749235474
polygons  3270    1017    0   79147   5447.634434160215   0.0 17.27617578376779   175.83484593428713  10422.149235474006
```


jwalgran:
@jfrankl I grabbed an 8-value green color ramp from Color Brewer to bootstrap the styling in https://github.com/azavea/nyc-trees/blob/feature/jcw/add-block-edge-regions/src/tiler/style/default.mss do you have any opinions or suggestions for improvement?

RickMohr:
Looks great.

jfrankl:
Looks awesome Justin.

For the color ramp, we could use the primary green color as the darkest color:

image

['#7ac143', '#8ec95f', '#a1d278', '#b3d990', '#c4e1a9', '#d5eac2', '#e6f1dc', '#f6f9f4']

Or we could use the dark green from color brewer and have it go through the primary green:

image

['#005824', '#266f2b', '#418933', '#5ba13a', '#75bb41', '#a1d178', '#cde6b6', '#f6f9f4']

I don't have a good way to test this, but if you're able to get these on a map, I could take a look. I suspect that the second one will make up better.

jwalgran:
The screenshots have an opacity of 50%, which is going to screw with any ramp we choose. Is there a clever way to "correct" or "offset" a color palate to account for the grey showing through?

jfrankl:
Not that I know of.

Honestly, the colorbrewer colors are fine.

hectcastro:
I took this branch for a spin this morning. Not going to comment on any of the visuals because I don't have much expertise in that stuff. Things are definitely faster than before, mostly because we're doing a lot less work (good!).

I noticed that some of the new queries can get slow because of sorting after `JOINs`. Often times the sorts are so big that they spill onto disk. I increased the PostgreSQL `work_mem` setting from `4MB` (the default) to `10MB`, and that appears to make `+1000ms` queries that were spilling onto disk stay in memory and complete in `150-200ms`:

---

image

jwalgran:
I think the sorting is vestigial at this point. It was carried over from the previous queries and, since we are not filtering or taking values that are dependent on the order of the survey rows, we should be able to remove the ORDER BY.

maurizi:
@jwalgran The sorting is not vestigial.

We need to sort by `Blockface.id` to eliminate duplicate blockface rows, because there may be multiple surveys for a single blockface, and we don't want to count blockfaces as mapped more than once.

jwalgran:
Since we are going to filter out "I could not finish" surveys, doesn't ``DISTINCT ON`` protect us from double counting?

maurizi:
We could drop the sort on `survey.created_at`.  We need to keep the sort on `blockface.id` because we can't use `DISTINCT ON` without it

maurizi:
Superseded by #1608

Issue #1572, opened 2015-05-06
Map tiles are broken on Survey and Group territory maps


maurizi:
Looks to be due to a recent change to CartoCSS:
```
[TILE RENDER ERROR] -- 404: Error: ERROR:  column "restriction" does not exist
LINE 1: SELECT ST_AsBinary("geom") AS geom,"restriction","survey_typ...
                                           ^

Full sql was: 'SELECT ST_AsBinary("geom") AS geom,"restriction","survey_type" FROM (SELECT
  block.id,
  block.geom,
  turf.group_id AS turf_group_id,
  CASE
    WHEN block.is_available THEN
      CASE
        WHEN turf.group_id = 8 THEN 'reserved'
        WHEN turf.group_id IS NOT NULL
          OR reservation.id IS NOT NULL THEN 'unavailable'
        ELSE 'available'
      END
    ELSE
      CASE
        WHEN turf.group_id = 8 THEN 'surveyed-by-me'
        ELSE 'surveyed-by-others'
      END
  END AS survey_type

  FROM survey_blockface AS block
  LEFT OUTER JOIN survey_territory AS turf
    ON block.id = turf.blockface_id
  LEFT OUTER JOIN survey_blockfacereservation AS reservation
    ON (block.id = reservation.blockface_id
        AND reservation.canceled_at IS NULL
        AND reservation.expires_at > now() at time zone 'utc')
) AS query WHERE "geom" && ST_SetSRID('BOX3D(-73.87207031250001 40.41349604942447,-73.08105468749999 41.0130657867288)'::box3d, 4326)'
```

PR #1580, opened 2015-05-07
Fix tile generation several maps.

maurizi:
Styling based on the restriction column was added in 3b1bebf6, but
windshaft bombs out if a property referenced in CartoCSS is not in the
SQL query.

Adding the constant ``'none' as restriction`` to these queries will fix
the issue.

Connects to #1572


jwalgran:
+1 confirmed that all the maps are visible

PR #1585, opened 2015-05-07
rearranged button placement on survery confirmation

### LINKED ISSUE #1540: Make Yes button more prominent on survey review page
Make Yes the first button in the list rather than "No, restart" or "No, submit for review." It can also be a bit larger or otherwise more prominent than the other two buttons.

designmatty:
fixes #1540


jfrankl:
+1

Issue #1560, opened 2015-05-05
Progress page: switching views accumulates click handlers


RickMohr:
1) Go to progress page and open developer tools network tab
2) Click on a blockface. We request `blockedge/near` and `progress-page-blockedge-popup` once each.
3) Change to a different view (e.g. "Group")
4) Click on a blockface. We request `blockedge/near` and `progress-page-blockedge-popup` twice each.

Each time you change the view and click on a blockface we add an additional pair of requests.

PR #1596, opened 2015-05-08
Fix event handling for selecting blockfaces on progress page.

maurizi:
We were attaching duplicate event handlers whenever the mode changed.
This caused two issues: we made duplicate AJAX calls after switching modes,
and we allowed clicking on blockfaces in group mode (which we shouldn't)

Connects to #1560


RickMohr:
+1, tested

Issue #1589, opened 2015-05-07
Training process quiz exception


hectcastro:
Occurs after `/training/progress/intro_quiz/`:

```
Internal Server Error: /training/progress/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app/apps/home/training/types.py", line 193, in wrapper
    score__gte=self.quiz.passing_score))
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 691, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 709, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1331, in add_q
    clause, require_inner = self._add_q(where_part, self.used_aliases)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1358, in _add_q
    current_negated=current_negated, connector=connector)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1225, in build_filter
    lookups, value)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/fields/related.py", line 1561, in get_lookup_constraint
    lookup_class(Col(alias, target, source), val), AND)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/lookups.py", line 82, in __init__
    self.rhs = self.get_prep_lookup()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/lookups.py", line 85, in get_prep_lookup
    return self.lhs.output_field.get_prep_lookup(self.lookup_name, self.rhs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/fields/__init__.py", line 646, in get_prep_lookup
    return self.get_prep_value(value)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/fields/__init__.py", line 915, in get_prep_value
    return int(value)
[...]

PR #1597, opened 2015-05-08
Fix anonymous user exception on training page

steventlamb:
I was able to reproduce this in development.

before:
screenshot from 2015-05-08 12 15 14
after:
screenshot from 2015-05-08 12 16 09

------------------------------------------
connects to #1589 on github

As discussed in person, this is the result of some code assuming an
anonymous user is a logged in user. This commit forces realization
of the user and returns None for the id. The solutions was inspired by
this stackoverflow page:
http://stackoverflow.com/questions/17623234/django-simplelazyobject

This is in response to a bug that turned up on production. Fortunately
the only time this bug would turn up is with a case in which a 404
appropriate. This fix makes sure that a 404 is returned.


hectcastro:
:+1: 

Open a new PR against the release branch?

Issue #1582, opened 2015-05-07
Quit surveys are shown as mapped


maurizi:
Steps to reproduce:
1. Start mapping a block, but quit by choosing "I cannot map this block because"
2. Go to the progress map.
3. Note that the block you quit mapping is marked as "mapped"

maurizi:
We probably need to audit our tiler and contribution queries to ignore any surveys that have a `quit_reason`

Issue #1573, opened 2015-05-06
Users still have a blockface reserved after mapping if the block is later marked as available


maurizi:
Scenario 1:
1. A user reserves a blockface, maps it, and then chooses "I'm done for now"
2. Later after QA/QC a census admin determines that the block should be remapped, and marks it as available
3. It is now shown as reserved by the user who originally mapped it.  Since they may have made additional reservations in the mean time, this can push that user over the 12 blockface reservation limit.

Scenario 2:
1. A user reserves a blockface, maps it, and then chooses "No, restart"
2. It is still reserved by that user, and they can remap it immediately.

Scenario 2 seems correct, but scenario 1 does not.  I propose that when surveying a blockface *as an individual mapper*, when a user completes a survey we mark their `BlockfaceReservation` as "expired".  If they then choose "No, restart", we should give them a new reservation.

kdeloach:
Alternatively, if the user selects "Yes" on the "Does this look right?" question, we could mark the reservation as expired then.

steventlamb:
We've gone back and forth about what should and shouldn't happen on the survey confirmation page, and it was decided, against what would be intuitive, to make the 'yes' button on the confirmation page essentially do nothing and treat the confirmation page as merely an opportunity to flag. This is confusing as it is, and modifying it slightly to mark the reservation as expired but *not* modify the availability of the blockface could make it more-so because these data points should ideally be connected.

The former solution has the advantage of being simpler, the latter is easier.

dboyer:
To confirm, @maurizi 's first scenario is definitely incorrect. If a census admin marks a block as available again, it should not be reserved for anyone. The expected use case is that Parks will mark it as available based on QA/QC and then reserve it for a Parks mapper to complete.

kdeloach:
We discovered another issue (regression?). If the user quits the survey for any reason "I cannot map this block because..." the block remains reserved by that user. Whatever solution we implement for this issue should also address this scenario as well.

dboyer:
I believe that the block should not be made available again or marked as reserved by that user if they select "I cannot map this block because..." The use case is blocks that are under construction, have hazards, or are otherwise inaccessible. Parks will check the data for a quit reason and make the block available again if necessary. I can confirm that though.

steventlamb:
@dboyer @kdeloach please see the `PROPOSAL` commit linked immediately above this message.

dboyer:
I emailed Parks about this but have not heard back yet.

[...]

PR #1598, opened 2015-05-08
Only count completed surveys when determining progress and map styling

maurizi:
Connects to #1582

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -216,7 +216,8 @@ class User(NycModel, AbstractUser):
     def surveys(self):
         """Return surveys user has participated in"""
         from apps.survey.models import Survey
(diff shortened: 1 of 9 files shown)


maurizi:
Some unit tests are failing, I'll reassign this once it is ready

maurizi:
Unit tests fixed

kdeloach:
I can't reproduce the original issue. If I select "I cannot map this block because..." and enter a quit reason, I still have the block reserved.

maurizi:
@kdeloach As we discussed, this does not fix issues around whether the blockface is still consider reserved, only what color it is styled as.  Reservation issues will be handled as part of a different PR.

kdeloach:
I was confused by another issue #1573. After testing again, it appears to be working.

kdeloach:
+1

Issue #1584, opened 2015-05-07
Send email after user is checked into training event


dboyer:
When user is checked into an event, TreesCount sends an email reminding them to attend future events and includes a link to a survey.

jwalgran:
@dboyer copy?

dboyer:
Subject: Thank you for attending a TreesCount! 2015 training event

Hi Volun<b>treer</b>!

Now that you are <b>Treerifically Trained</b> you will be able to sign up for and attend mapping events held by groups! 

Remember, training events always start with a refresher on how to collect the data and use the app. You are welcome to attend additional training events at any time, and will still be able to sign up for these events if you would like a refresher. 

At a mapping event, you will be able to pick up equipment and start mapping right away! Staff and groups will be on hand to give you equipment and help troubleshoot any problems, but at mapping events there is no time specifically set aside as a training refresher.

We would love to hear your feedback on the volun<b>treer</b> experience! Please follow <a href= "https://www.surveymonkey.com/s/5HT3B59" this link</a> to fill out a quick survey which will help us target our engagement efforts and shape future events.

Thank you for helping count, map and care for NYC’s street trees!

steventlamb:
I don't think we can bold plain text emails, or put hyperlinks. Mind if I rephrase the link paragraph to:
```
We would love to hear your feedback on the voluntreer experience! Please follow this link to fill out a quick survey which will help us target our engagement efforts and shape future events:
https://www.surveymonkey.com/s/5HT3B59
```

dboyer:
Can the survey monkey URL be linked at least? Or no links at all?

maurizi:
@dboyer We can't put hyperlinks.  Most email clients will automatically turn a URL into a link though.

dboyer:
That's fine then.

PR #1601, opened 2015-05-08
Send email after event checkin

steventlamb:
connects to #1584 on github

Not sure if this is sufficient. If the user keeps clicking the check-toggle button - they're gonna send a *lot* of emails.


jwalgran:
The email should ``{% extends "mail/base_email.txt" %}`` and wrap the body with ``{% block content %}`` to ensure that the standard email footer text is included.

jwalgran:
Confirmed that this sends an email only when checking in a user to their first training event.

Issue #1568, opened 2015-05-06
Checkin page (via "See all" RSVPs link) allows premature check-in


RickMohr:
1) Login as a group administrator and go to the check-in page of an event not currently in progress.
1a) You can get there by going to the event page and adding `/checkin/` in the address bar
1b) You can also get there via the "See all" RSVP's link on the email attendees page.

You can check people in (or un-check them in), and increase the RSVP count even though the event is not in progress.

I think you should only be able to change people's check-in status or increase the RSVP count if the event is in progress.

Also, it's pretty strange to have "check-in" links at all when you click "See all" RSVP's on the email attendees page.

PR #1602, opened 2015-05-08
Prevent checking in users until an event is "starting soon".

maurizi:
Connects to #1568

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/admin_checkin.html
+++ b/src/nyc_trees/apps/event/templates/event/admin_checkin.html
@@ -43,9 +43,11 @@
         <div class="row">
         {% endif %}
             <div class="col-xs-12">
-                <div class="btn-checkin-container pull-right">
-                {% include "event/partials/checkin_button.html" %}
-                </div>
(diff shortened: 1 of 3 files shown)


RickMohr:
The check-in links are gone, but there are still free things about the check-in page that don't make sense for the "See all RSVPs" link:
* The title of the page is "Check-in"
* The "+" button to up the RSVP limit is still active
* There is a "Start Mapping" button

Rather than trying to re-use the checkin page, why not handle "see all RSVPs" with a (scrolling) modal?

dboyer:
What do we define as "starting soon?" How long before the event will users be able to check-in RSVPs?

maurizi:
@dboyer The same time we use for when to show the blue alert bar - 1 hour before the event starts

dboyer:
That's perfect. Just wanted to make sure to include that in the group admin guide.

maurizi:
@RickMohr I've updated this with your suggestion of showing a modal instead of linking to the admin checkin page

RickMohr:
+1, tested. Thanks for the scrollbar.

Issue #1591, opened 2015-05-07
Remove loading blockface fixtures from setupdb.sh


jwalgran:
We do not want to clobber is_available

PR #1603, opened 2015-05-11
Remove loading blockfaces from setupdb.sh

jwalgran:
For a few reasons

- The database for every stack already contains these blocks, and they
  will not be changing over the course of census. We will only be
  getting additions.

- The block rows have a mutable ``is_available`` column used for
  tracking the status of the row, and we do not want to overwrite it.

- It slows down the deploy process.

Connects to #1591


kdeloach:
+1

steventlamb:
+1

Issue #1581, opened 2015-05-07
"My Progress" map shows no tiles


RickMohr:
1) Go to https://treescount.azavea.com and log in as rmohr / rmohr
2) Click "MY TREES" and note that this user has mapped 2 blocks
3) Go to the Progress page 
4) Open the developer tools, Network tab, filter by "png", and click the clear button
5) Choose "My Progress" from the drop-down

The map zooms to the extent of the mapped blocks, but no map tiles are displayed. The network tab shows that only base map tiles were requested, no blockface tiles.

This problem does not happen if logged in as admin / admin.

hectcastro:
Reproduced following the steps above. Adding a note that switching to "All Progress" and back to "My Progress" does fire off new requests.

PR #1604, opened 2015-05-11
Fix "My Progress" map tiles not appearing the first time

kdeloach:
This fixes a race condition where the map `zoomend` event would not fire the
first time a user selects "My Progress" on the Progress map page.

I solved this by replacing the zooming event handlers with a promise.

Connects #1581

### DIFF
--- a/src/nyc_trees/js/src/map.js
+++ b/src/nyc_trees/js/src/map.js
@@ -93,13 +93,26 @@ function createAndGetControls(options) {
     return {map: map, multiControl: multiControl};
 }
 
+// Return a promise which is resolved when map is done zooming.
 function fitBounds(map, bounds) {
+    if (!bounds) {
+        return $.Deferred().resolve().promise();
+    }
+
     // GeoDjango bounds are [xmin, ymin, xmax, ymax]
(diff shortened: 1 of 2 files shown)


src/nyc_trees/js/src/map.js
```diff
@@ -188,28 +201,10 @@ function addGridLayer(map, options) {
...
-    if (waitForZoom) {
-        _addAfterZoom(map, layer);
-    } else {
-        map.addLayer(layer);
-    }
-}
-
-function _addAfterZoom(map, layer) {
-    // Add layer to map after zoom animation completes.
-    // (Otherwise spurious tile requests will be issued at the old zoom level.)
```
RickMohr: Consider restoring this comment to https://github.com/azavea/nyc-trees/pull/1604/files#diff-2780080b0b4bcd298f4a429c851f2f8bR128

RickMohr:
+1, tested. Nice solution.

src/nyc_trees/js/src/map.js
kdeloach: Thanks, I shouldn't have removed this.

PR #1606, opened 2015-05-11
Reservations: don't generate map tiles at low zooms

RickMohr:
Connects #1593

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
+++ b/src/nyc_trees/apps/survey/templates/survey/reserve_blockface.html
@@ -20,7 +20,12 @@
                   </div>
               </div>
               <div id="reservations-help-text" class="col-xs-7 text-left">
-                  Click on a block edge to add or remove it from your reservations
+                  <div class="js-zoomed-in-help">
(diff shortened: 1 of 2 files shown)


jwalgran:
+1 this works perfectly. I played with the styling a little to try and direct peoples attention to go to the action bar. What do you think about bold and blue for the low zoom instructions?

screen shot 2015-05-11 at 1 01 38 pm

RickMohr:
Good idea to make that text more noticeable. Blue might not be best though since we use it for links a lot of places. @designmatty, what do you think?

designmatty:
Two thoughts:

1: I think $brand-primary (green) would work better since like @RickMohr said, blue is reserved for links.
2: What if you only made changes to the "Zoom In" text? Bolded it and made it green instead of blue and left the rest of the text as is. Just a thought.

RickMohr:
@designmatty, how would I best apply that? From the source it looks like `label-primary` should work, but `<span class="label-primary">Zoom in</span>` doesn't change the color.

RickMohr:
Somewhat more noticeable:
capture

jwalgran:
I like just highlighting the action a lot.

Issue #1565, opened 2015-05-06
Email content showing escaped characters


jwalgran:
When rendering plain text emails, HTML escaping needs to be disabled

![screen shot 2015-05-06 at 8 00 38 am](https://cloud.githubusercontent.com/assets/17363/7495974/35d770be-f3c6-11e4-8beb-4539be8e9997.png)

steventlamb:
relevant: http://stackoverflow.com/questions/8774902/how-to-prevent-auto-escape-in-django-templates

jwalgran:
also relevant https://github.com/azavea/nyc-trees/pull/1406/files

steventlamb:
True, but with the way we send emails being kinda frameworky, it probably makes sense to abstract this too.

PR #1607, opened 2015-05-11
Disable autoescape in base text email template

jwalgran:
The content of a plain text email never needs to be escaped. The Django docs explain that autoescape settings are passed from parent to child templates.

https://docs.djangoproject.com/en/1.7/topics/templates/#for-template-blocks

Connects to #1565


kdeloach:
+1 Tested

PR #1608, opened 2015-05-11
Finish displaying aggregated block edge region on progress page

maurizi:
Improves performance at lower zoom levels.

Build on top of #1578

Connects to #1594


maurizi:
Waffle

src/nyc_trees/apps/survey/models.py
```diff
@@ -13,6 +13,18 @@
 from libs.mixins import NycModel
 
 
+class Borough(NycModel, models.Model):
+    geom = models.MultiPolygonField()
+    name = models.CharField(max_length=32)
+    code = models.IntegerField(db_index=True, unique=True)
```
jwalgran: We don't use ``code`` in any sorts or joins (We did in an early draft of the feature) so we can drop these indexes.

maurizi: Will do.  I assume the same goes for `NeighborhoodTabulationArea.code`?

jwalgran: Yes.

jwalgran:
Some new validation errors are failing a few unit tests. I suspect the blockface foreign keys need ``blank=True`` in addition to the existing ``null=True``

```
  File "/opt/app/apps/survey/views.py", line 595, in _mark_survey_blockface_availability
    survey.blockface.full_clean()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 1012, in full_clean
    raise ValidationError(errors)
ValidationError: {'nta': [u'This field cannot be blank.'], 'borough': [u'This field cannot be blank.']}

jwalgran:
Bug:

1. Load the progress page
2. Switch to "Groups"
3. Switch back to "All Users"

An exception is thrown that prevents the mode change.

screen shot 2015-05-11 at 2 23 34 pm

jwalgran:
The color ramp labels extend outside the mobile legend popup area

screen shot 2015-05-11 at 2 29 38 pm

@designmatty some general style feedback would be helpful, too.

maurizi:
@jwalgran This has been rebased on develop and all the issues you noted have been addressed

jwalgran:
+1 works

I like the look of a 1px #eee border around the ramp row instead of a border around each cell along with removing the right and left padding from the 100 and 0 respectively to push the labels out to the edges.

screen shot 2015-05-12 at 8 48 32 am

maurizi:
@jwalgran I think that looks nicer too, but I'll leave it up to design.  Getting those ramps to be properly responsive was tricky, and I don't want to mess it up by mucking with it any further.

dboyer:
I'm planning to ask Jeff to review the progress map once it is on staging so he can comment on the color ramp border.

PR #1609, opened 2015-05-11
Bug/1573/clear blockface reservation when complete

steventlamb:


jwalgran:
+1 I have confirmed the changes in behavior. If I survey a reserved block, and then set ``is_available`` on that block back to ``True``, the block is available for reservation again. If I quit a survey with a reason, that block is no longer available.

Issue #1564, opened 2015-05-06
Event registration failure


hectcastro:
Appears to be occurring on a `POST`:

```
Internal Server Error: /group/prospect-park-alliance/event/gBSmYCaSPP7rcx8wGLVdNH/register/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 32, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 69, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 79, in wrapper
    params = callable_or_dict(request, *args, **wrapper_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/usr/local/lib/python2.7/dist-packages/django/db/transaction.py", line 394, in inner
    return func(*args, **kwargs)
  File "/opt/app/apps/event/views.py", line 289, in register_for_event
    EventRegistration.objects.create(user=user, event=event)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 372, in create
    obj.save(force_insert=True, using=self.db)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 589, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 617, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 698, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
[...]

PR #1610, opened 2015-05-11
Prevent possible IntegrityError triggered by clicking RSVP button

kdeloach:
Change this query from a `create` to a `get_or_create` so that Django
does not bother inserting a new EventRegistration record (and possibly
raising an IntegrityError) if a record already exists.

**Note:** It seems like this error is a product of having load balanced app servers which is difficult to test on our local (non load balanced) dev instances. Since I'm not able to reproduce this error, I can't guarantee for sure that this actually resolves the issue.

Connects #1564


maurizi:
+1

Issue #150, opened 2014-12-17
Checked-in users can continue mapping in a group's area until midnight the day of the event


dboyer:

dboyer:
@jwalgran Parks just confirmed that midnight the day of the event is the cutoff for all data entry, even if they are mapping on paper. They don't want data entry to linger for days after an event.

maurizi:
@dboyer How do they access the survey page? Should they treecorder link in the nav still go to the event until midnight the day of?

dboyer:
Good question. Yes, having the Treecorder link go to the location of the event until midnight seems logical.

PR #1613, opened 2015-05-12
Allow mapping at events all day

maurizi:
- Refactors treecorder link to show a modal when there is ambiguity about how to survey
- Allows mapping events until midnight the day of the event, if the user was checked in to that event

Connects to #150

### DIFF
--- a/src/nyc_trees/apps/core/templates/base.html
+++ b/src/nyc_trees/apps/core/templates/base.html
@@ -40,6 +40,9 @@
     {# This block exists for placing elements that need to be children of the body, like bootstrap modals #}
     {% block extra_content %}
     {% endblock extra_content %}
+    {% if treecorder_urls|length > 1 %}
+      {% include "home/partials/treecorder_modal.html" %}
+    {% endif %}
 
     {% block footer %}
         <footer>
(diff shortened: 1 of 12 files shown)


maurizi:
@dboyer Let me know what you think of the text here:
modal1

dboyer:
Let's change "how do you want to map" to "where do you want to map." When someone reserves block edges, they are asked about how they want to map in terms of paper vs. app so I'd rather use where here, not how.

Also, take out the comma after event. Otherwise, I think it's good.

kdeloach:
Should we also update the dashboard to show a notification for each event in progress? Right now it only shows the first one.

image

kdeloach:
I have to click links in the modal window twice for it to work. The first time I click the nav deactivates.

image

image

dboyer:
Ideally, we would show the notification for each event in progress, but do we have space for that?

maurizi:
@dboyer I don't think we have the space for it, especially on mobile.  @jfrankl or @designmatty could weigh in, but I remember discussing showing multiple event alerts there before and ditching the idea.

maurizi:
@azavea-bot rebuild

RickMohr:
I think it should say "Map my reserved block edges" instead of "Map your reserved block edges"

maurizi:
@kdeloach The modal nav menu issue we discussed is fixed

RickMohr:
Otherwise +1. Tested a lot of cases and they all worked.

maurizi:
@azavea-bot rebuild

kdeloach:
+1

dboyer:
I'm fine with only showing one notification for an event starting soon.

PR #1615, opened 2015-05-12
Update layout boxes fn

maurizi:
Fixes crashes when trees extend beyond the end of the blockface polyline.

Based on top of #1614 

Connects to #838


steventlamb:
+1 tested

PR #1616, opened 2015-05-12
Sticky tree headings

designmatty:


jfrankl:
+1

jfrankl:
This is fantastic.

maurizi:
This looks amazing!  Great work @RickMohr and @designmatty!

src/nyc_trees/sass/partials/_survey.scss
```diff
@@ -254,8 +316,8 @@
     background: #f3f3f3;
     margin-bottom: -10px;
     position: absolute;
-    left: 0;
-    right: 0;
+    left: -2rem;
+    right: -2rem;
```
maurizi: Changing these 2 lines introduces a slight regression: it causes a horizontal scroll on the treecorder for mobile.

RickMohr:
In a narrow chrome desktop browser the expand/collapse arrow is cut off by the scrollbar and a slightly different height:
capture

RickMohr:
When just the bottom of Tree 1 is visible I see this:

capture

Then scrolling down slightly, this:

capture

Is that as intended?

If I then collapse Tree 1 I see this:
capture

That's not correct, since we're looking at the middle of Tree 2. I can fix it in the JavaScript.

designmatty:
@RickMohr What if we did something like this to prevent users from collapsing a tree like you did. 
fix-sticky

designmatty:
@RickMohr What if we did something like this to prevent users from collapsing a tree like you did. 
fix-sticky

designmatty:
@RickMohr This sprint I have a very small amount of time left to work on NYC. For my time constraints I made desktop scrolling not stick (exception: when the user resizes their browser to mobile sizes). I did make sure the sticky elements are not covered by scrollbars on mobile. 

Since, I have little time left and have another task to do this sprint I recommend holding off on fixing the windows chrome desktop scrollbar issue until next sprint.

RickMohr:
@designmatty, yes, a narrow desktop browser window seems like an uncommon case. I was using that to test because in device emulation mode on chrome I can't get the map to scroll and so can't select a block in the Treecorder.

RickMohr:
@designmatty, I fixed the collapse issue we were discussing above.

PR #1617, opened 2015-05-12
Fix 500 error when emailing event registrants

kdeloach:
The `from_email` argument should not be a list.

To test, send an email to event registrants and inspect the console output:

Right:
image

Wrong:
image

Connects #1519


maurizi:
+1, works

Issue #1611, opened 2015-05-11
Production progress map: groups view takes 5 seconds to load


RickMohr:
1) Go to https://treescount.nycgovparks.org/blockedge/progress/
2) Choose "Groups" from the drop-down

The request `https://treescount.nycgovparks.org/blockedge/group-borders.json` takes 5 seconds.

Two questions:
1) Why is fetching a few dozen polygons not extremely fast?
2) Nothing is changing, so why is the result not cached?

kdeloach:
This takes ~15s to load on my dev instance just to return an empty data set.

hectcastro:
I'll take a look at this today.

maurizi:
My guess (which is unverified) is that we're making DB queries in a loop to get the group territory.

I fought the Django ORM a lot to try to prevent that, so if it is the case the easiest solution will probably be to switch to raw SQL.

hectcastro:
Of the queries produced to generate the group view, one stands out because it returns 5-6MB of data (based on production data):

```sql
SELECT "survey_territory"."id", 
       "survey_territory"."created_at", 
       "survey_territory"."updated_at", 
       "survey_territory"."group_id", 
       "survey_territory"."blockface_id", 
       "survey_blockface"."id", 
       "survey_blockface"."geom", 
       "survey_blockface"."is_available", 
       "survey_blockface"."expert_required", 
       "survey_blockface"."source", 
       "survey_blockface"."created_at", 
       "survey_blockface"."updated_at" 
FROM   "survey_territory" 
       INNER JOIN "survey_blockface" 
               ON ( "survey_territory"."blockface_id" = "survey_blockface"."id" 
                  ) 
WHERE  "survey_territory"."group_id" IN ( 24, 2, 3, 26, 
                                          4, 27, 5, 6, 
                                          7, 8, 9, 10, 
                                          28, 11, 12, 30, 
                                          13, 31, 25, 32, 
                                          14, 29, 15, 16, 
                                          17, 18, 33, 19, 
                                          20, 21, 22, 34, 23 ) 
```

After that, getting the extent for each group's `GeometryCollection` in `group_borders_geojson` takes a while because each groups turf contains a bunch of elements.

None of this is getting cached because the response is coming back directly from Django with no cache headers. I tried using the `cache_control` decorator along with `group_borders_geojson`, but I haven't been able to figure out how to make it work with our function based views:

```python
group_borders_geojson = route(GET=do(json_api_call,
[...]

PR #1619, opened 2015-05-12
Improve group geojson performance

maurizi:
Improves performance by removing an unecessary JOIN to the territory table, and adds a last modified header for caching.

Connects to #1611.

### DIFF
--- a/src/nyc_trees/apps/survey/layer_context.py
+++ b/src/nyc_trees/apps/survey/layer_context.py
@@ -83,6 +83,10 @@ def get_context_for_territory_admin_layer(group_id):
                                   {'group': group_id})
 
 
+def get_group_territory_modification_time(request):
+    return _get_max_modification_time([Group])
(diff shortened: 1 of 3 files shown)


kdeloach:
+1 Selecting "Groups" from the progress page dropdown used to take ~15 seconds and now it takes ~50MS.

src/nyc_trees/apps/survey/layer_context.py
```diff
@@ -83,6 +83,10 @@ def get_context_for_territory_admin_layer(group_id):
                                   {'group': group_id})
 
 
+def get_group_territory_modification_time(request):
```
hectcastro: This doesn't seem to account for scenarios where groups get deleted, or when borderers are edited. Since the request is quick after removing the `JOIN`, it seems fine to leave it as is without introducing `Last-Modified`.

PR #1622, opened 2015-05-12
Send a copy of bulk event emails to group admin

kdeloach:
This adds the current user and group admin to the list of recipients
for bulk event emails.
 
Connect #241


kdeloach:
@azavea-bot rebuild

jwalgran:
+1 I confirmed that the ``set()`` ensures that the same person does not show up in the recipients list more than once and that messages are sent to both the logged in user and the and group admin, in addition to the registrants.

PR #1626, opened 2015-05-12
Always create event map PDF when editing an event

kdeloach:
Occasionally, I will run into an issue on my local dev instance where
the event map PDF fails to generate, which prevents me from RSVPing to
events. One way to trigger this issue is by manually deleting your static
folder.

Although not likely to be an issue on production, I don't think it's a
bad idea to have a workaround in place for regenerating event map PDFs.

### DIFF
--- a/src/nyc_trees/apps/event/views.py
+++ b/src/nyc_trees/apps/event/views.py
@@ -59,20 +59,11 @@ def _process_event_form(form, request, event=None):
         # Lat/lng were not submitted or could not be converted to floats.
         pass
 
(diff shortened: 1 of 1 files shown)


kdeloach:
What do you think about this change?

RickMohr:
@kdeloach, if the point is to have a workaround to regenerate event map PDFs in unusual situations, I don't think we need this because we already have the workaround of changing the event location.

kdeloach:
Right, good point! Closing.

Issue #1129, opened 2015-04-11
Enhance health check endpoint for tiler


hectcastro:
Currently the tiler health check targets `/`. This doesn't really exercise the tiler application much and doesn't give us a good representation of the service's health.

kdeloach:
Any suggestions for metrics to add to the tiler health check endpoint?

hectcastro:
Main objective here was to find something better than `/` to use for the health check. If Windshaft can't talk to Redis or PostgreSQL, hitting `/` still returns a 200 with useless text. Windshaft exposes a `/version` endpoint too, but that behaves the same way.

There may be something [around here](https://github.com/CartoDB/Windshaft/blob/0.19.4/lib/windshaft/server.js#L389) that we can reuse. All I really care about is knowing when the tiler can't render tiles.

jwalgran:
Yup, I am attaching a route that just opens test connections to both postgres and redis, using the client libraries that are already installed under windshaft.

jwalgran:
On staging the postgres connection is failing.

```javascript
{
  "cache": {
    "ok": true
  },
  "database": {
    "error": {
      "name": "error",
      "length": 86,
      "severity": "FATAL",
      "code": "3D000",
      "file": "postinit.c",
      "line": "775",
      "routine": "InitPostgres"
    }
  }
}

jwalgran:
Assigned to hectcastro while he is investigating the failure

jwalgran:
The staging load balancer it showing ``”ok”: true`` for the database.

PR #1627, opened 2015-05-12
Add a health check endpoint to the tiler

jwalgran:
We want to be able to detect connectivity issues between the tiler and the two services it depends on, the Postgres database and the Redis cache. This commit adds a ``/health-check`` endpoint that attempts to connected to both services and returns the status of those connections in a JSON response.

Test by stopping the Postgres and Redis services and verifying that ``/health-check`` returns an appropriate message.  You can also hack the code or set ``NYC_TREES_TILER_HEALTH_TIMEOUT=1`` to test the timeout functionality. 

Connects to #1129


hectcastro:
:+1: 

Also, I tacked on a commit that makes the ELB use this new endpoint for health checks.

Issue #1632, opened 2015-05-13
Data dictionary export task fails on borough_id


jwalgran:
```
[2015-05-13 13:12:20,686: DEBUG/MainProcess] Task accepted: apps.census_admin.tasks.write_data_dictionary[1dc0ef98-a427-4129-891f-e85c7c4d109c] pid:30283
[2015-05-13 13:12:20,693: ERROR/Worker-1] Chord u'8f1b7d49-385a-47b2-9aec-6c6c0c97edd2' raised: 'Dependency 1dc0ef98-a427-4129-891f-e85c7c4d109c raised KeyError(u"u\'borough_id\'",)'
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/backends/base.py", line 561, in on_chord_part_return
    ret = j(timeout=3.0, propagate=propagate)
  File "/usr/local/lib/python2.7/dist-packages/celery/result.py", line 691, in join_native
    raise value
KeyError: u'borough_id'
[2015-05-13 13:12:20,698: ERROR/MainProcess] Task apps.census_admin.tasks.write_data_dictionary[1dc0ef98-a427-4129-891f-e85c7c4d109c] raised unexpected: KeyError(u'borough_id',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/census_admin/tasks.py", line 230, in write_data_dictionary
    models = [_model_context(fq_name) for fq_name in fq_names]
  File "/opt/app/apps/census_admin/tasks.py", line 252, in _model_context
    for field in Model._meta.fields]
  File "/opt/app/apps/census_admin/tasks.py", line 260, in _field_context
    name = _shapefile_fields[model_name][name]
KeyError: u'borough_id'

PR #1633, opened 2015-05-13
Add new blockface fields to data dictionary mapping.

maurizi:
Fixes the dump_db command, which was failing on the data dictionary step.

Connects to #1632


jwalgran:
+1 task now completes successfully

PR #1638, opened 2015-05-13
Don't include unnecessary CSV fields.

maurizi:
Connects to #1631

### DIFF
--- a/src/nyc_trees/apps/core/management/commands/dump_db.py
+++ b/src/nyc_trees/apps/core/management/commands/dump_db.py
@@ -9,17 +9,26 @@ from django.core.management.base import BaseCommand
 from django.db import transaction, connection
 from django.utils.timezone import now
 
-from apps.core.models import TaskRun
+from apps.core.models import TaskRun, User, Group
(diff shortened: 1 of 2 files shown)


src/nyc_trees/apps/core/management/commands/dump_db.py
```diff
@@ -63,7 +74,10 @@ def handle(self, *args, **kwargs):
 
         dump_id = today.isoformat()
 
-        dump_model_tasks = [dump_model.s(fqn, dump_id) for fqn in CSV_MODELS]
+        dump_model_tasks = [dump_model.s(fqn, None, dump_id)
+                            for fqn in CSV_MODELS]
+        dump_model_tasks = [dump_model.s(fqn, fields, dump_id)
```
jwalgran: Looks like this should be ``dump_model_tasks += ...``

maurizi: Fixed

jwalgran:
+1 This removes the unnecessary columns and all the files are still being generated.

``doc.html`` still has references to the nonexistent fields. I am fine with either tackling it as part of this PR, or opening a separate issue.

PR #1643, opened 2015-05-13
Fix tiler health check by threading through database name

hectcastro:
In EC2, the RDS user and database name do not match. These changes pass through the backend database name so that the health check can make use of it.

Connects #1129


jwalgran:
+1

Tested by

1. Provisioning my dev tiler
1. Editing ``/etc/nyc-trees.d/env/NYC_TREES_DB_NAME`` to be the wrong database name
1. Launching the tiler with ``envdir`` and seeing the error message returned from /health-check
1. Editing ``/etc/nyc-trees.d/env/NYC_TREES_DB_NAME`` to have the correct database name
1. Launching the tiler with ``envdir`` and seeing success returned from /health-check

Issue #1530, opened 2015-05-01
Restrict the functionality of banned users and show ambassador status on check-in page


jwalgran:

RickMohr:
Currently our User model has unused fields `is_flagged`, `is_banned`, and `is_ambassador`. We should either do something with them or remove them.

dboyer:
Parks would like the ban option. Waiting to hear back from them on flagged and ambassador.

jwalgran:
ambassador and flagged are for reporting, but won't be shown in the UI

dboyer:
Parks confirmed the following info:

- Checking the banned box on the Django admin page should prevent that user from being able to log into TreesCount.

- Checking the flagged box should cause a flag on the user table. Parks will use that data, joined to the survey table, to see all surveys added by flagged users.

- If the ambassador flag is checked, an "ambassador" note should be next to that user's username on the event check-in page. The ambassador designation does not need to be displayed anywhere else. If group admins know who the ambassadors are, they can designate those people to serve as assistants to more novice mappers.

dboyer:
To clarify, displaying the ambassador status on the event check-in page can just be the word "Ambassador" under that person's user name. There is no need to have a filter by ambassador function or special icon or anything like that.

PR #1646, opened 2015-05-13
Restrict access from banned users

kdeloach:
Check if user is banned before processing any views. If they *are*
banned, immediately logout user.

How to test:
  1. Login as any user
  2. Set `is_banned` to True for that user
  3. Refresh the page
  4. You be logged out now, and unable to login until the ban is lifted

Connects #1530

### DIFF
--- a/src/nyc_trees/nyc_trees/middleware.py
+++ b/src/nyc_trees/nyc_trees/middleware.py
@@ -7,6 +7,7 @@ import re
 import waffle
 
 from django.conf import settings
+from django.contrib.auth import logout
 from django.shortcuts import redirect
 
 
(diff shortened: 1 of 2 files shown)


src/nyc_trees/nyc_trees/middleware.py
```diff
@@ -25,3 +26,13 @@ def process_view(self, request, view_func, view_args, view_kwargs):
 
         if not allowed:
             return redirect(self.redirect_url)
+
+
+# Author: Tony Abou-Assaleh
+# Source: http://stackoverflow.com/a/7871831
```
kdeloach: I was just going to redirect to the logout page at first, but got the idea to return the `contrib.auth` logout view from this person on SO.

kdeloach:
I will add support for `is_ambassador` in a separate PR.

RickMohr:
+1, tested. Great solution.

PR #1648, opened 2015-05-13
Fix Nginx bug in logrotate postrotate script

hectcastro:
See also:

  - https://github.com/azavea/ansible-nginx/pull/4
  - https://bugs.launchpad.net/nginx/+bug/1450770

### DIFF
--- a/deployment/ansible/roles.txt
+++ b/deployment/ansible/roles.txt
@@ -6,7 +6,7 @@ azavea.postgresql,0.2.1
 azavea.postgresql-support,0.2.0
 azavea.postgis,0.2.0
 azavea.redis,0.1.0
-azavea.nginx,0.2.0
+azavea.nginx,0.2.1
 azavea.mapnik,0.1.0
 azavea.logstash,0.1.0
 azavea.relp,0.1.1


jwalgran:
I repeated the steps in azavea/ansible-nginx#4 and I did see records appended to the original log after the rotate.

I did, however also see an error message:

```
root@app:/var/log/nginx# logrotate /etc/logrotate.d/nginx
error: error opening /var/log/nginx/nyc-trees-app.access.log.1.gz: No such file or directory
root@app:/var/log/nginx# ls
access.log  error.log  nyc-trees-app.access.log  nyc-trees-app.access.log.1
```

I am not sure if this is innocuous, and part of the manual test, or a symptom of a problem.

:+1:

hectcastro:
I think this error is expected behavior for the `ogrotate` script. These changes are trying to prevent the running Nginx process from holding on to a reference of the moved `nyc-trees-app.access.log.1` vs. the new`nyc-trees-app.access.log`. It should write new messages to `nyc-trees-app-access.log` after the rotate if things are working correctly.

PR #1649, opened 2015-05-13
Show ambassador status on event admin check-in page

kdeloach:
This also sorts the users on the event admin check-in page so that
ambassadors are at the top of the list.

Connects #1530


kdeloach:
Before:
image

After:
image

jwalgran:
+1 works as advertised

Issue #115, opened 2014-12-04
Event emails need an "Unsubscribe" link


RickMohr:
Per Jeff, emails to users about events should have an "Unsubscribe to emails about this event" link. If the user clicks it they should be opted out of emails for that event and taken to their profile settings page "Emails" tab, which should display the "you have unsubscribed" message shown in the wireframes.

kdeloach:
Seems like a nice-to-have. Won't have lots of email for an event, and you can unsubscribe on the profile page.

Simple solution -- link to profile page section.

hectcastro:
If Amazon starts logging complaints in SES, then the need for this becomes more pressing.

kdeloach:
@dboyer Wonder if I could get your opinion on this issue.

> If the user clicks it they should be opted out of emails for that event and taken to their profile settings page...

We can easily create a link that unsubscribes a user from event emails, however, it will require more effort to automatically login a user and redirect them to the user settings page. There is also a security risk involved if we create a URL that can automatically login as anybody.

The alternative I'd like to propose is that when a user clicks the unsubscribe link, we display a static page that says something like "You have been unsubscribed from event XYZ...". What do you think?

jwalgran:
I was thinking we could deal with just a static link to /user/settings/#emails-pane

kdeloach:
That would certainly be the easiest solution. I already have this feature like 99% working if that makes any difference (https://github.com/kdeloach/nyc-trees/commit/f46bb93e676e8a10e6c712f7368ff83b082ecb33).

dboyer:
Getting to this late, but I don't feel strongly about either approach. A general unsubscribe with no link to profile settings is fine as is the route Justin proposed.

hectcastro:
Confirmed on release stack.

PR #1654, opened 2015-05-14
Add unsubscribe link to event emails

kdeloach:
Connects #115
Connects #1660

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -6,6 +6,7 @@ from __future__ import division
 from django.contrib.auth.models import AbstractUser, UserManager
 from django.contrib.gis.db import models
 from django.core.exceptions import ValidationError
+from django.core.signing import TimestampSigner, BadSignature, SignatureExpired
 from django.core.urlresolvers import reverse
 from django.db.models import Q
 from django.utils.text import slugify
@@ -30,6 +31,7 @@ REFERRER_AD = (
 
 
 class NycUserManager(UserManager):
+
     def get_by_natural_key(self, username):
         # For login, make username case-insensitive
         return self.get(username__iexact=username)
@@ -226,6 +228,17 @@ class User(NycModel, AbstractUser):
     @property
     def reservations_map_pdf_url(self):
         return url_if_cooked(self.reservations_map_pdf_filename)
+
+    def make_token(self):
+        return TimestampSigner().sign(self.username)
+
+    # Source: http://grokcode.com/819/one-click-unsubscribes-for-django-apps/
+    def is_valid_token(self, token):
+        try:
+            TimestampSigner().unsign(token)
+        except (BadSignature, SignatureExpired):
+            return False
+        return True
 
     class Meta:
         ordering = ['username']
--- a/src/nyc_trees/apps/event/templates/event/email_rsvps.html
+++ b/src/nyc_trees/apps/event/templates/event/email_rsvps.html
@@ -33,12 +33,12 @@
             </div>
             <div class="modal-body species-modal-content">
                 <ul class="list-unstyled">
-                {% for event_reg in event_registrations %}
+                {% for rsvp in event_registrations %}
                     <li>
-                    {% if user.profile_is_public %}
-                    <a href="{{ user.get_absolute_url}}">{{ user.username }}</a>
+                    {% if rsvp.user.profile_is_public %}
+                    <a href="{{ rsvp.user.get_absolute_url}}">{{ rsvp.user.username }}</a>
                     {% else %}
-                    {{ user.username }}
+                    {{ rsvp.user.username }}
                     {% endif %}
                     </li>
                 {% endfor %}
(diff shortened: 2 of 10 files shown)


kdeloach:
This is waiting for feedback from #115.

kdeloach:
@jwalgran @hectcastro This has been updated to display a friendly error message if someone tries to use an invalid or expired token. Also, event emails are always sent from the default configured sender instead of the group contact.

src/nyc_trees/apps/event/templates/event/email_rsvps.html
```diff
@@ -33,12 +33,12 @@ <h4 class="modal-title">Event Registrants</h4>
             </div>
```
hectcastro: Potentially unrelated to this PR, but I'm getting an `XMLHttpRequest` cross-origin request error on the "See all" link of this page.

src/nyc_trees/apps/core/models.py
```diff
@@ -227,6 +229,18 @@ def blocks_mapped_count(self):
...
         return url_if_cooked(self.reservations_map_pdf_filename)
 
+    def make_token(self):
+        return TimestampSigner().sign(self.username)
+
+    # Source: http://grokcode.com/819/one-click-unsubscribes-for-django-apps/
+    def is_valid_token(self, token):
+        try:
+            # Valid for 2 days
+            TimestampSigner().unsign(token, max_age=60 * 60 * 48)
```
hectcastro: After thinking about this over lunch, I couldn't think of a good reason to bound the token with a TTL. Not something that needs to be addressed here though.

kdeloach:
I've added another commit to this PR for issue #1660.

src/nyc_trees/apps/mail/libs.py
```diff
@@ -36,10 +36,22 @@ def send_to(user, message_type, *args, **kwargs):
...
     to = user.email
 
-    msg = EmailMessage(subject, body_text, from_email, [to])
+    headers = {}
+    if reply_to:
+        headers.update({
+            'Reply-To': reply_to
+        })
+
+    msg = EmailMessage(subject, body_text, from_email, [to], headers=headers)
```
hectcastro: This is missing a list for BCC addresses after `[to]`.

src/nyc_trees/apps/event/templates/event/email_rsvps.html
kdeloach: Interesting, I thought it just opens a static Bootstrap modal.

src/nyc_trees/apps/core/models.py
kdeloach: True, this may be overkill.

src/nyc_trees/apps/mail/libs.py
kdeloach: It's optional according to the API docs.

src/nyc_trees/apps/event/templates/event/email_rsvps.html
hectcastro: It opens the model, but also throws that error in the console.

kdeloach: I see, that's very strange.

src/nyc_trees/apps/core/models.py
kdeloach: Removed.

src/nyc_trees/apps/event/views.py
```diff
@@ -429,3 +445,26 @@ def increase_rsvp_limit(request, event_slug):
     return {
         'max_attendees': event.max_attendees
     }
+
+
+@transaction.atomic
+def event_email_unsubscribe(request, event_slug, token):
```
hectcastro: I think there needs to be a test in here somewhere to confirm that the username coming back with the token is the user from the current session. If I click a valid unsubscribe link of another user while logged in as another, it unsubscribes them.

src/nyc_trees/apps/mail/libs.py
hectcastro: Nevermind. Apparently this works without the BCC argument.

hectcastro:
Unsubscribe flow works other than the hitch I ran into by accidentally clicking the unsubscribe link of an RSVP as the event organizer, which erroneously unsubscribed the person who RSVP'd.

`Reply-To` header is correctly being applied to messages, and the `From` block includes the event contact name. This was also tested manually in EC2 to ensure that SES is not using its whitelist to filter outbound messages based on `Reply-To`.

src/nyc_trees/apps/event/views.py
kdeloach: Valid concern, but this is by design. If we check against the current user session, that assumes the user must login first, which defeats the purpose of having a one-click link to unsubscribe. It's also pretty convenient during testing. Hope that makes sense.

kdeloach:
Thanks for testing. Since event organizers have no way of obtaining user tokens in production, I think that case will not be a cause for concern.

kdeloach:
@RickMohr Could use another pair of eyes on this one.

kdeloach:
Reopened as #1674 to target release branch.

Issue #1666, opened 2015-05-15
404 when requesting trusted mapper status


dboyer:
1. While not logged in, I went to the group page for A Tree Grows in Brooklyn and clicked the request independent mapper status button. 
2. I was prompted to login and did so with my dboyer account which has census admin privileges.
3. group/tree-grows-brooklyn/request-trusted-mapper-status/ loaded but displayed as a 404.

dboyer:
Maybe related. 

1.I went to the homepage then back to groups page then a tree grows in Brooklyn page. 
2. Clicked request independent mapper status again
3. Was taken to the reservations instructions page where all requirements were checked as completed.

RickMohr:
@dboyer, your second issue (landing on the reservations instructions page) is what we do if you are the group admin or are already a trusted mapper for the group in question. Is one of those true in this case?

RickMohr:
For these (second issue) cases we could instead show a page saying something like "You already have independent mapper status for this group."

In addition, if you are logged in and have independent mapping rights for the group, the group page could show a message like "You have independent mapping privileges in this group's Census Zone." rather than showing the independent mapping information and button.

kdeloach:
The endpoint for requesting mapper status only accepts POST, not GET requests, which results in a 404. We should probably redirect back to the group detail page after login.

RickMohr:
Created #1668 for the second issue.

kdeloach:
For the "reservations instructions page" issue:

If the user has independent mapper status already, clicking this button should take them to the Reservations page. If you don't have all the requirements necessary to make reservations you get forwarded to the Reservations Instructions page. We should either disable this button or figure out why qualified trusted mappers are being forwarded to the Reservations Instructions page.

PR #1667, opened 2015-05-15
Allow requesting group independent mapper status when logged out

RickMohr:
If you click 'request independent mapper status" for a group when logged out, after logging in you will be redirected to the group detail page where you can click the button again.

Connects #1666


kdeloach:
+1 Tested

Issue #1659, opened 2015-05-14
Out of date screenshot of Training materials


maurizi:
The text and screenshot should probably be updated to reflect that you must choose "No more trees on this block" in order to input the distance to the end of the block.

Current screenshot:
![](https://treescount.azavea.com/static/img/training/Distance_To_End.png)

maurizi:
This is on: page 1 of "Wrapping up", page 5 of "The mapping Method"

dboyer:
Use the attached instead with the filename Distance_To_End.png so that it matches the current path.

![screenshot_2015-05-14-22-39-02](https://cloud.githubusercontent.com/assets/516764/7646556/13deebc6-fa93-11e4-9b9c-1f8366d2dd85.png)

dboyer:
Also, I'll talk to Parks and update the text myself if necessary.

jwalgran:
It is important that the case of the file name and extension match exactly.
The last tile I replaced one of these images a change in extension case led
to problems.

hectcastro:
Confirmed on release stack.

PR #1669, opened 2015-05-15
Update stale training image for distance to end

hectcastro:
To test, ensure that you rebuild your application server static assets.

Connects #1659

---

treescount__2015
treescount__2015 2


kdeloach:
+1 Confirmed

image

Issue #906, opened 2015-03-18
Add monitoring/alarm to make sure reservation reminder job is running and mails are sent in a timely fashion - Use existing SNS topic


jwalgran:

kdeloach:
After discussing with Hector, we decided to implement this as a custom watchman health check. Pingdom will send us an email alert if the health check fails.

PR #1671, opened 2015-05-15
Add watchman health check for send_reminder_emails cron job

kdeloach:
This adds a new section to the watchman health check to report the
status of cron jobs.

For the send_reminder_emails cron job we report OK if the job has run
at least once in the past 48 hours.

Connects #906

### DIFF
--- a/src/nyc_trees/apps/survey/checks.py
+++ b/src/nyc_trees/apps/survey/checks.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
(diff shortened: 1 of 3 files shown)


kdeloach:
The reason for using 48 hours compared to 24 hours is because the send_reminder_emails last run time is saved as a Date instead of a DateTime. Since we don't know which hour the task was previously executed there's no way to get an hours delta. So this health check will only fail if the days delta is greater than 1.

hectcastro:
I tested cases where the `TaskRun` table was empty, the last `send_reminder_emails` task was run 2 days out, 1 day out, and today. All test cases produced expected results.

:+1:

PR #1673, opened 2015-05-15
Fix species modal

### LINKED ISSUE #1661: Species modal is larger than the screen on iPhone 4
img_0002

jfrankl:
Some of the modal styles were being overridden. Removing these overrides and removing the overflow property from the html element fixes the issue. Fixes #1661.


jwalgran:
Please reopen this pull against the ``release/6.0.150514`` branch

kdeloach:
+1 Confirmed that the species modal no longer overflows on both mobile and desktop.

image

Issue #1660, opened 2015-05-14
Make all event bulk event email send from a verified address by set the Reply-To header to the event contact


jwalgran:
Amazon SES will only allow sending from verified addresses. Currently the "email attendees" page fails to send emails because it is attempting to set the "from:" address to the group admin. ~~We can fix this by either:~~

1. ~~Sending all "email attendees" messages from treescount.help@parks.nyc.gov~~
1. ~~Using the AWS console to verify the addresses of all event contact. This will require personal coordination with the owners of the email accounts and will require ongoing maintenance to add new contacts.~~

We can use the Reply-To header get around this problem.

```python

if event.contact_name:
    from_email = '%s <%s>' % (event.contact_name, settings.DEFAULT_FROM_EMAIL)
else:
    from_email = settings.DEFAULT_FROM_EMAIL

msg = EmailMessage("Test Subject", "Test Body", from_email, ["recipient@domain.com"], [],
                   headers={'Reply-To': 'event-contact@foo.bar'})

dboyer:
The second option doesn't seem realistic. We could make the first work if the help text above the email form mentioned that group admins should include their email address in the email text so that event registrants can contact them. I'll run this by Parks.

jwalgran:
I emailed Parks and proposed @hectcastro 's suggestion of using the Reply-To field, so that all mail will come from the single, verified address, but the Reply-To header will be set to the event contact email.

jwalgran:
Parks agreed to the Reply-To option

hectcastro:
Confirmed on release stack.

PR #1674, opened 2015-05-15
Add unsubscribe link to event emails

kdeloach:
Connects #115
Connects #1660

### DIFF
--- a/src/nyc_trees/apps/event/views.py
+++ b/src/nyc_trees/apps/event/views.py
@@ -128,13 +128,13 @@ def event_detail(request, event_slug):
 def event_email_page(request, event_slug):
     event = get_object_or_404(Event, slug=event_slug,
                               group=request.group)
-    event_registrations = event.eventregistration_set.all() \
-        .select_related('user')
+    rsvps = EventRegistration.objects.filter(event=event, opt_in_emails=True) \
+                                     .select_related('user')
     return {
         'event': event,
         'group': event.group,
-        'event_registrations': event_registrations,
-        'rsvp_count': len(event_registrations),
+        'event_registrations': rsvps,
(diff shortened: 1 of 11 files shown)


src/nyc_trees/apps/event/views.py
```diff
@@ -429,3 +445,26 @@ def increase_rsvp_limit(request, event_slug):
...
+
+    if not user.is_valid_token(token):
+        # Invalid token
+        return {}
+
+    EventRegistration.objects.filter(event=event, user=user) \
+                             .update(opt_in_emails=False)
+
+    # Redirect to Emails tab on user profile page.
+    return redirect(reverse('user_profile_settings') + '#emails-pane')
```
hectcastro: If I'm not logged in, after clicking the unsubscribe link I get sent to the app and it successfully unsubscribes me, but I don't get any notification -- I just get sent to the login page. After I login, I get sent to `/user/settings`, but not the Emails tab.

This makes it look like I need to login to unsubscribe (even though I don't). I think that maybe we need a different redirect for when the user is not logged in.

kdeloach: I noticed this too. It's actually a bug in the login page `next_url` thing. It's not serializing the hash correctly. I'll take another look at it.

kdeloach: Decided to fix this by displaying a static page after unsubscribing to an event instead of redirecting to the user settings page.

kdeloach: Updated.

kdeloach:
Added a commit to display a friendly message after unsubscribing from an event email instead of redirecting to the login page.

kdeloach:
Preview:
image

Issue #621, opened 2015-02-17
E-mail quota check failure when enabled


hectcastro:
When `EMAIL_BOTO_CHECK_QUOTA` is enabled, the following exception is thrown during the e-mail sending process:

```
Internal Server Error: /accounts/register/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/registration/views.py", line 79, in dispatch
    return super(RegistrationView, self).dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 87, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/registration/views.py", line 35, in post
    return self.form_valid(request, form)
  File "/usr/local/lib/python2.7/dist-packages/registration/views.py", line 82, in form_valid
    new_user = self.register(request, **form.cleaned_data)
  File "/opt/app/apps/login/backends.py", line 14, in register
    .register(request, **cleaned_data)
  File "/usr/local/lib/python2.7/dist-packages/registration/backends/default/views.py", line 91, in register
    request=request,
  File "/usr/local/lib/python2.7/dist-packages/registration/models.py", line 94, in create_inactive_user
    registration_profile.send_activation_email(site, request)
  File "/usr/local/lib/python2.7/dist-packages/registration/models.py", line 303, in send_activation_email
    email_message.send()
  File "/usr/local/lib/python2.7/dist-packages/django/core/mail/message.py", line 286, in send
    return self.get_connection(fail_silently).send_messages([self])
  File "/opt/app/apps/core/mail/backends/boto.py", line 35, in send_messages
    remaining_quota = self.mailer.get_remaining_message_quota()
  File "/opt/app/libs/boto_mailer.py", line 29, in get_remaining_message_quota
    limit = int(float(quota('Max24HourSend')))
TypeError: 'Element' object is not callable
```

Using the AWS CLI on the same box successfully returns quota information:

```
ubuntu@ip-10-0-3-53:~$ AWS_DEFAULT_REGION=us-east-1 aws ses get-send-quota
{
[...]

PR #1675, opened 2015-05-15
Fix failure when email quota checking is enabled

kdeloach:
This fixes an exception that is raised when EMAIL_BOTO_CHECK_QUOTA is
set to True.

Connects #621


hectcastro:
Tested on the release stack and it works:

image

I think we should consider enabling `EMAIL_BOTO_CHECK_QUOTA` again, but maybe isolate this call from the e-mail sending process so that sending e-mails doesn't fail if getting the quota or parsing the response fails.

Issue #1676, opened 2015-05-15
[Part of 1592] Group admin can change event capacity to less than the number of people registered


dboyer:
1. I set up an event with a capacity of two.
2. I RSVPed two people to the event.
3. I logged in as the group admin and changed the event capacity to one.
4. The event then had 2 of 1 people registered - /group/american-hornbeam-appreciation-society/event/a-new-event-may-15/

We should not allow the group admin to change the event capacity to fewer than the number of people already registered.

dboyer:
This has become an issue on production - https://treescount.nycgovparks.org/group/prospect-park-alliance/event/GRkZLRtmbJefzJqfBozuVa/

jwalgran:
I mistakenly fixed this before noticing that it was in the milestone for next sprint.

PR #1677, opened 2015-05-16
Prevent decreasing max attendance below RSVP count

jwalgran:
If 12 people RSVP to an event, it should not be possible to decrease the value of max_attendees to 10.

Connects to #1676


hectcastro:
Works. :+1:

maurizi:
@jwalgran The fix for this still hasn't made it to production right?  It might be worth making a small data migration as well, to up the event capacity to be at least equal to the number of registered attendees.

jwalgran:
I want to avoid spending the time writing, testing, and reviewing a migration for localized issue easily correctible through the UI.

Issue #1678, opened 2015-05-16
Clarify event email message when there are not users opted in to emails


jwalgran:
If a single person RSVPs for an event, but unsubscribes from event emails, and the group admin attempt to send a email to all RSVPed users, they are shown this message

"There are no voluntreers RSVPed for your event. Please try again when there is at least 1 voluntreer RSVPed."

This is confusing because there is, in fact, one user RSVPed.

jwalgran:
A PR implementing this change was merged

PR #1680, opened 2015-05-16
Clarify event e-mail message when no RSVPs have opted in

hectcastro:
An event may have at least 1 RSVP, but those RSVPs may not have opted in to event e-mail notifications.

Connects #1678


kdeloach:
+1

Issue #1665, opened 2015-05-14
Erratic scroll behavior on iOS Safari when the cursor is in treecorder text box


jwalgran:
In the treecorder on Mobile Safari, scrolling while focused on a text field results in the cursor "floating" outside the text box.

http://cl.ly/2X2g0v2d3X0H/erratic-ios-scrolling-on-treecorder.mov

This does not happen all forms.

http://cl.ly/213c231T3x2s/ios-scrolling-on-event-form.mov

jfrankl:
@jwalgran Could you provide iOS and iPhone version information? I can't reproduce this on the office iPhone 4 + iOS7.

jwalgran:
It happens on my real 4s with the latest iOS (8.3 I think) as well as the
simulator 4s and 5s

jfrankl:
Working on the fix with @RickMohr. Here is the current status: 87d34b6914648e97adc5de741c3046a6b4726430

I was able to fix the glitchy scrolling with keyboard open, but it broke the fixed "Tree 1" elements. Rick is working on getting that to work again.

RickMohr:
Should be fixed by https://github.com/RickMohr/nyc-trees/tree/fix-sticky-titles

jfrankl:
Fixed by #1685.

PR #1685, opened 2015-05-18
Fix scrolling glitch on Treecorder for iOS7

RickMohr:
Connects #1665


hectcastro:
@azavea-bot rebuild

jwalgran:
The title of the pull says iOS 7, but this was an issue on iOS 8.3 as well.

+1 I confirmed that the cursor stays in the field while scrolling on iOS 8.3 on a hardware iPhone 4s, simulator iPhone 8, Chrome on OS X, and Chrome on Android Kit Kat.

Issue #1684, opened 2015-05-16
Add a management command to install the ESRI:102718 NY state plane projection


jwalgran:
The state plane projection which gives us the most accurate geometry construction is not included with PostGIS by default. We will need to add a command to our [setupdb.sh](https://github.com/azavea/nyc-trees/blob/develop/scripts/aws/setupdb.sh) script to ensure that we have a row for SRID 102718 in ``spatial_ref_sys``. The query we need to run is already [documented](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/survey/fixtures/README.md#source-data).

jwalgran:
This may not be required. Looking into it.

hectcastro:
Before attempting to update `spatial_ref_sys`, we need to set the owner to `rds_superuser`:

```sql
ALTER TABLE public.spatial_ref_sys OWNER TO rds_superuser;
```

PR #1682, opened 2015-05-16
Use state plane projection for trees layout, for improved accuracy

strk:
The state plane projection more closely matches real-world measurements from surveyors.
Here are the results comparing utm18n (coming out of _st_bestsrid) and state plane:
```
 tree_id | survey_id | block_id | dist_requested | dist_utm18n | dist_stateplane 
---------+-----------+----------+----------------+-------------+-----------------
     528 |       165 |   343906 |         172.50 |      192.95 |          172.52
     529 |       165 |   343906 |          35.30 |       39.48 |           35.30
     530 |       165 |   343906 |          32.10 |       35.90 |           32.10
     531 |       165 |   343906 |          31.50 |       35.23 |           31.50
     532 |       165 |   343906 |          31.90 |       35.68 |           31.90
     533 |       165 |   343906 |          29.50 |       32.99 |           29.50
     534 |       165 |   343906 |          17.00 |       19.01 |           17.00
     535 |       165 |   343906 |          25.00 |       27.96 |           25.00
     536 |       165 |   343906 |          58.20 |       65.09 |           58.20
     537 |       165 |   343906 |          15.30 |       17.11 |           15.30
     538 |       165 |   343906 |          22.10 |       24.72 |           22.10
     539 |       165 |   343906 |          90.50 |      101.22 |           90.50
     540 |       165 |   343906 |          36.40 |       40.71 |           36.40
     541 |       165 |   343906 |          77.40 |       86.57 |           77.40
     542 |       165 |   343906 |          87.70 |       98.09 |           87.70
     543 |       165 |   343906 |          30.70 |       34.34 |           30.70
     544 |       165 |   343906 |         116.50 |       34.19 |          116.50
(17 rows)
```

PR #1686, opened 2015-05-18
Add state plane projection install to setupdb.sh

jwalgran:
This projection is required for more accurate geometry construction.

Connects to #1684


RickMohr:
+1 see comment on #1682

Issue #1655, opened 2015-05-14
RSVP emails don't work if event PDF is missing


kdeloach:
This happens frequently on my local dev instance and I just saw this in staging so I think the issue needs to be addressed.

If the event PDF fails to be generated (or was deleted) this blocks all event registration emails from sending. We should generate this PDF if it doesn't exist.

![image](https://cloud.githubusercontent.com/assets/43062/7639764/575d4146-fa4c-11e4-95f2-b766cc0bc307.png)

kdeloach:
The celery task that sends the RSVP emails eventually gives up when the PDF is missing.

Relevant stack trace:
```
[2015-05-14 15:14:24,346: ERROR/MainProcess] Task apps.core.tasks.wait_for_default_storage_file[a5a71f86-8a77-4c2d-a2d2-2150b384f8da] raised unexpected: MaxRetriesExceededError("Can't retry apps.core.tasks.wait_for_default_storage_file[a5a71f86-8a77-4c2d-a2d2-2150b384f8da] args:(u'event_map/this-is-required_dEGL8WkzzV7pdccHJt6VQa.pdf',) kwargs:{}",)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/core/tasks.py", line 17, in wait_for_default_storage_file
    self.retry()
  File "/usr/local/lib/python2.7/dist-packages/celery/app/task.py", line 665, in retry
    self.name, request.id, S.args, S.kwargs))
MaxRetriesExceededError: Can't retry apps.core.tasks.wait_for_default_storage_file[a5a71f86-8a77-4c2d-a2d2-2150b384f8da] args:(u'event_map/this-is-required_dEGL8WkzzV7pdccHJt6VQa.pdf',) kwargs:{}```

kdeloach:
The workaround for this is to edit the event location to rebuild the PDF.

jwalgran:
Another workaround: delete the filename from the Django admin and then open the event detail page.

kdeloach:
I couldn't figure out an adequate solution to this problem. I fixed the bug where registration emails wouldn't send if the PDF is missing, but that's it. Closing until this happens in production.

PR #1687, opened 2015-05-18
Send RSVP email even if PDF doesn't exist

kdeloach:
If the PDF can't be found on the disk after the maximum number of
retries, return None, instead of raising an exception. This ensures that
the RSVP email notification still sends even without the PDF attachment.

How to test:
1. Change an event's map pdf filename to a non-existent file (or delete the pdf inside your VM).
2. RSVP for that event.
3. In celery output you should see it try to find the PDF a few times, then give up, and send the RSVP email anyway.

Refs #1655


RickMohr:
+1. I deleted an event PDF, RSVPd to the event, saw celery try 15 times, and then saw the email sent without the PDF.

hectcastro:
Nice addition.

Issue #1679, opened 2015-05-16
Cancel button on event edit details page takes you to group edit page


hectcastro:
1. Attempt to edit an event
2. Click cancel
3. Redirected to group edit page

kdeloach:
We use this form for both "adding a new event" and "editing an existing event". The cancel link behavior makes sense for the former but not the latter.

PR #1688, opened 2015-05-18
Add context sensitive cancel link

kdeloach:
When adding an event, the cancel link should return to the group edit
page. When editing an existing event, the cancel link should go to the
event detail page.

Connects #1679


maurizi:
+1, works great

Issue #1668, opened 2015-05-15
Group detail: "Request Independent Mapper Status" button shown even if you already have it


RickMohr:
If you have independent mapping rights for a group, the group page could show a message like "You have independent mapping privileges in this group's Census Zone" rather than showing the independent mapping information and "request" button.

PR #1689, opened 2015-05-18
Disable trusted mapper button if user has been approved

kdeloach:
This disables the trusted mapper button on the group detail page if the
user has been approved as a trusted mapper. Some help text is also
displayed.

Connects #1668

### DIFF
--- a/src/nyc_trees/apps/core/helpers.py
+++ b/src/nyc_trees/apps/core/helpers.py
@@ -29,9 +29,15 @@ def user_is_individual_mapper(user):
 
 def user_is_trusted_mapper(user, group):
     return user.is_authenticated() and \
-        TrustedMapper.objects.filter(group=group,
-                                     user=user,
-                                     is_approved=True).exists()
+        (user_is_group_admin(user, group) or
+         TrustedMapper.objects.filter(group=group,
+                                      user=user,
+                                      is_approved=True).exists())
+
+
+def user_mapper_status_pending(user, group):
+    return user.is_authenticated() and \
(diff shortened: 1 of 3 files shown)


kdeloach:
@jfrankl Help with syling/text? Here's what it looks like now:

image

jfrankl:
After independent mapper status has been granted, let's remove the "Request Independent Mapper Status" button rather than deactivate it.

RickMohr:
I agree it makes more sense to hide the button. Probably we should adjust the text? Something like:

#### Independent Mapping
You are an approved independent mapper for this group, so you can reserve and map blocks in its  Census Zone. Remember to contact your local loaning hub to check out individual tools.

@dboyer, what do you think?

RickMohr:
Also I think we should show still different text if you have requested independent mapper status for a group but your request has not been acted on.

kdeloach:
Good idea, will update.

kdeloach:
Added different text when your trusted mapper status request is pending.

kdeloach:
Squashed fixups. Let me know how this looks.

RickMohr:
It looks good to me. @dboyer, any adjustments to the wording for these three cases?

If user has not requested independent mapping status for a group (no change from current):

capture

If user's request has not been acted on:

capture

If user has been approved (or if they are the group admin):

capture

jwalgran:
Do we cover the case where trusted mapper status has been rejected?

kdeloach:
If it was rejected we show the "pending" message.

dboyer:
That text sounds great. Thanks for writing it.

RickMohr:
+1

Issue #1681, opened 2015-05-16
Create management command to bust tiler cache


hectcastro:
The management command should encapsulate something like the following:

```python
from apps.survey.models import *
b = Blockface.objects.first()
from django.utils.timezone import now
b.updated_at = now()
b.save()
```

PR #1690, opened 2015-05-18
Create management command to bust tiler cache

kdeloach:
Connects #1681


hectcastro:
Before running the command, the tile path prefix was `1430954698`. After running the command, it changed to `1431969592`.

@jwalgran Do we also want to add this to `setupdb.sh`? It is unfortunate that it invalidates the entire cache, but that may be better than accidentally having CloudFront cache empty tiles.

jwalgran:
We did a lot of work to make sure that a lack of cache would not over stress the tilers, so I do think we should add this as the last line of setupdb.sh, with a comment along the lines of "this must be run after all data changes"

RickMohr:
So deploying to production will have the same effect on the CloudFront cached tiles as mapping a block. Doesn't seem like a problem.

kdeloach:
Good suggestion. Updated `setupdb.sh` to run the `touch_blockface` management command.

hectcastro:
:+1:

PR #1706, opened 2015-05-21
Add AutoScalingPolicy to application and tile servers

hectcastro:
There are several high traffic periods in the day when mapping events occur. Typically, these happen in bursts and then taper off after an hour or two. In order to protect ourselves during these surges of traffic, scale up and scale down policies have been added to the application and tile servers.

Looking at the ELB graphs for overall request count, a threshold of 60 with a period of 60 seconds and 2 consecutive evaluation periods is a good combination to trigger a scale up event. Before the following minute is over, server capacity doubles.

On the other hand, a threshold of 75 with a period of 60 seconds and 10 consecutive evaluation periods is a safe combination to trigger a scale down event. This is more conservative and protects against lulls between evaluation periods where activity drops, but then picks back up in the next minute or two.

Regardless, the desired and minimum capacity for each AutoScalingGroup remains at 2.

### DIFF
--- a/deployment/troposphere/app_template.py
+++ b/deployment/troposphere/app_template.py
@@ -252,6 +252,7 @@ t.add_resource(cw.Alarm(
     'alarmAppServerRequestCountLow',
     AlarmDescription='Application server request count low',
     AlarmActions=[Ref(app_server_scale_down_policy)],
+    InsufficientDataActions=[Ref(app_server_scale_down_policy)],
     Statistic='Sum',
     Period=60,
     Threshold='75',
--- a/deployment/troposphere/app_template.py
+++ b/deployment/troposphere/app_template.py
@@ -169,6 +169,105 @@ app_server_load_balancer = t.add_resource(elb.LoadBalancer(
     )
 ))
 
+#
+# Auto Scaling Group Resources
+#
+app_server_launch_config = t.add_resource(asg.LaunchConfiguration(
+    'lcAppServer',
+    ImageId=Ref(app_server_ami_param),
+    IamInstanceProfile=Ref(app_server_instance_profile_param),
+    InstanceType=Ref(app_server_instance_type_param),
+    KeyName=Ref(keyname_param),
+    SecurityGroups=[Ref(app_server_security_group)]
+))
+
+app_server_auto_scaling_group = t.add_resource(asg.AutoScalingGroup(
+    'asgAppServer',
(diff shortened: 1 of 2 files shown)


hectcastro:
I think this is ready for review. Mostly to help protect against issues with events that have a lot of attendees or events that overlap.

/cc @kdeloach @jwalgran

deployment/troposphere/tiler_template.py
```diff
@@ -158,6 +158,106 @@
...
+            'metricLoadBalancerName',
+            Name='LoadBalancerName',
+            Value=Ref(tile_server_load_balancer)
+        )
+    ],
+))
+
+t.add_resource(cw.Alarm(
+    'alarmTileServerRequestCountLow',
+    AlarmDescription='Application server request count low',
```
kdeloach: Alarm description should be "Tile server".

deployment/troposphere/app_template.py
```diff
@@ -169,6 +169,106 @@
...
+            Value=Ref(app_server_load_balancer)
+        )
+    ],
+))
+
+t.add_resource(cw.Alarm(
+    'alarmAppServerRequestCountLow',
+    AlarmDescription='Application server request count low',
+    AlarmActions=[Ref(app_server_scale_down_policy)],
+    InsufficientDataActions=[Ref(app_server_scale_down_policy)],
```
kdeloach: How come the scale down policy alarms need `InsufficientDataActions` and the scale up policy alarms don't?

kdeloach:
+1 Looks good but I'm not able to test this.

deployment/troposphere/app_template.py
hectcastro: If no requests come in, and we're in a scaled up state (spending more money), the scale down event does not trigger. This protects us from the scenario where a scale up event occurs followed by long periods of no traffic (no need to spend more money).

If there are `InsufficientDataActions` for `RequestCount`, then there aren't any user requests coming through the ELB. No need to scale up in that scenario.

deployment/troposphere/tiler_template.py
hectcastro: Fixed, thanks.

deployment/troposphere/app_template.py
kdeloach: Thanks for the explanation.

Issue #1692, opened 2015-05-18
Allow the group admin to survey an event without needing to consume an RSVP slot and check in


jwalgran:
A request from Parks.

    "The use case that has come up is that a Partner Group's group admin may need
    to do data entry for any paper data collection volunteers for them, prior to
    the event's midnight deadline."

PR #1709, opened 2015-05-21
Allow group admins to survey their own events without needing an RSVP

kdeloach:
This change allows group admins to survey their own events without
taking up an RSVP slot. Clicking the Treecorder link in the nav while
an event is in progress should take the group admin to the event
check-in page.

To test:
1. Login as a group admin with an event in progress (don't RSVP)
2. Click the Treecorder link
3. You should be taken to the "survey for event" page

Connects #1692

### DIFF
--- a/src/nyc_trees/apps/event/helpers.py
+++ b/src/nyc_trees/apps/event/helpers.py
@@ -2,17 +2,21 @@ from apps.event.models import EventRegistration
 
 
 def user_is_rsvped_for_event(user, event):
-    # The filter uses user.id so that the user argument can be either
-    # a real User or the anonymous user (not logged in). The anonymous
-    # user is a SimpleLazyObject, which you cannot use in a filter.
-    return EventRegistration.objects\
-                            .filter(user=user.id, event=event)\
-                            .count() > 0
+    if not user.is_authenticated():
+        return False
+    # Group admins are implicitly RSVPd to their own events.
+    if event.group.admin == user:
+        return True
+    return EventRegistration.objects.filter(user=user, event=event).exists()
 
 
 def user_is_checked_in_to_event(user, event):
     """Return True if user is checked-in to event"""
-    return (user.is_authenticated() and
-            EventRegistration.objects.filter(user=user,
-                                             event=event,
-                                             did_attend=True).exists())
+    if not user.is_authenticated():
+        return False
+    # Group admins are implicitly RSVPd to their own events.
+    if event.group.admin == user:
+        return True
+    return EventRegistration.objects.filter(user=user,
+                                            event=event,
+                                            did_attend=True).exists()
--- a/src/nyc_trees/apps/event/models.py
+++ b/src/nyc_trees/apps/event/models.py
@@ -222,8 +222,8 @@ class EventRegistration(NycModel, models.Model):
         # Then get those events which are starting soon or have yet to end, or
         # any attended events from today, which can be mapped until midnight
         registrations = EventRegistration.objects \
-            .filter(user=user,
-                    event__begins_at__gte=beginning_of_day_utc) \
+            .filter(user=user) \
+            .filter(event__begins_at__gte=beginning_of_day_utc) \
             .filter(Q(did_attend=True, event__ends_at__lte=end_of_day_utc) |
                     Q(event__begins_at__lte=now_utc + STARTING_SOON_WINDOW,
                       event__ends_at__gt=now_utc)) \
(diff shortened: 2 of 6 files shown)


src/nyc_trees/apps/event/helpers.py
```diff
@@ -2,17 +2,21 @@
 
 
 def user_is_rsvped_for_event(user, event):
-    # The filter uses user.id so that the user argument can be either
-    # a real User or the anonymous user (not logged in). The anonymous
-    # user is a SimpleLazyObject, which you cannot use in a filter.
-    return EventRegistration.objects\
```
maurizi: This comment and the fix it references (using `user.id` instead of just `user` fixed a bug we encountered several times).  I would like us to keep it in place.

src/nyc_trees/apps/event/models.py
```diff
@@ -231,6 +231,16 @@ def my_events_now(cls, user):
...
         attended = [r.event for r in registrations if r.did_attend]
         non_attended = [r.event for r in registrations if not r.did_attend]
+
+        # Group admins don't need to RSVP to be able to survey events.
+        # These filters are copy & pasted from the query directly above.
+        admin_events = Event.objects \
+            .filter(group__admin=user, begins_at__gte=beginning_of_day_utc) \
+            .filter(Q(ends_at__lte=end_of_day_utc) |
+                    Q(begins_at__lte=now_utc + STARTING_SOON_WINDOW,
+                      ends_at__gt=now_utc))
```
maurizi: No need to do this now if you don't want to, but FYI a nice way to avoid this duplication is to do something like this:

```python
events_soon = Event.objects \
    .filter(begins_at__gte=beginning_of_day_utc) \
    .filter(Q(ends_at__lte=end_of_day_utc) |
            Q(begins_at__lte=now_utc + STARTING_SOON_WINDOW,
              ends_at__gt=now_utc))

registrations = EventRegistration.objects \
    .filter(user=user) \
    .filter(event_id__in=events_soon.values_list('id', flat=True))

attended = [r.event for r in registrations if r.did_attend]
non_attended = [r.event for r in registrations if not r.did_attend]

attended = attended + list(events_soon.filter(group__admin=user))
```

maurizi:
@azavea-bot rebuild

maurizi:
As the group admin, I can click the "Attending" button on the group detail page.  It will still say "Attending", but it makes the "Check-in guests" button appear/disappear.  We should probably just disable clicking on it if you are the group admin.

maurizi:
Events alerts for groups that you are the admin of do not appear in the header on the dashboard, and do not appear in the "Attending" section of the event list on the dashboard.

The event alert seems important since it is one of the easiest ways for group admins to be able to check users into the event.

The event not showing in the event list seems less important, but it is confusing, since if you go to the event detail, you are shown as "Attending"

src/nyc_trees/apps/event/helpers.py
kdeloach: I decided it's not worth making a query if we already know the answer.

src/nyc_trees/apps/event/models.py
kdeloach: OK I like this better, will update.

kdeloach: Assuming we can safely drop the `did_attend=True` criteria from this query.

kdeloach:
Added a commit to make the RSVP button static for group admins and also fixed event alerts not appearing on the dashboard.

kdeloach:
Added another commit to display events on the dashboard and event list.

kdeloach:
Pushed a fix to the event ordering in `next_event_starting_soon`. 
We can't actually do sorting when combining the results because the event begins_at field is no longer in context at that point (it's just a list of RSVPs). I moved the admin "RSVPs" to the beginning of the list so they take precedence over other events, and hopefully that will be good enough.

maurizi:
+1, works well.

I actually see *two* choices in the treecorder modal when I click on the treecorder link from the nav, one to "check in" which takes me to the admin check-in page, and one to "map" which takes me to the survey page, but that seems fine.

maurizi:
Though it looks like your unit test is failing

kdeloach:
You actually found an edge case I didn't think about. It only happens if a group admin RSVPs to their own event. However, group admins no longer have the ability to RSVP to their own events, so this problem should not happen in production.

Now that I think about it... maybe we should link group admins to the check-in page instead of directly to the survey page?

kdeloach:
Fixed unit test.

kdeloach:
Fixed the edge case where, if the group admin RSVPs and checks into their own events, it would display 2 links in the Treecorder urls panel. Now it should always display the check-in page for group admins.

maurizi:
But now your unit test is broken again :(

kdeloach:
Yup, I can see that.
image

kdeloach:
Fixed unit tests.

maurizi:
+1, works great

Issue #1694, opened 2015-05-18
Error requesting mapper status on Reservations Instructions


kdeloach:
![image](https://cloud.githubusercontent.com/assets/43062/7689112/dbce0248-fd77-11e4-9f3a-2a6b3b7aa3ca.png)

kdeloach:
Experienced on latest develop.

PR #1710, opened 2015-05-21
Fix "request mapper status" link on reservations instructions page

kdeloach:
This fixes the "Request independent mapper status" link (step 4) on the
reservations instructions page.

This also fixes a bug where you could bypass training by submitting a
POST to the request mapper status endpoint.

Connects #1694


maurizi:
+1, works

Issue #1711, opened 2015-05-21
[Part of 1592] Tree locations in the shapefile dump are incorrect


jwalgran:
The labels in this screenshot are ``tree_id,dist_to_end``

![image002](https://cloud.githubusercontent.com/assets/17363/7753855/f187b2e2-ff9c-11e4-89df-c89bad810dc7.jpg)

The trees are not placed along the block in ``tree_id`` order, which also means that the only tree with a ``dist_to_end`` value is not at the end of the block.

PR #1712, opened 2015-05-21
Fix sorting in tree shapefile query

jwalgran:
Sorting each aggregate by the tree id ensures all the arrays are in the same order, the order in which the trees were entered.

Test by comparing the ``trees.shp`` files generated by ``dump_db`` in a desktop GIS, before and after the fix.

Before

screen shot 2015-05-21 at 9 22 57 am

After

screen shot 2015-05-21 at 9 23 26 am

Connects to #1711

### DIFF
--- a/src/nyc_trees/apps/census_admin/trees.sql
+++ b/src/nyc_trees/apps/census_admin/trees.sql
@@ -6,11 +6,13 @@ WITH trees_for_this_survey AS(
     1.000*(CASE WHEN curb_location='OnCurb' THEN 2.5 ELSE 12 END) curb_offset,
     1.000*distance_to_tree dist
   FROM survey_tree
-  ORDER BY id
 ),
 aggregated_trees AS (
-  SELECT survey_id, array_agg(curb_offset) width,
-         array_agg(0) length, array_agg(dist) dist, array_agg(id) tree_ids
+  SELECT survey_id,
+         array_agg(curb_offset ORDER BY id) width,
+         array_agg(0) length,
+         array_agg(dist ORDER BY id) dist,
+         array_agg(id ORDER BY id) tree_ids
   FROM trees_for_this_survey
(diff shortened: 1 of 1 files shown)


kdeloach:
Testing this now.

maurizi:
+1, code looks good (did not test)

kdeloach:
I wasn't able to finish testing in time but I can review this tomorrow.

kdeloach:
The data dump from this branch compared to develop seemed to yield identical results for me. Logically, the code looks sound, but I wasn't able to reproduce the problem. Happy to re-test if you can help me reproduce the original issue.

kdeloach:
+1

src/nyc_trees/apps/census_admin/trees.sql
```diff
@@ -6,11 +6,13 @@ WITH trees_for_this_survey AS(
     1.000*(CASE WHEN curb_location='OnCurb' THEN 2.5 ELSE 12 END) curb_offset,
     1.000*distance_to_tree dist
   FROM survey_tree
-  ORDER BY id
 ),
 aggregated_trees AS (
-  SELECT survey_id, array_agg(curb_offset) width,
-         array_agg(0) length, array_agg(dist) dist, array_agg(id) tree_ids
+  SELECT survey_id,
+         array_agg(curb_offset ORDER BY id) width,
```
hectcastro: Interesting function. So this is just evaluating the expression and making it into an array?

jwalgran: Yeah. The companion function is ``unnest`` which turns an array into rows.

maurizi: Yes: http://www.postgresql.org/docs/9.3/static/functions-aggregate.html

Issue #1695, opened 2015-05-19
Prevent n.0 from raising a ValidationError when submitted as an integer field value


hectcastro:
Relevant traceback:

```
Internal Server Error: /survey/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 91, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 125, in newreq
    outp = req_function(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 106, in newreq
    outp = req_function(request, *args, **kwargs)
  File "/opt/app/apps/survey/views.py", line 537, in submit_survey
    create_result = _create_survey_and_trees(request)
  File "/usr/local/lib/python2.7/dist-packages/django/db/transaction.py", line 394, in inner
    return func(*args, **kwargs)
  File "/opt/app/apps/survey/views.py", line 614, in _create_survey_and_trees
    tree.clean_and_save()
  File "/opt/app/libs/mixins.py", line 27, in clean_and_save
    self.full_clean()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 1012, in full_clean
    raise ValidationError(errors)
ValidationError: {'circumference': [u"'11.0' value must be an integer."]}
```

jwalgran:
Options:

- Server side convert n.0 values to n before setting them on the model
- Client side round the value on and show the rounded value in the text box
- Client side round the value silently on POST

I like the server side option.

hectcastro:
I think the priority on this one needs to go up. Looking at the traffic patterns, people are generating this exception, and then trying again in rapid succession, which causes bunches of errors.

PR #1716, opened 2015-05-22
Convert any floats into integers for circumference and diameter fields.

maurizi:
Connects to #1695

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -610,6 +610,12 @@ def _create_survey_and_trees(request, event=None):
     for tree_data in tree_list:
         if 'problems' in tree_data:
             tree_data['problems'] = ','.join(tree_data['problems'])
+
+        # Convert any floats into integers for all the integer fields in Tree
+        for field in ('circumference', 'stump_diameter'):
+            if field in tree_data:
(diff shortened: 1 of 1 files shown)


kdeloach:
+1 Tested.

src/nyc_trees/apps/survey/views.py
```diff
@@ -610,6 +610,12 @@ def _create_survey_and_trees(request, event=None):
     for tree_data in tree_list:
         if 'problems' in tree_data:
             tree_data['problems'] = ','.join(tree_data['problems'])
+
+        # Convert any floats into integers for all the integer fields in Tree
+        for field in ('circumference', 'stump_diameter'):
+            if field in tree_data:
+                tree_data[field] = int(float(tree_data[field]))
```
hectcastro: Do we need to worry about rounding here?

```python
$ python
Python 2.7.9 (default, Mar 11 2015, 13:47:12)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> int(float(2.55))
2
```

maurizi: You're right, that probably *is* more correct, especially if they put in something like `12.95`

Issue #1697, opened 2015-05-19
The action bar appears and disappears on first progress page group click


jwalgran:
http://cl.ly/0U1e3m3Z3S3b ([mov file](http://cl.ly/0U1e3m3Z3S3b/first-group-click-popup-disappears.mov))

1) Go to progress page, "Groups" view
2) Click on a group polygon

The group name appears briefly in the legend area but then disappears.
Clicking on subsequent groups works fine.

PR #1718, opened 2015-05-22
Turn off zoom fixes on the progress map when reloading any layer

maurizi:
We need to put events on zoomend for the all progress view, but it
interferes with the selection on the user and group views.

We had been explicitly turning the zoomend handler off, but it was only
affecting user progress, not group progress.  Moving the `.off(...)` calls
should make it turn off for both.

Connects to #1697


jwalgran:
+1 the action bar now appears on first click and I did not see any regression in zoom-related or click-related behaviour.

PR #1719, opened 2015-05-22
Round floats to the nearest integer instead in survey submission.

maurizi:
This seems like more correct behaviour that will lead to better data
quality.

Connects to #1695

### DIFF
--- a/src/nyc_trees/apps/survey/views.py
+++ b/src/nyc_trees/apps/survey/views.py
@@ -614,7 +614,7 @@ def _create_survey_and_trees(request, event=None):
         # Convert any floats into integers for all the integer fields in Tree
         for field in ('circumference', 'stump_diameter'):
             if field in tree_data:
-                tree_data[field] = int(float(tree_data[field]))
+                tree_data[field] = int(round(float(tree_data[field])))
 
         tree = Tree(survey=survey, **tree_data)
         tree.clean_and_save()


kdeloach:
+1

src/nyc_trees/apps/survey/views.py
```diff
@@ -614,7 +614,7 @@ def _create_survey_and_trees(request, event=None):
         # Convert any floats into integers for all the integer fields in Tree
         for field in ('circumference', 'stump_diameter'):
             if field in tree_data:
-                tree_data[field] = int(float(tree_data[field]))
+                tree_data[field] = int(round(float(tree_data[field])))
```
hectcastro: I haven't been able to find any evidence that `float` adds anything here:

```python
Python 2.7.6 (default, Mar 22 2014, 22:59:56)
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> x = 2.33
>>> y = 2.93
>>> int(round(x))
2
>>> int(round(y))
3
>>> int(round(float(x)))
2
>>> int(round(float(y)))
3
```

maurizi: @hectcastro The data is coming in from the client as a string, not a float. (`"2.5"`, not `2.5`).
`float(...)` converts the string to a number.

hectcastro: Good point.

hectcastro:
:+1:

Issue #1683, opened 2015-05-16
Leaflet controls go on top of the search bar on Android


maurizi:
Tested on a Nexus 5, running Lollipop, on both Chrome and Firefox.

You can trigger this in several ways, but an easy one is:

1) zoom in on the progress map until blockfaces are visible 2) tap on a blockface 3) open the search control s
![tmp_30280-screenshot_2015-05-16-14-23-52-96919449](https://cloud.githubusercontent.com/assets/4432106/7667485/c86b6db8-fbd7-11e4-997d-270ff935420d.png)

maurizi:
Also an issue on the event list map (tested in Chrome):
![tmp_30280-screenshot_2015-05-16-14-28-41-1097617232](https://cloud.githubusercontent.com/assets/4432106/7667490/f0b42972-fbd7-11e4-954b-7e2c10e76047.png)

I have not tested the reservations or treecorder maps, but I suspect this would occur there as well.

PR #1720, opened 2015-05-22
Ensure that address search box is above leaflet controls.

maurizi:
Important for android browsers where the virtual keyboard resizes the
browser window, which can shrink the leaflet map enough to make the search
and leaflet controls conflict.

Connects to #1683


maurizi:
@jwalgran Should I reopen this against the release branch?

jfrankl:
+1

Issue #1701, opened 2015-05-20
No error validation message shown for Distance to End field


maurizi:
When entering an invalid value in "Distance to End", or leaving it blank, you should see a validation message appear after hitting submit.

This does not occur for me, on any browser.

The submission *is* prevented until a valid value is entered.

PR #1721, opened 2015-05-22
Fix field validation for the "Distance to End" field

maurizi:
The validation for "Distance to End" was broken by the recent iOS fixes
that scroll the page on validation failures.  The JS was trying to scroll
to the fieldset containing the input that failed validation, but the
Disatnce to End field is not inside a fieldset.

To fix this and prevent future regressions, I have changed it so that we
fall back to scrolling to the input itself if we cannot find a fieldset.

Connects to #1701


maurizi:
Closing so I can reopen against the release branch.

PR #1722, opened 2015-05-22
Fix field validation for the "Distance to End" field

maurizi:
The validation for "Distance to End" was broken by the recent iOS fixes
that scroll the page on validation failures.  The JS was trying to scroll
to the fieldset containing the input that failed validation, but the
Disatnce to End field is not inside a fieldset.

To fix this and prevent future regressions, I have changed it so that we
fall back to scrolling to the input itself if we cannot find a fieldset.

Connects to #1701


jwalgran:
+1 (not tested, there is something wrong with getting to the survey page on my VM, not JS related)

kdeloach:
+1 Verified that "distance to end" validation errors will scroll to the correct field.

Issue #199, opened 2014-12-24
Allow census admins to map any block, regardless of individual mapper or trusted mapper status


jwalgran:

dboyer:
@jwalgran Parks would like census admins to be able to edit data anywhere. So they automatically get individual mapper status and can map in any group's area without requiring approval from the group first. They can also map in a group's area even if that group has denied individual mappers.

kdeloach:
I will verify the existing behavior, ensure that census admins can reserve/survey any block regardless of individual/trusted mapper status, and add unit tests.

kdeloach:
Didn't realize this was not assigned to the current sprint milestone. Will hold off on this until we finish all issues in the current sprint first.

jwalgran:
Parks has requested, and I think it is a good idea, to use ``is_superuser = 't' `` to lift the ``group_territory`` restriction in https://github.com/azavea/nyc-trees/blob/develop/src/tiler/sql/user_reservable.sql#L29-L31

jwalgran:
This doesn't have to happen this sprint, so I am going to put it into the backlog.

PR #1724, opened 2015-05-22
Allow census admins to map any available block, regardless of group.

maurizi:
Adds a separate reservable layer for census admins, which shows allows
them to map any available unreserved block, regardless of trusted mapper
status for blocks in group territory.

Connects to #199


kdeloach:
+1 Looks good but I have not tested it.

RickMohr:
+1. As a census admin I reserved and mapped a block in a group's territory. I could not reserve a block that had already been mapped in that territory. As a non-census-admin I could not reserve any blocks in that territory.

PR #1725, opened 2015-05-22
Set search zindex to appear higher than leaflet controls

jfrankl:
image


maurizi:
@jfrankl Should I close https://github.com/azavea/nyc-trees/pull/1720 in favor of this?

jfrankl:
Sorry, I got mixed up and thought that was an issue rather than a PR.

Issue #1728, opened 2015-05-23
Tree shapefile dump fails because of a null geometry [Part of 1714]


kdeloach:
Occurred in production on May 23 2015 10:30 AM EST.

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/census_admin/tasks.py", line 117, in dump_trees_to_shapefile
    geom=0, geom_type='Point')
  File "/opt/app/apps/census_admin/tasks.py", line 194, in _dump_to_shapefile
    'geometry': json.loads(get_val(row, geom)),
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
TypeError: expected string or buffer
```

jwalgran:
Using a production data dump, I ran [trees.sql](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/census_admin/trees.sql) with an additional ``geom IS NULL`` clause and it returned a single row for which a geometry was not generated.

```
 geom |  id  | survey_id | species_id | distance_to_tree | distance_to_end | circumference | stump_diameter | curb_location  | status | species_certainty | health | stewardship | guards | sidewalk_damage | problems |          created_at           |          updated_at
------+------+-----------+------------+------------------+-----------------+---------------+----------------+----------------+--------+-------------------+--------+-------------+--------+-----------------+----------+-------------------------------+------------------------------
      | 2200 |       414 |        126 |               24 |            15.2 |            27 |                | OffsetFromCurb | Alive  | Yes               | Good   | 3or4        | None   | NoDamage        | None     | 2015-05-22 22:04:26.153112+00 | 2015-05-22 22:04:26.15314+00
```

Survey details

![screen shot 2015-05-26 at 9 10 23 am](https://cloud.githubusercontent.com/assets/17363/7817071/19cd5836-0387-11e5-9e1d-13726154fc46.png)

[...]

PR #1732, opened 2015-05-26
Update layoutBoxes to not fail on short block

jwalgran:
The tree shapefile export started failing because the SQL query that uses layoutBoxes to generate a geometry for each surveyed tree generated a NULL geometry for the second of two trees in production survey 414. Increasing the tiny offset used to generate a short segment at the tree location avoids the issue.

Connects to #1728


strk:
See https://github.com/treekit/treekit/commit/1ef16d75d4a4b54b90deedc1601b2c110d0c2058 for a better handling of this.

strk:
I'd recommend upgrading the box layout code to https://github.com/treekit/treekit/commit/efb8e376bbf4863934c8c7c69457d96b63d1199e

jwalgran:
Closing in favor of pulling in upstream changes

Issue #1664, opened 2015-05-14
No error validation messages shown on iPhones


maurizi:
Steps to reproduce:
1. Go to the treecorder.
2. Leave a field unchecked
3. Go to the bottom and choose "Add another Tree"

Expected: It focuses on the first field with an error and displays a bubble with a validation message
Actual: It focuses on the first incorrect field, but there is no validation message.

Error validations do appear for me on Firefox (Android and desktop), but not on iOS (iPhone 4 on iOS7 and iPhone 4S on iOS 8).

Do some research into existing solutions for showing custom validation messages.

jwalgran:
iOS does not show form validation messages according to http://caniuse.com/#feat=form-validation

![can-i-use](https://cloud.githubusercontent.com/assets/17363/7642201/16f40926-fa44-11e4-9b1e-bb86deed7355.png)

maurizi:
We do still get error *checking*, so users won't be submitting invalid data, like negative numbers.

However, not knowing why clicking "Submit" made the screen focus on a field is not very good UX.
We could show validation messages ourselves for iOS devices, instead of relying on the browser to do so.

dboyer:
@jfrankl What are your thoughts on this from a UX perspective?

jfrankl:
It would definitely be preferable to have validate messages. Otherwise, it
just looks like you accidentally scrolled up on the page. I'm more
concerned about someone not knowing that the form didn't validate than
knowing why it didn't validate (though I'm concerned about both). Right now
it's not clear that you need to make a change.

On Wed, May 20, 2015 at 4:38 PM, dboyer <notifications@github.com> wrote:

> @jfrankl <https://github.com/jfrankl> What are your thoughts on this from
> a UX perspective?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/azavea/nyc-trees/issues/1664#issuecomment-104030780>.
>



-- 
Jeff Frankl, UI/UX Designer

Azavea  |  340 N 12th St, Suite 402, Philadelphia, PA
Web azavea.com  |  Blog azavea.com/blogs  |  Twitter @azavea
<http://twitter.com>

PR #1736, opened 2015-05-27
Show error validation messages on iOS and old Android browsers

maurizi:
Switches from using builtin browser validation messages to custom messages.

We need to show custom error messages on iOS and older Android browsers, as they will not do so for us.  Since we are doing so for some browsers, I have opted to switch to custom error messages on all browsers.

We still rely on html5 for the actual validation checking, which does work on all mobile browsers.

Connects to #1664

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/partials/radios.html
+++ b/src/nyc_trees/apps/survey/templates/survey/partials/radios.html
@@ -7,6 +7,7 @@
 {% endif %}
 <div class="field">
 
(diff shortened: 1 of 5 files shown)


maurizi:
custom-errors

RickMohr:
Otherwise, works great!

maurizi:
@RickMohr I have added some animations when we hide elements after the data becomes valid.

I've tested it on firefox and Chrome in desktop and mobile emulation modes, can you retest on your iPhone?

RickMohr:
+1, big improvement!

PR #1737, opened 2015-05-28
Remove mouseover effect on progress circles on a group's page

### LINKED ISSUE #1735: Remove mouseover effect on progress circles on a group's page
While it matches the behavior on the dashboard/home page, it also implies that you can click the circles and see something, which you can't. Let's get rid of that effect.

jfrankl:
Mouseover effect does not show on group page. Adding ```.circles-nohover``` to ```.progress-circles``` will remove this effect from any group of progress circles.

Fixes #1735.


steventlamb:
+1, tested

PR #1738, opened 2015-05-29
Update layoutBoxes to 67a5f279 version

jwalgran:
Pulled in from https://github.com/treekit/treekit

The most important part of this update is https://github.com/treekit/treekit/commit/1ef16d75d4a4b54b90deedc1601b2c110d0c2058 which fixes a problem where null geometries were created for trees on short blocks.

Connects to #1728


jwalgran:
On hold while parks reviews the changes to the generated geometries.

jwalgran:
I pulled in the latest version of layoutBoxes that includes an additional fix for handling points beyond the end of the block.

jwalgran:
There is a detailed analysis in the fileshare (data/nyc-trees-layout-boxes-comparison-2015-06-02) showing that this updated geometry constructor fixes the null geometry problem but does not generate different point locations that previous versions.

Issue #1696, opened 2015-05-19
Remove blockface geom field from the Django admin


kdeloach:
The field does not display properly, and we don't want editing to be available anyway.

![image](https://cloud.githubusercontent.com/assets/43062/7707030/e0f66ac2-fe1b-11e4-8c85-56dd0898ddcb.png)

dboyer:
Unless this is a quick fix, we don't need to address it since the census zone can be seen on the publicly visible group detail page.

PR #1741, opened 2015-06-02
Load openlayers from SSL-compatible CDN

steventlamb:
connects to #1696 on github

@dboyer @kdeloach in #1696, you suggested removing this widget because it's broken. This PR should fix it. I'm still open to removing it though.

### DIFF
--- a/src/nyc_trees/apps/survey/admin.py
+++ b/src/nyc_trees/apps/survey/admin.py
@@ -24,6 +24,10 @@ class SurveyAdmin(admin.ModelAdmin):
 class BlockfaceAdmin(admin.ModelAdmin):
     search_fields = ('id',)
     list_filter = ('is_available',)
+    exclude = ('geom',)
+
+    def has_add_permission(self, request):
+        return False
 
 
 class BlockfaceReservationAdmin(admin.ModelAdmin):


maurizi:
Unless there is a way to make the map uneditable, I think we should remove the field entirely.  We don't want a census admin editing the blockface geometry, and so we shouldn't provide that functionality.

My 2¢

steventlamb:
ok, pushed another commit to handle it that way. I have no preference.

src/nyc_trees/apps/survey/admin.py
```diff
@@ -6,6 +6,8 @@
 from django.contrib import admin
 import apps.survey.models as m
 
+from django.contrib.gis.admin.options import GeoModelAdmin
```
hectcastro: This unused import is causing the build to fail. Only mentioning it here to speed up feedback since we may (or may not) want this in the next release.

kdeloach:
I will test this shortly. Consider opening up a PR against the release branch.

steventlamb:
ok. So we'll go ahead with the full fix (removing the field), both commits?

kdeloach:
I think so.

kdeloach:
+1

steventlamb:
closing in favor of https://github.com/azavea/nyc-trees/pull/1744 (new branch name)

PR #1744, opened 2015-06-03
Hide geom field in blockface admin

steventlamb:
connects to #1696 on github

we never want the user to be able to edit it, and it also is not working
as currently implemented.


kdeloach:
+1 Branch name is more descriptive.

Issue #1739, opened 2015-05-29
Exception when attempting to send bulk event email


jwalgran:
```
/group/bronx-park-east-community-association/event/cancelled-training-530/email/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 32, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/opt/app/apps/core/decorators.py", line 58, in wrapper
    return view_fn(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 79, in wrapper
    params = callable_or_dict(request, *args, **wrapper_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/opt/app/apps/event/views.py", line 183, in event_email
    send_to(user, 'event', **msg_kwargs)
  File "/opt/app/apps/mail/libs.py", line 46, in send_to
    to = user.email
AttributeError: 'NoneType' object has no attribute 'email'

kdeloach:
This group does not have a group admin. The workaround could be to just assign one.

jwalgran:
Error logged again May 29th 2015, 17:30:30.120

jwalgran:
I am going to let Parks know about the workaround.

PR #1745, opened 2015-06-03
Fix exception when sending bulk event email

kdeloach:
This fixes the case where sending event emails for groups with no group
admin would throw an exception.

To test:
1. Find an event for a group with no group admin (or set the group
admin to null)
2. Send a bulk event email
3. Email should send successfully

Connects #1739


jwalgran:
Looks good. Testing now.

jwalgran:
+1 I replicated the bug on ``develop`` and verified the fix on this branch

Issue #1713, opened 2015-05-21
Requests to /favicon.ico making it to Django


hectcastro:
Having Django even think about this request is probably better avoided. Preferably Nginx would deal with it.

PR #1746, opened 2015-06-04
Serve /favicon.ico from nginx

steventlamb:
connects #1713 on github

primarily use favicon link tags to server more robustly. however, some clients send requests to /favicon.ico, putting load on gunicorn appservers. This as a failsafe against those cases.


deployment/ansible/roles/nyc-trees.app/templates/nginx-nyc-trees-app.conf.j2
```diff
@@ -6,6 +6,20 @@ server {
 
   access_log /var/log/nginx/nyc-trees-app.access.log logstash_json;
 
+  location /favicon.ico {
```
hectcastro: I think that you should be able to delete `src/nyc_trees/apps/core/static/favicon.ico` and replace this entire `location` block with:

```nginx
location = /favicon.ico {
  rewrite /favicon.ico /static/img/favicon/favicon.ico;
}
```

steventlamb: I'm not sure this will work, because in this case the mentioned `favicon.ico` file will have a hash segment appended to it in production. i just deleted my static folder and ran `npm run build` and this is my output:
```
vagrant@app:/var/www/nyc-trees/static/img/favicon$ ls
android-chrome.3383adc6.png    favicon.cabfcaf1.ico
apple-touch-icon.1d8bb3b9.png  manifest.f18abea0.json
```
So I think the easiest way to handle it is to make a copy in a separate path that is outside of the rewrite rules, as I've done with this PR. Does that sound right?

steventlamb: I'm happy to replace the `alias` part with `rewrite`.

hectcastro: I have to re-test today. After provisioning in test mode, the favicon file at that path didn't have a hash.

hectcastro: You're right. I didn't clear things out first.

Is the `favicon.ico` file in this PR valid? When I try to view it, I get something that looks like a broken image.

steventlamb: I'm pretty sure it was working when I was making requests to /favicon.ico but I'll definitely double check. I'm having the same issue with github as you. The only explanation I can think of is that this is actually a `.png` file renamed to `favicon.ico`. Seems like that should work as per http://en.wikipedia.org/wiki/ICO_%28file_format%29.

jwalgran:
Are we getting burned by ``.ico`` not being listed as binary in ``.gitattributes``?

steventlamb:
good idea, I'll look at / do that as well

steventlamb:
@hectcastro @jwalgran updated to use rewrite instead of alias, fix `.gitattributes` and broken image.

hectcastro:
:+1:

PR #1748, opened 2015-06-04
Feature/jdf/fix circle ios

### LINKED ISSUE #1747: Clicking species and users progress circles on mobile Safari on iOS doesn't load pop-up window
I was not logged in and was on the homepage. Clicking the blocks progress circle takes you to the map, but nothing happens when you click the users or species progress circles.

jfrankl:
Fixes #1747.

Not sure why, but not having cursor: pointer on the progress circles prevents these modals from opening. Adding them seems to fix this.


jwalgran:
+1 tabs on users and species now show the popups

PR #1752, opened 2015-06-08
Update CloudFormation templates so that resources are up-to-date

hectcastro:
In the process of creating a new stack from an existing template, the following resources and their parameters became stale and required updates:

- CloudFront Distribution/MinimumProtocolVersion

If you specify the `IamCertificateId` property and specify SNI only for the `SslSupportMethod` property, you must use `TLSv1` for the minimum protocol version. If you don't specify a value, AWS CloudFormation specifies `SSLv3`.

- AutoScalingGroup/NotificationConfigurations

`NotificationConfiguration` is now `NotificationConfigurations` (plural) and now accepts an array of `NotificationConfigurations`.

### DIFF
--- a/deployment/requirements.txt
+++ b/deployment/requirements.txt
@@ -1,4 +1,4 @@
 ansible==1.8.4
 awscli==1.7.13
 boto==2.34.0
-troposphere==0.7.1
+troposphere==1.0.0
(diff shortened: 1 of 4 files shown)


kdeloach:
+1

Issue #1754, opened 2015-06-09
Private events do not appear on registrants dashboard under the events they're attending


dboyer:
After a user RSVPs to a private event, that event doesn't appear under the list of events the user is attending that are shown on the user's dashboard.

jwalgran:
I confirmed this fix in staging.

PR #1756, opened 2015-06-10
Show private events in user event lists

RickMohr:
"Your Events This Week" on the Events page wasn't showing a user's private events.
* Removed the `is_private=False` filter from `EventList.immediate_events`
* That routine is used nowhere else, so the change is safe

The dashboard wasn't showing a user's private events.
* Removed the `is_private=False` filter from `EventList.all_events`
* That routine is also used with the `EventList.trainingFilters`
  so they need an `is_private=False` filter
* I believe that no other lists are affected by this change, and I think I tested them all,
  but with the multiple levels of indirect linkages it's hard to tell

Connects #1754

### DIFF
--- a/src/nyc_trees/apps/event/event_list.py
+++ b/src/nyc_trees/apps/event/event_list.py
@@ -125,9 +125,11 @@ class EventList(object):
         right_now = now()
         filtersets = {
             EventList.trainingFilters: OrderedDict([
-                (_ALL, None),
-                (_TRAINING, lambda qs: qs.filter(includes_training=True)),
-                (_MAPPING, lambda qs: qs.filter(includes_training=False)),
+                (_ALL, lambda qs: qs.filter(is_private=False)),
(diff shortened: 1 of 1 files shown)


RickMohr:
@azavea-bot rebuild

maurizi:
With this change, I am able to see a private event (that I am not RSVPed to) from a group I am following on the "recommended" section on the dashboard.

RickMohr:
@maurizi, thanks for the catch. Fixed.

steventlamb:
With this change, `trainingFilters` are no longer abstract and semantic, because they carry knowledge of their specific implementation. I think it would be better to decouple the two uses of `all_events` and make a new `EventList` called `private_events` with the filter applied on its base queryset, because new `EventList`s are cheap. Then have that event list use the (unmodified) `trainingFilters`.

RickMohr:
After some investigation we decided to leave this as-is since it works.

steventlamb:
+1

Issue #1734, opened 2015-05-26
Tiler processes respawned


jwalgran:
There was a tiler 5xx alarm from AWS at Tuesday 26 May, 2015 21:19:38 UTC

Kibana has 4 "tiler" entries around that time

/#/discover?_g=(time:(from:'2015-05-26T21:10:12.187Z',mode:absolute,to:'2015-05-26T21:40:12.188Z'))&_a=(columns:!(_source),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:tiler)),sort:!('@timestamp',desc))

```javascript
{
  "_index": "logstash-2015.05.26",
  "_type": "logs",
  "_id": "AU2SF34roolW4VnYJwQC",
  "_score": null,
  "_source": {
    "@version": 1,
    "@timestamp": "2015-05-26T21:18:37.775+00:00",
    "source_host": "ip-10-0-1-63",
    "message": "May 26 21:18:37 kernel:[ 3838.466156] init: nyc-trees-tiler main process (817) terminated with status 8",
    "facility": "user",
    "severity": "warning",
    "program": "kernel",
    "processid": "-",
    "tags": [
      "syslog"
    ],
    "host": "10.0.1.63:58249"
  },
  "fields": {
    "@timestamp": [
      1432675117775
    ]
  },
  "highlight": {
    "message": [
      "May 26 21:18:37 kernel:[ 3838.466156] init: nyc-trees-@kibana-highlighted-field@tiler@/kibana-highlighted-field@ main process (817) terminated with status 8"
    ]
  },
  "sort": [
    1432675117775
  ]
}
```

```javascript
{
  "_index": "logstash-2015.05.26",
  "_type": "logs",
  "_id": "AU2SF34roolW4VnYJwQD",
  "_score": null,
  "_source": {
    "@version": 1,
    "@timestamp": "2015-05-26T21:18:37.775+00:00",
    "source_host": "ip-10-0-1-63",
    "message": "May 26 21:18:37 kernel:[ 3838.466167] init: nyc-trees-tiler main process ended, respawning",
    "facility": "user",
    "severity": "warning",
    "program": "kernel",
    "processid": "-",
    "tags": [
      "syslog"
    ],
    "host": "10.0.1.63:58249"
  },
  "fields": {
    "@timestamp": [
      1432675117775
    ]
  },
  "highlight": {
    "message": [
[...]

PR #1757, opened 2015-06-12
Prevent health check from setting HTTP response more than once per call.

maurizi:
Connects to #1734

### DIFF
--- a/src/tiler/.jshintrc
+++ b/src/tiler/.jshintrc
@@ -6,5 +6,6 @@
         "require": false,
         "module": true,
         "exports": true
-    }
+    },
+    "predef": [ "-Promise" ]
 }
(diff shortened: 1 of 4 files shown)


jwalgran:
Looks good. Pulling down and testing now.

jwalgran:
+1

Tests:

Repeated requests to the endpoint did not increase ``SELECT sum(numbackends) FROM pg_stat_database;`` from Postgres or the size of a ``CLIENT LIST`` from Redis.

I reduced the timeout to 5ms and repeatedly hit the endpoint. I was getting both successes and timeouts and did not see any errors in the console or crashes, or orphaned Redis connections.

I  misconfigured the NYC_TREES_CACHE_HOST and NYC_TREES_DB_HOST and saw timeouts and error messages without errors in the console or crashes.

PR #1766, opened 2015-06-17
Fix link in About & FAQ page

maurizi:


jwalgran:
+1

http://www.nycgovparks.org/trees/treescount/independent-mapping is the correct page

Issue #1765, opened 2015-06-17
Surveys being entered twice with same data


dboyer:
From Parks: There are a handful of scenarios where there will be two surveys per block with identical information, entered (roughly) 10 seconds apart.

This could be related to the user clicking "yes" twice when they perceive some delay in submitting the data.

Deb has screenshots of this so find her if you work on this. This is a bug fix that falls under warranty.

PR #1767, opened 2015-06-19
If submit succeeds, don't reenable the submit button

RickMohr:
Without this change I saw multiple posts when clicking "Submit" multiple times.
With this change I only saw one submit.

I also verified that you can't submit the same data twice by using the browser back button.

Connects #1765


maurizi:
+1, tested

Issue #1759, opened 2015-06-15
Add 75% opaque Stamen Toner label layer on top of block edges and switch to label-less basemap


jwalgran:
Mappers are having trouble seeing the street labels on the Positron base map, which are often covered by the block edge layer.

I hacked together a solution as a proof of concept in https://github.com/azavea/nyc-trees/tree/feature/jcw/toner-labels-on-top. It needs a little cleanup.

dboyer:
We also will only want to show the labels once the user has zoomed in a bit.

jwalgran:
The labels clash with the borough/neighborhood aggregate layer, so we should only show them when we are at the block edge zoom levels on the progress map. We could show the labels at all levels on the treecorder, since there are no conflicting features.

maurizi:
Some complication to this:

 - There is *no* way to have a tile layer above a vector layer in the version of leaflet we're using (see https://github.com/Leaflet/Leaflet/issues/2395).  This means that the currently selected blockface on the progress map and the shopping cart layer on the reservations page will still be occluding the map labels.

 - I'm running into the same label issue we saw when going from neighborhood -> borough layers on the progress map, though I assume we can use a similar workaround.

maurizi:
Do we want to have the labels be *under* the neighborhood/borough polygons when zoomed out on the progress page?  It looks a little barren otherwise.

jwalgran:
Parks has seen these screenshots and are OK with them. When you are zoomed in to a reasonable level the streets are visible.

![screen shot 2015-06-12 at 11 50 18 am](https://cloud.githubusercontent.com/assets/17363/8219217/42487d36-14fd-11e5-8196-8b253cbe92d3.png)

![screen shot 2015-06-12 at 11 50 26 am](https://cloud.githubusercontent.com/assets/17363/8219233/65909ba2-14fd-11e5-93f9-1c8fc8904ca1.png)

I don't think we should show the labels under the borough/hood polygons in the first pass. It will be easy to implement that way (as a simple minZoom property on the label layer), and we can get an opinion from Parks on how it looks.

maurizi:
If the main concern is on the treecorder map, then it would be easiest to show labels on top for everything *but* the progress map (which would just stay the same).

The label layer is running into the same issues we had with other other layers on the progress map, where even when setting a minZoom it leaves an afterimage when you zoom past it's minZoom level.
[...]

PR #1768, opened 2015-06-19
Use stamen toner labels on top of a label-less basemap for most maps.

maurizi:
This is intended to prevent the blockface tile layer from obscuring the
street names, to help in identifying the correct blockface while mapping.

I kept the progress map as it was, because I was running into issues
with labels interfering with the labels on the aggregate neighborhood
and borough layers.

I also excluded the event list map and the survey detail map, because
they do not have a blockface layer, so there is no need for a separate
label layer.

Thanks to @jwalgran for the initial implementation!

Connects to #1759 

To test, please check that labels are visible and above the blockface layer on the following maps: reservations, treecorder, census zone administration, reservations PDF, and event PDF.

### DIFF
--- a/src/nyc_trees/js/src/event_list_page.js
+++ b/src/nyc_trees/js/src/event_list_page.js
@@ -23,6 +23,9 @@ $(dom.mapToggle).one('shown.bs.tab', function() {
             mapModule.create({
                 geolocation: true,
                 search: true,
+                // There is no raster tile layer, so labels embedded in the
+                // basemap will work just fine
+                withLabels: true
             });
             return;
         }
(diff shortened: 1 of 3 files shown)


jwalgran:
+1 All the maps look good

screen shot 2015-06-19 at 12 04 16 pm
screen shot 2015-06-19 at 11 57 30 am
screen shot 2015-06-19 at 11 56 14 am
screen shot 2015-06-19 at 11 53 06 am
screen shot 2015-06-19 at 11 52 19 am
screen shot 2015-06-19 at 11 51 44 am
screen shot 2015-06-19 at 11 49 11 am
screen shot 2015-06-19 at 12 07 37 pm

PR #1769, opened 2015-06-19
Fix label positions on the progress map for concave polygons.

maurizi:
Some of the neighborhood polygons on the progress map have centroids that
lie outside the polygon, which leads to their percentage label being
placed inside a different neighborhood (which will then have two labels).

By adding `text-placement: interior` to the CartoCSS, we tell Mapnik to
prefer placing labels on the interior of polygons if the centroid is not
inside the poygon.

Prior to this fix:
centroid1

After:
fixed2


jwalgran:
+1 (I did not test this locally)

Issue #1749, opened 2015-06-04
Change Treecorder zoom level


dboyer:

dboyer:
Zoom in to show just a couple blocks around the event location, the user's location, or the reserved blocks.

PR #1770, opened 2015-06-22
Set a higher zoom level when zooming to a location

jwalgran:
This commit introduces a new zoom level "preset" to be used when zooming to the geolocation of the device, or to the location of a mapping event. This is a usability improvement suggested by Parks based on feedback from mapping events.

Geolocation zoom before

screen shot 2015-06-22 at 10 11 21 am

After

screen shot 2015-06-22 at 10 13 27 am

Connects to #1749


steventlamb:
+1, tested

Issue #1764, opened 2015-06-17
Add dot indicating user's location on the Treecorder map


dboyer:
Open question whether the dot would move with the user or just be placed at their initial location.

maurizi:
We should stop tracking location once the survey form is up, to avoid draining battery life.

PR #1772, opened 2015-06-22
Show user location on Treecorder map

RickMohr:
For testing, points near Azavea are shown in southern Manhattan.

To test:
* Use a mobile device (ideally a tablet, wide enough to show both the map and the input form)
* Reserve some blocks at the southern tip of Manhattan
* Open Treecorder. You should see a green dot somewhere around there.
* Walk around. The dot should move.
* Start entering data. The dot should disappear.

Connects #1764

### DIFF
--- a/src/nyc_trees/js/src/map.js
+++ b/src/nyc_trees/js/src/map.js
@@ -16,7 +16,9 @@ module.exports = {
     addGridLayer: addGridLayer,
     fitBounds: fitBounds,
     getDomMapAttribute: getDomMapAttribute,
-    hideCrosshairs: hideCrosshairs
+    hideCrosshairs: hideCrosshairs,
+    startTrackingUserPosition: startTrackingUserPosition,
+    stopTrackingUserPosition: stopTrackingUserPosition
 };
 
 function create(options) {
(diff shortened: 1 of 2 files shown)


src/nyc_trees/js/src/map.js
```diff
@@ -211,3 +213,57 @@ function getDomMapAttribute(dataAttName, domId) {
...
+}
+
+function onLocationFound(e) {
+    var map = this,
+        marker = map._userLocationMarker,
+        latlng = e.latlng;
+
+    // For testing, move points near Azavea to southern Manhattan
+    var azaveaLatLng = L.latLng(39.9583208, -75.1585257),
+        manhattanLatLng = L.latLng(40.7030809, -74.0129269);
```
maurizi: I'm not sure you want these in here...

RickMohr: This PR can't be tested or demonstrated from Azavea without that code. It should be safe to leave it in, and we can also ask Parks to test it from staging and then remove it.

steventlamb:
+1 tested

Issue #1726, opened 2015-05-22
Improve the performance of rendering the teammate autocomplete


jwalgran:
As of  6c1d6bd1, on a vagrant VM, rendering the ``survey.html`` template with 2200+ users causes gunicorn to peg the CPU and not finish within the Nginx timeout. The same code, with the same data works in production.

Inlining the ``teammate_item.html`` template into ``teammates.html`` prevents the CPU spike in development, but performance degrades as you increase the user count.

maurizi:
The fact that inlining the template "fixed" the issue makes me suspicious that the Django debug toolbar is to blame, since it shows every template that was included in the page and context for each one.

jwalgran:
Parks is also getting feedback this is now slow an unwieldy with ~3000 users. They suggested making the teammate list work like the species list, where you must enter a few characters before the autocomplete list is displayed. This is a good idea, but there could be users with 1 or 2 character usernames, so, we need to make sure those users are selectable.

dboyer:
This is unblocked and we should go ahead and implement.

PR #1773, opened 2015-06-25
Remove inline teammate autocomplete in favor of AJAX backed autocomplete

maurizi:
The select tag based autocomplete was proving to be not performant enough
with a significant number of users (over 3000).

Moving to an AJAX-based autocomplete should alleviate the performance
issues we have encountered with it.

Connects to #1726

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/partials/teammate_item.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/teammate_item.html
@@ -1,4 +1,0 @@
-<option value={{ teammate.user_id }}>
-    {{ teammate.username }}
-    {% if teammate.full_name %} &mdash; {{ teammate.full_name }} {% endif %}
-</option>
--- a/src/nyc_trees/apps/event/templates/event/partials/teammates.html
+++ b/src/nyc_trees/apps/event/templates/event/partials/teammates.html
@@ -1,7 +1,0 @@
-<select class="teammate" style="width: 100%;">
-    {# A blank option to allow for clearing the selection #}
-    <option value="">&nbsp;</option>
-    {% for teammate in teammates %}
-        {% include "event/partials/teammate_item.html" %}
-    {% endfor %}
-</select>
--- a/src/nyc_trees/apps/survey/helpers.py
+++ b/src/nyc_trees/apps/survey/helpers.py
@@ -3,33 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
-from apps.core.models import User
-from apps.users import can_show_full_name
-
 from apps.survey.models import Survey, Territory
-
-
-def teammates_for_mapping(current_user):
-    attendees = User.objects.exclude(id=current_user.id) \
-                            .filter(is_active=True) \
-                            .order_by('username')
-    return _teammates_context(attendees)
-
-
-def _teammates_context(users):
-    return [_teammate_user_context(u) for u in users]
-
-
-def _teammate_user_context(user):
-    if can_show_full_name(user):
-        full_name = ("%s %s" % (user.first_name, user.last_name)).strip()
-    else:
-        full_name = ""
-    return {
-        "user_id": user.id,
-        "username": user.username,
-        "full_name": full_name
-    }
 
 
 def group_percent_completed(group):
--- a/src/nyc_trees/apps/survey/routes.py
+++ b/src/nyc_trees/apps/survey/routes.py
@@ -130,6 +130,9 @@ blockface = route(GET=do(json_api_call, v.blockface))
 
 blockface_near_point = route(GET=do(json_api_call, v.blockface_near_point))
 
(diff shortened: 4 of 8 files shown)


kdeloach:
:clap:

jwalgran:
There is a style regression that can be fixed with ``<input id="teammate-select" />`` → ``<input class="teammate" id="teammate-select" />``

jwalgran:
There is some crummy grammar in the select2 help text

screen shot 2015-06-26 at 8 22 55 am

Looks like this is the offending line

https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/shim/select2.js#L3505
```javascript
formatInputTooShort: function (input, min) { var n = min - input.length; return "Please enter " + n + " or more character" + (n == 1 ? "" : "s"); },
```

We may want to patch this line directly in ``shim/select2.js`` rather than worry about an upgrading the library to get such a small fix. 

stego

maurizi:
@jwalgran I pushed a change to fix the styling regression.  As for the select2 text, I'm not really sure what you're suggesting I change it to.  The grammar seems Ok, if not perfect.  Do you have a suggestion for what to change it to?

jwalgran:
+1

On my iPhone 4s I verified that I can still search for teammates with either username or full names (when the mapper has made their full name public)

I exercised this with Chrome mobile mode throttling the connection to "Regular 2G 250kbps". Typing the letter "a" matches just over 1600 results, and the response returned in about 800ms, which is good.

screen shot 2015-06-26 at 9 41 46 am

On "Good 3g" this drops to < 200ms

screen shot 2015-06-26 at 9 51 39 am

I dumped the query generated by the endpoint and it runs in about ~7ms with current production data, which is also good, and suggests that we will be able to handle a much larger user count in the future if we need to.

```
EXPLAIN ANALYZE SELECT "core_user"."id", "core_user"."password", "core_user"."last_login", "core_user"."is_superuser", "core_user"."username", "core_user"."first_name", "core_user"."last_name", "core_user"."email", "core_user"."is_staff", "core_user"."is_active", "core_user"."date_joined", "core_user"."created_at", "core_user"."updated_at", "core_user"."individual_mapper", "core_user"."requested_individual_mapping_at", "core_user"."profile_is_public", "core_user"."real_name_is_public", "core_user"."group_follows_are_public", "core_user"."contributions_are_public", "core_user"."achievements_are_public", "core_user"."is_flagged", "core_user"."is_banned", "core_user"."is_ambassador", "core_user"."is_minor", "core_user"."zip_code", "core_user"."referrer_parks", "core_user"."referrer_group_id", "core_user"."referrer_ad", "core_user"."referrer_social_media", "core_user"."referrer_friend", "core_user"."referrer_311", "core_user"."referrer_other", "core_user"."field_training_complete", "core_user"."training_finished_getting_started", "core_user"."training_finished_the_mapping_method", "core_user"."training_finished_tree_data", "core_user"."training_finished_tree_surroundings", "core_user"."training_finished_wrapping_up", "core_user"."training_finished_intro_quiz", "core_user"."training_finished_groups_to_follow", "core_user"."opt_in_stewardship_info", "core_user"."progress_page_help_shown", "core_user"."survey_geolocate_help_shown", "core_user"."reservations_page_help_shown", "core_user"."reservation_ids_in_map_pdf", "core_user"."reservations_map_pdf_filename" FROM "core_user" WHERE (NOT ("core_user"."id" = 611) AND "core_user"."is_active" = True AND (UPPER("core_user"."username"::text) LIKE UPPER('%a%') OR ("core_user"."real_name_is_public" = True AND (UPPER("core_user"."first_name"::text) LIKE UPPER('%a%') OR UPPER("core_user"."last_name"::text) LIKE UPPER('%a%'))))) ORDER BY "core_user"."username" ASC;

 Sort  (cost=321.12..327.45 rows=2531 width=206) (actual time=7.162..7.253 rows=1638 loops=1)
   Sort Key: username
   Sort Method: quicksort  Memory: 519kB
   ->  Seq Scan on core_user  (cost=0.00..178.05 rows=2531 width=206) (actual time=0.013..2.951 rows=1638 loops=1)
         Filter: (is_active AND (id <> 611) AND ((upper((username)::text) ~~ '%A%'::text) OR (real_name_is_public AND ((upper((first_name)::text) ~~ '%A%'::text) OR (upper((last_name)::text) ~~ '%A%'::text)))))
         Rows Removed by Filter: 1382
 Total runtime: 7.374 ms
```

jwalgran:
On the topic of the help text grammar,  when a sentence includes "or more" I think it suggests using the plural noun. The latest version of select2 implements this.

https://github.com/select2/select2/blob/bd6148e39f753ed2880f674a90b42ffe446b893a/src/js/select2/i18n/en.js#L21

```javascript
var message = 'Please enter ' + remainingChars + ' or more characters';
```

Googling reveals that there is some contention between grammar experts on the topic. I am cool with leaving it alone.

PR #1778, opened 2015-07-17
Allow logging in via either username or email address

maurizi:
This allows email login for both normal login and the Django Admin site,
but I only updated the labels and error messages for the normal login page.

Connects to #1777

### DIFF
--- a/src/nyc_trees/apps/login/forms.py
+++ b/src/nyc_trees/apps/login/forms.py
@@ -72,9 +72,25 @@ class NycRegistrationForm(RegistrationFormUniqueEmail):
     )
 
     def clean_username(self):
-        if self.cleaned_data['username'].lower() in User.reserved_usernames:
+        username = self.cleaned_data['username']
+
+        if username.lower() in User.reserved_usernames:
             raise ValidationError('This username is not available.')
+
+        if User.objects.filter(email__iexact=username).exists():
+            raise ValidationError('That username is already in use. '
+                                  'Please supply a different username.')
+
         return super(NycRegistrationForm, self).clean_username()
+
+    def clean_email(self):
+        email = self.cleaned_data['email']
+
+        if User.objects.filter(username__iexact=email).exists():
+            raise ValidationError('That email address is already in use. '
+                                  'Please supply a different email address.')
+
+        return super(NycRegistrationForm, self).clean_email()
 
 
 class UsernameOrEmailPasswordResetForm(forms.Form):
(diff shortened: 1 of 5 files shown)


src/nyc_trees/apps/login/forms.py
```diff
@@ -24,9 +24,11 @@
 
 
 class NycAuthenticationForm(AuthenticationForm):
+    username = forms.CharField(max_length=254,
+                               label="Username or Email Address")
 
     error_messages = {
-        'invalid_login': 'Please enter a correct username and password. '
+        'invalid_login': 'Please enter a correct username/email and password. '
                          'Note that password is case-sensitive',
```
maurizi: @dboyer How does the changed text look on this page?

maurizi:
reg

src/nyc_trees/apps/login/forms.py
dboyer: Can it say username or email instead of username/email?

maurizi: So `Please enter a correct username or email and password.`?

dboyer: I suppose it should be email address, but otherwise yes. I feel like the slash could be interpreted as the user needed to enter both.

maurizi:
@steventlamb I updated this to remove the dependency on django-dual-authentication, can you take another look?

src/nyc_trees/libs/auth_backend.py
```diff
@@ -0,0 +1,32 @@
...
+
+from django.contrib.auth.backends import ModelBackend
+
+from apps.core.models import User
+
+
+class EmailOrUsernameAuthentication(ModelBackend):
+    def authenticate(self, username=None, password=None):
+        user = None
+        try:
```
steventlamb: just to be safe - are we sure that `@` and `.` are not allowed in usernames?
I'm imagining a case with username/password/emails: `foo@bar.com / qux / treesRgood@gmail.com` and `lenny123 / qux / foo@bar.com`. The former user could accidentally log into the latter user's account.

maurizi: Hmm... `@` and `.` _are_ [allowed in usernames](https://github.com/django/django/blob/master/django/contrib/auth/models.py#L307-L309), and we already have some users who used email addresses as their username.

It's unlikely that someone signed up with an email address as their username which is not _their_ email, but I suppose we should try and handle that somehow.

maurizi: I've pushed a commit which should address this by adding some extra validation to prevent this cornercase from happening.

steventlamb:
+1, tested, works and looks good

Issue #1780, opened 2015-07-29
Show mapped and unmapped blocks on event map PDF


dboyer:
Use the colors from the progress map (green for mapped and brown for unmapped). May be useful to do in conjunction with #1779

jwalgran:
For this map to be accurate, any time a block is mapped within the territory, we need to trigger the regeneration of the PDF. When an event is in progress, this would trigger many times, enough to block our celery queues uncomfortably. As part of the implementation we should test how the staging environment reacts to a large number of queued map render requests.

maurizi:
It is possible to cancel a queued celery task, so one way to handle large
queues is to save the celery I'd in the DB and cancel any outstanding jobs
for the event when retriggering a PDF job.

dboyer:
Parks understands that the map currently only regenerates when the event location moves and is fine with keeping that as the only thing that causes a new map.

dboyer:
This feature is more about being able to assign block edges for mapping at the beginning of an event rather than having a constantly updated map during the event.

jwalgran:
My concern was around this situation:

- An event is created and scheduled two weeks into the future. It is never edited.
- In those two weeks, individual mappers survey block edges around the event location.
- On event day, the original event map PDF shows block edges as available, but those edges have been surveyed.

This may not be a likely occurrence, but I want to make sure that all the event organizers understand how the application works, and that these maps may not be current at the time they are downloaded.

Issue #1779, opened 2015-07-29
Change zoom level on event PDF map


dboyer:
When we zoomed in the map that users see when they visit the Treecorder to show just a couple blocks around the event location, the event PDF maps also seemed to have been zoomed in to show just a couple blocks.

Instead, the event map PDF should be at a 1:10,000 level - what it used to be.

PR #1785, opened 2015-07-30
Fix initial map zoom levels, and show mapped/unmapped blocks on event map PDF

RickMohr:
Here's a shot from the event map PDF. 
image

In B/W :
image

Connects #1780 
Connects #1779


RickMohr:
waffle

maurizi:
+1

PR #1788, opened 2015-07-30
Add data migration causing all event map PDFs to be regenerated

RickMohr:
Note we did this before this is copied from `src\nyc_trees\apps\event\migrations\0012_auto_20150424_1705.py`

Connects #1787


RickMohr:
To test:
1) Look at an existing event map PDF. Block edges are red, and it may be zoomed in pretty far.
2) Run migrations
3) Look at the PDF again. Block edges are brown or green, and it is zoomed to "neighborhood" level.

maurizi:
+1, tested

Issue #1781, opened 2015-07-29
Add one decimal place to overall progress percentage


dboyer:
So that it would show 16.1% rather than 16%, for example.

PR #1789, opened 2015-07-31
Always show one decimal place in dashboard progress percentage.

maurizi:
Connects to #1781


RickMohr:
+1

We carefully hid the digit before so it wouldn't overflow the circle at some resolutions. But I can't make that overflow happen so I think we're fine.

Issue #1783, opened 2015-07-29
Add link to resend activation email if user requests password reset for inactive account


dboyer:
Only include link if it's been seven days or less since the account was created

jwalgran:
The date math should use ``settings.ACCOUNT_ACTIVATION_DAYS`` and ``date_joined`` field on ``User``

dboyer:
I'm waiting on some text from Parks about this.

PR #1791, opened 2015-07-31
Add pages for resending activation emails on a failed password reset.

maurizi:
Connects to #1783 

When attempting a password reset for a user that registered less than 7 days ago:
inactive_resend

When attempting a password reset for a user that registered more than 7 days ago:
inactive_fail

Blocked waiting for confirmation of the text.


RickMohr:
+1 Tested password reset:
- not activated, less than seven days
- not activated, greater than seven days
- activated

Issue #1801, opened 2015-08-13
Make it easier to see the mapped and unmapped blockedges on the territory assigning page


maurizi:

dboyer:
This does not count as support.

PR #1803, opened 2015-08-19
Make the territory map selected blocks transparent.

maurizi:
This should allow determining what color the underlying block edge is,
and if it is mapped or unmapped, while still making it easy to determine
which block edges are selected or unselected.

Connects to #1801


maurizi:
territory

RickMohr:
While this addresses the specific complaint it seems likely to add confusion, since there are now many new colors to understand and they aren't shown in the legend. What does orange mean? What does it mean if I click on a blue block and it turns green? Why is yellow in the legend?

As an alternative, how about using a dashed yellow (opaque) line for the selected blocks?

jwalgran:
I agree that transparency didn't work as well as I though due to color mixing. Skinny fully opaque yellow lines or using  [``dashArray``](http://leafletjs.com/reference.html#path-dasharray) may be a . If not, a big button to toggle the visibility of the yellow selection layer may be enough.

dboyer:
Yeah, let's avoid additional colors. There are enough already. I agree with Justin's suggestions.

maurizi:
Closed in favor of https://github.com/azavea/nyc-trees/pull/1804

Issue #1800, opened 2015-08-13
Investigate progress map coloring for block edges that were marked as unmappable for a reason


maurizi:

dboyer:
Blocks marked by the user as "unmappable" do not appear on the progress map as mapped (green) or count towards the progress counter although the block edge can no longer be reserved / is no longer available for mapping, as designed.

maurizi:
From @dboyer: Unmappable blocks could be a different color on the progress map, but they would count toward the progress counter.

dboyer:
Blocks marked as unmappable should be shown as a third color on the progress map - a grey color similar in shade to the green and brown on the map. Blocks marked as unmappable count toward the progress counts and are not available to reserve and map again.

PR #1806, opened 2015-08-24
Count quit surveys as completed for aggregate counts.

maurizi:
Surveys quit for a specific reason (intended for things which prevent
mapping the block edge, like construction) are now considered as
completed for the purposes of aggregate counts such as percent completed
on the home page and the block and neighborhood percentages on the
progress map.

For the more detailed view of the all progress layer, and for the user and
group progress layers, we show quit blocks as gray, and the legend shows
that they could not be mapped.

On the event detail map, I chose to show quit blocks the same color as mapped
blocks, since the more pertinent information there is what is available.

Connects to #1800

### DIFF
--- a/src/nyc_trees/apps/core/models.py
+++ b/src/nyc_trees/apps/core/models.py
@@ -226,7 +226,7 @@ class User(NycModel, AbstractUser):
     def surveys(self):
         """Return surveys user has participated in"""
         from apps.survey.models import Survey
-        return Survey.objects.complete().filter(
+        return Survey.objects.filter(
             Q(user=self) | Q(teammate=self))
 
     @property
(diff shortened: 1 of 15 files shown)


maurizi:
unmappable
unmappable1

maurizi:
@dboyer This is what the updated progress map looks like:

unmappable
unmappable1

jwalgran:
Works. I confirmed all of these correct behaviors:

- Quitting a survey increases the site wide block count on the dashboard.
- Quitting a survey in a territory increases the block count on the group page.
- Quitting a survey increases the borough block count.
- Setting a quit block back to available does not change the group, borough, or site-wide counts.
- Saving a second survey on a previously quit block does not change the group, borough, or site-wide counts.
- Quit blocks show up as grey on the progress and reservation maps and green on event PDF maps.
- Quit blocks are not selectable on the reservation map.

:+1:

dboyer:
This is picky, but is there a way to darken the grey? It looks very close to the grey of the streets right now. Otherwise, looks great.

maurizi:
@dboyer I made the color darker.

Issue #1808, opened 2015-08-31
Make migration 0029 backward compatible (Part of #1807)


jwalgran:
We need to add ``null=True`` to this migration to prevent crashing during the next blue/green deploy

https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/apps/survey/migrations/0029_survey_submit_comment.py

PR #1809, opened 2015-08-31
Make survey migration 0029 backward compatible

jwalgran:
Temporarily allowing null values allows the migrated database to be shared between old and new versions of the running application, making blue/green deployments easier to manage.

Connects to #1808


maurizi:
+1, didn't test

Issue #1810, opened 2015-09-01
Make survey.submit_comment non-nullable after release 12


jwalgran:
This field was made nullable for backward compatibility with the previous release.

PR #1811, opened 2015-09-01
Make Survey.submit_comment non-nullable again.

maurizi:
Now that we have pushed release 12 into production, it should be safe
to remove the `null=True` from this field.

Connects to #1810

### DIFF
--- a/src/nyc_trees/apps/survey/migrations/0030_auto_20150901_1244.py
+++ b/src/nyc_trees/apps/survey/migrations/0030_auto_20150901_1244.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


jwalgran:
I think this needs a data migration to ensure that any null values are replaced with ``''``. There was a small window when the migrations were applied in production but the old app servers were still running. There could be a survey row with a null ``submit_comment``. Verifying that there are not non-null ``submit_comment`` values in prod would also suffice.

maurizi:
@jwalgran It includes a one-off default in the migration itself.  Django won't let you make a nullable field non-null without providing a default value.

jwalgran:
That's a nice feature. I did not realize the default value would be applied retroactively.

RickMohr:
+1

Issue #1812, opened 2015-09-24
Add dot showing your location to Reservations pages


dboyer:
The Treecorder currently has a dot showing your location. Add it to Reservations as well.

steventlamb:
Don't change initial extent. Show dot if it happens to be in current extent.

dboyer:
Is this on staging? I can't see the dot but that may be due to my location outside NYC.

maurizi:
@dboyer You should be able to see the dot somewhere in Philadelphia (likely not our actual location but the location of our ISP).

dboyer:
Unfortunately, not. It zoomed to our location but I see no dot in the greater Philly/Delaware region.

jwalgran:
I loaded up the staging site, used the Chrome location emulator and saw a dot on the reservations page.

<img width="855" alt="screen shot 2015-10-01 at 3 50 28 pm" src="https://cloud.githubusercontent.com/assets/17363/10235769/4d40d9d6-6854-11e5-99cb-4d45cadf8e78.png">

PR #1816, opened 2015-09-28
Show user location on the reservations page.

maurizi:
Connects to #1812


RickMohr:
+1, right code, works!

Issue #1813, opened 2015-09-24
Display street labels on aerial imagery in Treecorder review page


dboyer:
If the labels overlap the tree dots, we may need to not implement them. Let's see how this looks.

steventlamb:
The labels will overlay the tree dots. Play with translucency to see if it's acceptable.

PR #1817, opened 2015-09-28
Show street label layer on survey detail page

RickMohr:
This works well, except that I'm getting 503 errors on Stamen label tiles
at zoom 19 (our default zoom for many block edges) and on some tiles at zoom 18.
I don't think it's our problem since I also see these errors at their site, e.g.
http://maps.stamen.com/toner-labels/#18/40.71925/-73.99044
I reported the problem using their "report a bug" form.

Connects #1813


maurizi:
+1, looks good.

Issue #1818, opened 2015-09-29
Change answer text for when will treescount 2015 end question


dboyer:
Waiting to confirm the date.

dboyer:
Our goal is to complete data collection this fall, with the end date depending on variables such as weather and participation rates. Volunteers, community groups and expert staff are all vital contributors to this effort.

PR #1820, opened 2015-10-02
Change census end FAQ text.

jwalgran:
A new answer for "When will TreesCount! 2015 end?" provides by Parks.

screen shot 2015-10-01 at 5 30 27 pm

Connects to #1818


dboyer:
That change is correct.

Issue #1814, opened 2015-09-24
Select people can add trees after census end date


dboyer:
Select users will be assigned to the waffle flag through the django admin

Add text to the front page indicating that the census is over. Redirect blacklisted pages to the homepage, as before.

Remove sidebar links for non-mappers.

maurizi:
@dboyer Do you have the text for the home page?

dboyer:
We do not have home page text yet.

PR #1821, opened 2015-10-19
Add support for limiting access to the site after the census ends.

maurizi:
Uses the existing soft launch infrastructure to support full access to
the site when the census is suspended to a limited group of people, and
allow some access for other users.

Users who do not have full access can get to the progress map, group and
event lists, group and event detail pages, the FAQ and the main training
page (but not the training subpages).

Access to editing groups and events is blocked off unless the user has
full site access, as is access to following groups.

Connects to #1814 

To test, configure the `full_access` waffle flag from the Django admin site.  You can set the value for `Everyone` to `No` to test how limited access works regardless of the logged in user, or set the flag to `Unknown` to allow configuring which users the flag does and does not apply to via the other options available.

### DIFF
--- a/src/nyc_trees/apps/core/templates/core/partials/nav.html
+++ b/src/nyc_trees/apps/core/templates/core/partials/nav.html
@@ -26,16 +26,18 @@
                     <a href="javascript:" data-toggle="modal" data-target="#treecorder-survey-modal">Treecorder</a>
                 {% endif %}
             </li>
-            <li{% if active_page == "dashboard" %} class="active"{% endif %}>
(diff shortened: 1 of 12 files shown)


RickMohr:
I verified that with limited access I could only access pages/features as described, and with full access I could do everything.

RickMohr:
Parks will have to learn how to use ID numbers to manage the primitive waffle "Users" list. Adding a user (the common case) has good UI support to search by username. But to remove someone from the list you have to go to their admin user page, get the ID from the URL, and remove it from the list of waffle user ID's. @dboyer, think that's OK?

RickMohr:
+1

dboyer:
As long as they can add them by username, it's fine if they have to remove them by user ID number.

PR #1828, opened 2015-10-27
Hide RSVP button unless user has full_access

jwalgran:
Regular users cannot RSVP to events after the census officially ends.

Connects to #1826


RickMohr:
+1 RSVP button is gone

jwalgran:
Ok to merge this if it will expedite a a staging update.

PR #1829, opened 2015-10-28
Apply "winter" text changes from Parks

RickMohr:
* Leave existing text commented-out for easy restoration in spring
* On the logged-in dashboard I gave them a slightly different order than they requested. (It would be a lot harder to have events in the middle because of how this content is shared with the user profile.):
  * Currently: events, groups, reservations, rewards
  * They wanted: reservations, rewards, events, groups
  * I gave them: reservations, rewards, groups, events

Connects #1825

### DIFF
--- a/src/nyc_trees/apps/event/templates/event/event_list_page.html
+++ b/src/nyc_trees/apps/event/templates/event/event_list_page.html
@@ -10,7 +10,10 @@
   <div class="map-sidebar">
     <h2>Events</h2>
     <p class="pageheading-description-detail">
+      NYC Parks and TreesCount! partner groups hosted more than 900 events in 2015.
+<!--
       Check out our upcoming TreesCount! events and sign up to map street trees with fellow volun<strong>treers</strong>!
+-->
     </p>
     <p class="pageheading-description-detail">Click to view as a list or find an event nearby using the map.</p>
     <div id="action-bar" class="action-bar">
--- a/src/nyc_trees/apps/users/templates/users/partials/profile/reservations.html
+++ b/src/nyc_trees/apps/users/templates/users/partials/profile/reservations.html
@@ -24,13 +24,10 @@
 {% if reservations.count > 0 %}
     {% include "core/partials/polling_download_section.html" with poll_url=reservations.map_poll_url %}
 {% else %}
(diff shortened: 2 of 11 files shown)


maurizi:
@RickMohr Instead of commenting out text, can we wrap these in`{% flag full_access%}` blocks?  That will give us flexibility in turning the census off when Parks is ready to without having to coordinate that with our deployment times.

src/nyc_trees/apps/users/templates/users/partials/profile/reservations.html
```diff
@@ -24,7 +24,16 @@
 {% if reservations.count > 0 %}
     {% include "core/partials/polling_download_section.html" with poll_url=reservations.map_poll_url %}
 {% else %}
+    <p>Ready to map on your own? If you&rsquo;re interested in mapping over the winter, please email
+        <a href="mailto:Treescount@parks.nyc.gov">Treescount@parks.nyc.gov</a>
+        to sign up for a winter data collection course and learn how to
+        keep rolling.</p>
+    <p>You can also check back in the spring to see where we still need your help, and
+        <a href="{{ reservations_url }}">reserve more available block edges for mapping</a>.
```
maurizi: This URL will not work for someone who does not have `full_access` (it will redirect to the homepage).

RickMohr:
@maurizi, fixed up per Deb & Parks. Should be ready for prime time now.

maurizi:
@azavea-bot rebuild

src/nyc_trees/apps/event/templates/event/event_list_page.html
```diff
@@ -10,7 +10,10 @@
   <div class="map-sidebar">
     <h2>Events</h2>
     <p class="pageheading-description-detail">
+      NYC Parks and TreesCount! partner groups hosted more than 900 events in 2015.
+<!--
       Check out our upcoming TreesCount! events and sign up to map street trees with fellow volun<strong>treers</strong>!
+-->
     </p>
     <p class="pageheading-description-detail">Click to view as a list or find an event nearby using the map.</p>
```
maurizi: I think we should comment this text out as well, since both the map and the list will be empty over the winter until Parks starts loading events for the spring.

maurizi:
Other than the one comment (which we could maybe let slide for now), this looks good.

Issue #1831, opened 2015-11-02
App server provisioning fails because firefox 3* is no longer available


jwalgran:
Ansible error from CI builds

```
TASK: [nyc-trees.app | Install Firefox for UI tests] ************************** 
failed: [app] => {"failed": true}
stderr: E: Version '3*' for 'firefox' was not found
```

apt command output on an existing development VM. After an `apt-get update` firefox 38.x is replaced with 41.x

```
$ apt-cache madison firefox
   firefox | 38.0+build3-0ubuntu0.14.04.1 | http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
   firefox | 38.0+build3-0ubuntu0.14.04.1 | http://security.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
   firefox | 28.0+build2-0ubuntu2 | http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
   firefox | 28.0+build2-0ubuntu2 | http://archive.ubuntu.com/ubuntu/ trusty/main Sources
   firefox | 38.0+build3-0ubuntu0.14.04.1 | http://archive.ubuntu.com/ubuntu/ trusty-updates/main Sources
   firefox | 38.0+build3-0ubuntu0.14.04.1 | http://security.ubuntu.com/ubuntu/ trusty-security/main Sources
$ sudo apt-get update
Get:1 http://security.ubuntu.com trusty-security InRelease [64.4 kB]
Ign http://archive.ubuntu.com trusty InRelease
Get:2 http://archive.ubuntu.com trusty-updates InRelease [64.4 kB]
Ign http://ppa.launchpad.net trusty InRelease
Hit http://ppa.launchpad.net trusty Release.gpg
Get:3 http://security.ubuntu.com trusty-security/main Sources [98.0 kB]
Hit http://ppa.launchpad.net trusty Release
Hit http://ppa.launchpad.net trusty/main amd64 Packages
Hit http://archive.ubuntu.com trusty Release.gpg
Get:4 http://archive.ubuntu.com trusty-updates/main Sources [242 kB]
Get:5 http://security.ubuntu.com trusty-security/universe Sources [31.0 kB]
Hit http://ppa.launchpad.net trusty/main Translation-en
Get:6 http://security.ubuntu.com trusty-security/main amd64 Packages [357 kB]
Get:7 http://security.ubuntu.com trusty-security/universe amd64 Packages [117 kB]
[...]

PR #1832, opened 2015-11-02
Fix firefox provisioning.

maurizi:
The latest version of Firefox is 41, which wasn't matched by the filter.

Connects to #1831


jwalgran:
I could not run the UI tests on my dev VM without also updating the selenium python package. In the interest of getting a green CI build, we can fix that issue separately.

+1

PR #1834, opened 2015-11-02
Hide calls to action on event page

jwalgran:
There will be no events over the winter so this instructive text could
be misleading.

### Before

screen shot 2015-11-02 at 2 13 13 pm

### After

screen shot 2015-11-02 at 2 21 27 pm

Connects to #1825


RickMohr:
+1. Did not test.

Issue #1837, opened 2016-03-10
Remove events links


dboyer:
Remove the link to the Events page in the main navigation and the current Events sections on the Groups pages. Past events will remain visible on the Group pages as a way to recognize those events. TreesCount administrators will be able to set up events using the administrative pages.

PR #1843, opened 2016-04-06
Reduce access to events

RickMohr:
* Remove "Events" page from menu
* Remove "Current" section from group events
* Remove "+ New" option when editing a group

To allow possible future use of the code I chose to comment out rather than delete.

Connects #1837


RickMohr:
@azavea-bot rebuild

maurizi:
:+1:

PR #1846, opened 2016-04-13
Update the azavea.beaver ansible role to the latest version

maurizi:
The latest version of azavea.beaver uses 36.2.0 for the beaver version.

Only the most recent version of Beaver can be installed, as one of its
dependencies was renamed and old versions were pulled from PyPi.


jwalgran:
:+1:

Issue #1836, opened 2016-03-10
Add new rewards


maurizi:
Parks to provide text for each reward.

The rewards will only track activity from a date in 2016 specified by NYC Parks. The reward levels from 2015 will no longer be implemented. Any rewards achieved by a TreesCount user in 2015 will remain visible on their user profile page. Since previous rewards will be visible, the new rewards cannot use the same names as rewards from 2015.

2016 Rewards:

- Lavender Linden—100 trees
- Magenta Maple—300 trees
- Silver Sophora—500 trees
- Gold Gingko—750 trees
- Platinum Planetree—Most Trees counted
- Four Season Mapper - surveyed at least 3 block edges in AT LEAST four of the five seasons
    - Summer 2015 (June - August)
    - Fall 2015 (September - November)
    - Winter 2015 (December - February)
    - Spring 2016 (March - May)
    - Summer 2016 (June - August)

dboyer:
Parks confirmed the incentives are correct.

maurizi:
@dboyer -  a couple of questions:

 - Can we treat 'Platinum Planetree' the same way we did 'Mayoral Mapper' last year, and not compute it "live", but let Parks determine who won it?
 - Should we exclude surveys for Spring 2015 when calculating who has achieved 'Four Season Mapper', or should we count that as part of Summer 2015 (or do something else entirely)?

Also, we need additional text.  For each reward, we need:

 - A description of how to achieve the reward (Map 100 trees, etc).
 - A description of the prize given for achieving the reward, if there is one.

dboyer:
- Yes, Platinum Planetree can be determined by Parks.

- Let's not factor in Spring 2015 for Four Season Mapper and go with the dates Parks gave. So trees counted toward four season mapper start in June 2015.

- For description of how to achieve the reward, use:
Map 100 Trees, Map 300 Trees, Map 500 Trees, Map 750 Trees, Map the Most Trees in 2016, Map at Least 3 Block Edges in Four of the Five Mapping Seasons (Summer 2015, Fall 2015, Winter 2015, Spring 2016, Summer 2016)

- Working on getting the prize text

PR #1849, opened 2016-04-18
Add new achievements

maurizi:
New achievements are temporarily using the image for 'Mayoral Mapper'
until the final images are decided upon.

To enable showing what achievements a user already has, a data migration
will update achievements for all users to record their what their inactive
achievements are, since they are no longer active for 2016 and cannot be
unlocked with new surveys.

Connects to #1836

### DIFF
--- a/src/nyc_trees/apps/home/tests.py
+++ b/src/nyc_trees/apps/home/tests.py
@@ -3,6 +3,8 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from unittest import skip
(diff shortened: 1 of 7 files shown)


maurizi:
@azavea-bot rebuild

maurizi:
@azavea-bot rebuild

jwalgran:
Using a prod snapshot counted the achievements before running the migration, and then after. As expected, some users received additional achievement rows.

jwalgran:
👍

PR #1850, opened 2016-04-18
Remove failover site from CloudFormation

hectcastro:
The CloudFormation resource for CloudFront does not yet support Amazon Certificate Manager certificates. In order to continue having a failover site, the CloudFront distribution for it will need to be managed outside of CloudFormation.


maurizi:
Looks good to me

Issue #1839, opened 2016-04-06
Update text


dboyer:
home page, the logged in dashboard, the progress map, training, groups, events, reservations, rewards, about page, FAQ page, and the email templates

Waiting on text

PR #1852, opened 2016-04-19
Implement text changes requests by Parks.

maurizi:
Connects to #1839


maurizi:
@jwalgran since there are still some open questions about these text changes, I'm assigning this to myself for now.

I'll assign it back to you once they're answered.

PR #1854, opened 2016-04-19
Show blockface ID on treecorder screen.

maurizi:
On mobile the ID is visible at all times on the preview button (previewing itself is not implemented yet).

I attempted to make the header "sticky" for the desktop view so that the blockface ID would always be visible for desktop users as well, but I ran into cross-browsers styling issues and decided to leave it as-is.

treecorder1

Connects to #1840

### DIFF
--- a/src/nyc_trees/apps/survey/templates/survey/survey.html
+++ b/src/nyc_trees/apps/survey/templates/survey/survey.html
@@ -42,13 +42,16 @@
         </div>
         <div id="survey" class="hidden">
             <div id="tree-form-header">
-                <h4>Treecorder</h4>
+                <h4>Treecorder &ndash; Block <span data-block-id></span></h4>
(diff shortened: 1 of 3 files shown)


dboyer:
If inserting the comma into the block ID number is a hassle, we can leave it out.

maurizi:
@dboyer - it wasn't a hassle, which is why I did it.

jwalgran:
If you scroll to the bottom of the page, the footer with the TreeKit logo covers up the button.

screen shot 2016-04-20 at 5 03 01 pm

jwalgran:
+1

maurizi:
> If you scroll to the bottom of the page, the footer with the TreeKit logo covers up the button.

Yes, that was in the wireframe movie thingy

Issue #1845, opened 2016-04-07
When you are logged out, the progress map does not show block edge tiles


jwalgran:
This may be a regression created by https://github.com/azavea/nyc-trees/pull/1844

PR #1855, opened 2016-04-20
Fix loading tiles on progress map when logged out

maurizi:
We omitted the dropdown from the progress map when logged out, because
after removing the group view in 83aa2242 there was only a single mode
when not logged in.

This broke loading the tiles on the page, as JS reads the DOM to know
which mode is currently active.  Hiding the element instead of omitting it
fixes the issue while still leaving it not shown for logged out users.

Connects to #1845


jwalgran:
👍

Issue #1842, opened 2016-04-06
Show help overlay on Treecorder


dboyer:
When a user visits the Treecorder survey page for the first time, they will see an overlay (similar to the progress map) with informative text explaining they can click the button to view a preview of their data. The overlay will not be visible on subsequent visits to the survey page.

PR #1856, opened 2016-04-20
Show preview button help popup when first visiting the treecorder

maurizi:
To test, log in, reserve a block edge, and visit the treecorder.
Upon first getting to the data entry part of the treecorder you should see
a modal.  On later uses of the treecorder you should not see a modal.

Connects to #1842


jwalgran:
👍

Issue #1860, opened 2016-04-21
Implement final text changes


maurizi:
- Need to add "Reward in development. Stay tuned!"

- Possible changes to event text on rewards and dashboard pages?

dboyer:
- In registration email, change the first sentence after the activation email link to: "Once you have clicked the link above, your email address will be added to our list for updates on TreesCount! progress.

- Text not visible on Progress Map page. Tested as independent mapper with and without winter mapping access and public user and it never appeared.

- There's an extra period after "training materials here" on the training page when viewing it as a user who has completed training.

- There needs to be a space between "gear!" and "This block edge..." on the new reservation confirmation page.(blockedge/checkout-confirmation/)

maurizi:
Also need to change "Restart Block" to just "Restart" on the treecorder preview screen.

dboyer:
From Parks:

FAQ AND ABOUT
·         When can I view the data?

Please change NYC Street Tree Map to “NYC Tree Map”       

·         Please Remove “If you have an urgent question, please call 212-360-TREE”

dboyer:
- Remove events section from logged in dashboard

PR #1863, opened 2016-04-29
Implement final text changes from Parks

maurizi:
Connects to #1860


RickMohr:
+1 all looks good on the site.

PR #1870, opened 2016-05-06
fix #1866. map no longer cut off or showing through on the survey for…

### LINKED ISSUE #1866: Crosshairs on Treecorder block selection misaligned
On smaller screens, the crosshairs fall behind the help text or are not centered on the map.

designmatty:
@maurizi It turns out the sliver of map peaking through was there. On mobile, you needed to add a second tree and then collapse the trees. The sidebar wouldnt be tall enough to cover the screen.

This PR should fix that.

### DIFF
--- a/src/nyc_trees/sass/partials/_survey.scss
+++ b/src/nyc_trees/sass/partials/_survey.scss
@@ -68,6 +68,14 @@
             @media (max-height: $screen-xs-height) and (max-width: $screen-sm) {
                 top: 4.5rem;
             }
+            /**
+             * iPhone 5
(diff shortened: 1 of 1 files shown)


maurizi:
Testing on browserstack on an iPhone 5S, I see a very large amount of whitespace on the bottom of the survey page when on this branch.

designmatty:
so it turns out ios 7 does not like vh units. I found this solution https://gist.github.com/pburtchaell/e702f441ba9b3f76f587 and adapted it. I believe this fixes the problem.

maurizi:
:+1: Looks great.

Issue #1861, opened 2016-04-21
Respond to testing feedback


dboyer:

dboyer:
1. Add badge art 

2. Activity toward new rewards should counting on May 11. For Four Season Mapper, Spring 2016 is March-June and Summer 2016 is July to September

dboyer:
3. Remove "register for an event today" from the Ready, Set, Roll reward from 2015

4. Remove "need a wheel? visit one of our nearby loaning hubs or check one out at your next NYC Parks mapping event" from the Counter Cultured reward from 2015

PR #1871, opened 2016-05-06
Make changes based on feedback from Parks

maurizi:
- Add badges for new achievements
 - Update four seasons achievement to accommodate new date ranges.
 - Update 2016 achievements to only count trees surveyed after May 11th
 - Remove references to events and loaning hubs from 2015 achievements
 - Only require validation of distance to tree and curb location fields to
   view preview

Connects to #1861


src/nyc_trees/libs/sql.py
```diff
@@ -25,16 +25,22 @@ def _get_rows(sql, params=[]):
         return cursor.fetchall()
 
 
-def get_user_tree_count(user):
+def get_user_tree_count(user, from_date=None):
```
RickMohr: I don't see any calls to this function that pass the `from_date` parameter. Is it necessary?

RickMohr:
Otherwise all works and looks good.

RickMohr:
+1 once that question is addressed.

Issue #1867, opened 2016-05-05
Tiles not loading when visit block preview for a second time


dboyer:
See attached screenshots of the issue on iPad and Galaxy. Parks notes that when they visit the block preview the first time, it loads correctly. When they use the block preview multiple times on the same block, the preview map doesn't load. A page refresh is necessary.

![galaxy_preview](https://cloud.githubusercontent.com/assets/516764/15049835/6f324d4e-12be-11e6-8996-3013c86ac5aa.png)
![ipad_preview](https://cloud.githubusercontent.com/assets/516764/15049836/719c6042-12be-11e6-94a2-4f3958f5e73a.jpg)

PR #1872, opened 2016-05-06
Fix map not showing when switching between preview map and normal map

maurizi:
When changing the map element size (e.g if the phone orientation changes),
if the map is not currently visible it will get stuck in an invalid state.

Calling `.invalidateSize()` when switching between the preview map and the
normal survey map will cause Leaflet to requery the DOM for the map size,
fixing the issue.

I wasn't able to create issues with the normal survey map, but I put in a call to `invalidateSize()` for it when hiding the preview map, since it can't really hurt matters.

To test:
 - On the develop branch, show the preview map, close it, resize the browser window, and show the preview map again.  You should see the issue reported in #1867.
 - On this branch, repeat the same steps.  You should not experience the issue.

Connects to #1867

### DIFF
--- a/src/nyc_trees/js/src/surveyPage.js
+++ b/src/nyc_trees/js/src/surveyPage.js
@@ -522,6 +522,10 @@ function showPreview(trees) {
         });
         previewLayer.addTo(previewMap);
     }
+    // If the browser window is resized when the preview map is hidden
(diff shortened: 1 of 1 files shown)


RickMohr:
+1, works!

PR #1880, opened 2016-06-21
Bump Apache and JVM versions

hectcastro:
Version 7u95 of the OpenJDK JVM is no longer available.

See also: azavea/ansible-apache2#5


kdeloach:
+1 I was able to provision each VM flawlessly.

Issue #1882, opened 2016-07-12
Add rewards text


dboyer:
1. Change the descriptive text for the reward to include the year. Finished Online Training - 2015, Map 300 Trees - 2016, etc.

2. Add rewards text:

Lavender Linden 100 trees counted– 1, 2, Tree Shirts - 201(6) edition
Magenta Maple 300 trees counted – TreesCount! 2015 branded coffee tumbler. Keeps your coffee warm or cold!
Silver Sephora 500 trees counted– TreesCount! 2015 branded sports duffel bag – The first 50 voluntreers to reach 500 trees since 5/13/2016 will receive an enviable bright green bag for your sports duds.

PR #1883, opened 2016-07-26
Update achievement descriptions and rewards.

maurizi:
I also removed a few out-dated comments about hard-coded URLs.

Connects to #1882

### DIFF
--- a/deployment/ansible/services.yml
+++ b/deployment/ansible/services.yml
@@ -6,6 +6,9 @@
     - name: Update APT cache
       apt: update_cache=yes cache_valid_time=3600
 
+    - name: Install ACL utilities
+      apt: pkg="acl" state=present
+
   roles:
     - { role: "nyc-trees.common" }
 
--- a/src/nyc_trees/apps/users/models.py
+++ b/src/nyc_trees/apps/users/models.py
@@ -108,24 +108,22 @@ class AchievementDefinition(object):
 achievements = OrderedDict([
     (AchievementDefinition.FOLLOW_GROUPS, AchievementDefinition(
         name='In Pursuit of Mappiness',
-        description='Follow 5 Groups',
-        description_achieved='Followed 5 Groups',
+        description='Follow 5 Groups - 2015',
+        description_achieved='Followed 5 Groups - 2015',
         badge='img/badges/ic_badge_follow.png',
         achieved=lambda user: Follow.objects.filter(user=user).count() >= 5
     )),
     (AchievementDefinition.ONLINE_TRAINING, AchievementDefinition(
         name='Ready, Set, Roll',
-        description='Finish Online Training',
-        # NOTE: using hard-coded URL here because it's too early for reverse()
-        description_achieved="Finished Online Training",
+        description='Finish Online Training - 2015',
+        description_achieved="Finished Online Training - 2015",
         badge='img/badges/ic_badge_online_training.png',
         achieved=lambda user: user.online_training_complete
     )),
     (AchievementDefinition.TRAINING_EVENT, AchievementDefinition(
         name='Treerifically Trained',
-        description='Attend a Training Event',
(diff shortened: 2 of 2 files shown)


hectcastro:
Can you please try applying this diff to your PR to see if it appeases Jenkins? It seems to be working for me locally:

```diff
diff --git a/ansible.cfg b/ansible.cfg
index 0451870..c75e7ba 100644
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,3 @@
 [defaults]
+pipelining = True
 callback_whitelist = profile_tasks
```

maurizi:
@hectcastro That does not seem to have worked :cry:

hectcastro:
I installed the `acl` package on the `services` VM manually and retried the Jenkins job to help confirm that it would resolve the issue (given that I cannot reproduce it locally):

```diff
diff --git a/deployment/ansible/services.yml b/deployment/ansible/services.yml
index d0398ac..6ee3fb1 100644
--- a/deployment/ansible/services.yml
+++ b/deployment/ansible/services.yml
@@ -6,6 +6,9 @@
     - name: Update APT cache
       apt: update_cache=yes cache_valid_time=3600
 
+    - name: Install ACL utilities
+      apt: pkg="acl" state=present
+
   roles:
     - { role: "nyc-trees.common" }
```

maurizi:
@hectcastro That may not be necessary? Naginator restarted the failed build for d2d32ef and it seems to have succeeded...

hectcastro:
That was me initiating the retry, and I installed the `acl` package on the `services` VM manually after it failed against the latest in this PR.

maurizi:
I see, thanks for the assist!

deployment/ansible/services.yml
```diff
@@ -6,6 +6,9 @@
     - name: Update APT cache
       apt: update_cache=yes cache_valid_time=3600
 
+    - name: Install ACL utilities
```
jwalgran: It would be good to leave a commit message explaining why we need to add this and the `piplining = True` option.

src/nyc_trees/apps/users/models.py
```diff
@@ -223,44 +221,47 @@ def __init__(self, name, description, description_achieved, badge,
...
-        description='Map 500 Trees',
-        description_achieved='Mapped 500 Trees',
+        description='Map 500 Trees - 2016',
+        description_achieved='Mapped 500 Trees - 2016',
         badge='img/badges/Silver-sophora.png',
         achieved=lambda user: get_user_tree_count(user, REWARD_START) >= 500,
         active=True,
-        reward="Reward in development. Stay tuned!"
+        reward="""TreesCount! 2015 branded sports duffel bag – The first 50
+            volun<b>treer</b>s to reach 500 trees since 5/11/2016 will receive
```
jwalgran: The date here is `5/11/2016` but [the issue](https://github.com/azavea/nyc-trees/issues/1882) text has the date as `5/13/2016`. Just double checking that this change is understood by everyone.

deployment/ansible/services.yml
hectcastro: This package includes the `setfacl` tool, which Ansible 2.1+ uses to set POSIX ACLs to share modules with unprivileged users (`postgres` in this case).

Because setting `pipelining = True` didn't work during the first pass, I suspect that we can omit that and just stick with installing the `acl` package.

See: http://docs.ansible.com/ansible/become.html

jwalgran:
These look correct, but I have been having trouble getting my VMs running in order to see the changes in the app.

src/nyc_trees/apps/users/models.py
maurizi: Yeah, the rewards are counted starting from `5/11` but Parks had their first official mapping event on `5/13`. I talked with @dboyer and we decided to change the date.

jwalgran:
👍 

rewards

PR #1884, opened 2016-07-29
Use AliasTarget keyword arguments for parameters

hectcastro:
Ensure that we are supplying arguments to the `AliasTarget` resource in the correct way. Previously, keyword arguments were not available, so the order of constructor arguments meant something. That meaning was altered along the way, which led to silent parameter swapping.

---

**Testing**

The [`nyc-trees-staging-deployment`](http://civicci01.internal.azavea.com/view/nyc-trees/job/nyc-trees-staging-deployment/497/console) job in Jenkins already executed a deployment to staging with the changes in this PR. Ensure that the application is working correctly with whatever new changes were introduced.


maurizi:
:+1: Changes look reasonable, and the recent changes are up on the staging server.

Issue #1879, opened 2016-06-16
Tool to append lat/lng to a CSV of tree measuements


jwalgran:
Refer to 2016_06_15_TreesCount_Support_Hours_Geometry_Constructor.docx

maurizi:
New spreadsheet:
projects/NYCParks_StreetTreeCensus/downloads/2016_11_23_NYC_Parks_Data_for_Treecorder.xlsx

jwalgran:
Some technical discussion that happened during the estimation process

> We can adapt the JS getLatLngs function to work in Python, get the start and end points of the specified blockface, and determine to which blockface point the supplied survey start point is closer. Since we are only comparing distances, not using them for math, we don’t have to worry about the fact that we will be running linear distance calculations on spherical points.

> Psudo code

```
lat_lngs = get_lat_lngs(blockface)
distance_to_start = calculate_distance(start_point, lat_lngs[0])
distance_to_end = calculate_distance(start_point, lat_lngs[len(lat_lngs) - 1])
survey.is_mapped_in_blockface_polyline_direction = distance_to_start < distance_to_end
```

PR #1891, opened 2017-01-18
Add management command to calculate tree locations from CSV

maurizi:
This command scans a CSV file of survey data and executes the Treekit geometry construction algorithm to calculate tree locations.
Lat/lng fields are appended to each row of the original CSV.

This builds on top of work done by @kdeloach, and adds the ability to determine which end of the blockface the survey was started at by comparing the distance from the newly provided X/Y points to both the start and ends of the blockface geometry.

I also corrected a bug in the original command: the command was making the `distance_to_tree` field be from the start of the blockface for each tree, but the geometry constructor expects `distance_to_tree` to be the distance from the previous tree (or the start of the blockface) which is conveniently what the spreadsheet provides.

There were a number of duplicate rows in the provided spreadsheet, as well as several rows with no blockface ID. I decided to drop any rows with no Blockface ID, as well as drop any duplicate rows.

This is the input CSV I used (it is the `WITH_XY` sheet of the provided spreadsheet, sorted by Blockface ID and Tree ID).
[2016_11_23_NYC_Parks_Data_for_Treecorder.txt](https://github.com/azavea/nyc-trees/files/714938/2016_11_23_NYC_Parks_Data_for_Treecorder.txt)

This is the output I received.
[output.txt](https://github.com/azavea/nyc-trees/files/720508/output.txt)

I verified my results by manually running through the Treecorder for Blockface `203652`, and verified that the Lat/Lng provided by the Geometry Constructor when used via the UI matched the Lat/Lng provided by this command.

Connects to #1879

### DIFF
--- a/src/nyc_trees/apps/survey/management/commands/treecorder.py
+++ b/src/nyc_trees/apps/survey/management/commands/treecorder.py
@@ -11,13 +11,18 @@ import itertools
 
 from collections import namedtuple
 
+from django.contrib.gis.gdal.srs import SpatialReference, CoordTransform
+from django.contrib.gis.geos import Point
 from django.core.management.base import BaseCommand
 
-from apps.survey.models import Survey
+from apps.survey.models import Survey, Blockface
 from apps.survey.views import _survey_geojson
 
 
-Record = namedtuple('Record', ['report_id',
+Record = namedtuple('Record', ['record_id',
+                               'report_id',
+                               'boro_name',
+                               'boro_code',
                                'plot_number',
                                'block_face_number',
                                'tree_id_number',
@@ -28,7 +33,25 @@ Record = namedtuple('Record', ['report_id',
                                'from_st',
                                'to_st',
                                'address_order',
-                               'side_of_centerline'])
+                               'side_of_centerline',
+                               'make_edit',
+                               'start_end',
+                               'dist_to',
+                               'mapped_correctly',
+                               'comment',
+                               'x_point',
+                               'y_point'])
+
+
+ny_state_plane = SpatialReference(
+    '+proj=lcc +lat_1=40.66666666666666 +lat_2=41.03333333333333 '
+    '+lat_0=40.16666666666666 +lon_0=-74 +x_0=300000 +y_0=0 '
+    '+ellps=GRS80 +datum=NAD83 +to_meter=0.3048006096012192 +no_defs ')
+ny_state_plane.validate()
+
+wgs84 = SpatialReference(4326)
+
+ny_to_latlng = CoordTransform(ny_state_plane, wgs84)
 
 
 class Command(BaseCommand):
@@ -55,10 +78,15 @@ class Command(BaseCommand):
                                             lambda r: r.block_face_number)
 
         for blockface_id, rows in grouped_records:
-            rows = list(rows)
+            # The WITH_XY spreadsheet has a bunch of duplicate survey rows,
+            # need to deduplicate...
+            seen = set()
+            rows = [row for row in rows
+                    if row.tree_id_number not in seen and
+                    not seen.add(row.tree_id_number)]
             survey = self.create_survey(rows[0])
 
-            if not survey.blockface_id:
+            if survey is None:
                 continue
 
             trees = list(self.create_trees(rows))
@@ -72,7 +100,7 @@ class Command(BaseCommand):
 
             # Combine tree location with original CSV row.
             for row, point in itertools.izip(rows, points):
-                lat, lng = point['coordinates']
+                lng, lat = point['coordinates']
                 writer.writerow(row + (lat, lng))
 
     def create_survey(self, row):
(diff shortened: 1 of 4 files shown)


kdeloach:
:clap: Feel free to squash my WIP commit.

maurizi:
Waffle

src/nyc_trees/apps/survey/management/commands/treecorder.py
```diff
@@ -0,0 +1,148 @@
...
+        # give it a CoordTransform object
+        survey_start_point = Point(float(row.x_point), float(row.y_point),
+                                   srid=102718)
+        survey_start_point.transform(ny_to_latlng)
+
+        blockface_start = Point(blockface.geom.coords[0][0], srid=4326)
+        blockface_end = Point(blockface.geom.coords[0][0], srid=4326)
+
+        start_dst = blockface_start.distance(survey_start_point)
+        end_dst = blockface_end.distance(survey_start_point)
```
maurizi: Typo: This should be `survey_end_point` here

src/nyc_trees/apps/survey/management/commands/treecorder.py
```diff
@@ -0,0 +1,148 @@
...
+        survey.is_mapped_in_blockface_polyline_direction = start_dst < end_dst
+
+        return survey
+
+    def create_trees(self, rows):
+        """
+        Return survey formatted data to calculate tree locations.
+        Assumes all rows belong to the same blockface.
+        """
+        rows = list(rows)
```
jwalgran: I see this function being called in one place, with the output of a list comprehension as the `rows` parameter. Is this call to `list` redundant?

src/nyc_trees/apps/survey/management/commands/treecorder.py
```diff
@@ -0,0 +1,148 @@
...
+            return None
+
+        # We need to pass an srid to call .transform(), but it is ignored if we
+        # give it a CoordTransform object
+        survey_start_point = Point(float(row.x_point), float(row.y_point),
+                                   srid=102718)
+        survey_start_point.transform(ny_to_latlng)
+
+        blockface_start = Point(blockface.geom.coords[0][0], srid=4326)
+        blockface_end = Point(blockface.geom.coords[0][0], srid=4326)
```
jwalgran: `blockface_start` and `blockface_end` are being set to `coords[0][0]`, which does not look right. I think `blockface_end` should be to
`coords[length_of_first_dimension-1][length_of_second_dimension-1]`

I used [an online CSV to GeoJSON converter](http://www.convertcsv.com/csv-to-geojson.htm) (flipping the lat and lng columns) and loaded the resulting file into QGIS with the blockface layer and Google satellite imagery.

Here is a screenshot of blockface 225494

screen shot 2017-01-20 at 12 21 58 pm

This appears to support the idea that the geometries were constructed from the wrong start point.

src/nyc_trees/apps/survey/management/commands/treecorder.py
```diff
@@ -0,0 +1,148 @@
...
+            # Execute Treekit geometry constructor to determine
+            # tree locations.
+            points = _survey_geojson(survey, trees)
+
+            # Parse GeoJSON result.
+            points = (json.loads(p) for p in points)
+
+            # Combine tree location with original CSV row.
+            for row, point in itertools.izip(rows, points):
+                lat, lng = point['coordinates']
```
jwalgran: The values in the `output.txt` attached to the PR appear to be swapped. I think this needs to be

```python
lng, lat = point['coordinates']

**Review by jwalgran — CHANGES_REQUESTED** [review]

File: src/nyc_trees/apps/survey/management/commands/treecorder.py (hunk 4 above)
Comment (maurizi) [reply src/nyc_trees/apps/survey/management/commands/treecorder.py]: Fixed in  	401beb2

File: src/nyc_trees/apps/survey/management/commands/treecorder.py (hunk 3 above)
Comment (maurizi) [reply src/nyc_trees/apps/survey/management/commands/treecorder.py]: Doh!

I changed this to be `Point(blockface.geom.coords[0][-1], srid=4326)` (the last point of the first segment) in  401beb2.

I'm not sure what blockfaces (if any) have multiple line segments, but the [`getLatLngs` in `mapUtils.js`](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/lib/mapUtil.js#L21-L28) always uses the first segment, so I chose to mirror that.

File: src/nyc_trees/apps/survey/management/commands/treecorder.py (hunk 2 above)
Comment (maurizi) [reply src/nyc_trees/apps/survey/management/commands/treecorder.py]: Not necessary, removed in  	401beb2

**Comment by maurizi** [comment]
New CSV after the PR fixups were made:
[output.txt](https://github.com/azavea/nyc-trees/files/720508/output.txt)

File: src/nyc_trees/apps/survey/management/commands/treecorder.py [hunk 5]
```diff
@@ -0,0 +1,147 @@
...
+            return None
+
+        # We need to pass an srid to call .transform(), but it is ignored if we
+        # give it a CoordTransform object
+        survey_start_point = Point(float(row.x_point), float(row.y_point),
+                                   srid=102718)
+        survey_start_point.transform(ny_to_latlng)
+
+        blockface_start = Point(blockface.geom.coords[0][0], srid=4326)
+        blockface_end = Point(blockface.geom.coords[0][-1], srid=4326)
```
jwalgran, line 127: Blockface 225494 now looks correct

screen shot 2017-01-23 at 9 54 13 am

and compares visually to the output produced by the treecorder

screen shot 2017-01-23 at 10 28 15 am

jwalgran (approved):
Looks good. I spot checked a few additional blocks comparing my QGIS project with the original xlsx provided by Parks.

